Issue #2, opened 2019-04-17
Option for histograms as alternative to KDE


williamjameshandley:
Currently we only support scatter plots and kernel density estimated contours as options to ``NestedSamples.plot_2d`` and ``NestedSamples.plot_1d``. It would be good if there were an option to plot histograms and histogram-estimated contours, in a similar manner to [corner.py](https://corner.readthedocs.io/en/latest/pages/custom.html)

This could also be fed into the GUI, since histograms may be fast enough to compute so that they can be plotted dynamically.

PR #10, opened 2019-07-07
Option of using 1D histograms instead of KDE plots

lukashergt:
Provides part of the enhancement for #2 "Option for histograms as alternative to KDE".

# Changes overview

* Appended `.gitignore`

* Created function `hist_1d()` in plot.py to be called instead of the KDE plot from the function `plot_1d()`.
  - Uses (optionally) astropy's `hist` function which has the same functionality as matplotlib's own hist function with the extension of moresophisticated algorithms for determining bins. The bins keyword e.g. can be set to 'knuth' (default) or 'blocks' (even better but less reliable). Reverts to matplotlib's `hist` if astropy is not installed.
  - The histograms are renormalised such that the highest bar has a height of unity. This simplifies the choice of y-limits and since the y-axis isn't labelled in the diagonal 1D plots of the triangle plots, the absolute height isn't of much interest.
  - added documentation to the hist_1d function
  - cleaned code to make it pep8 conform

* In sampler.py in `MCMCSamples.plot_2d` the argument `types` is now accepting a dictionary specifying the plot types for the `diagonal`, `lower` and `upper` plots. `plot_2d` now also accepts `diagonal_kwargs`, `lower_kwargs` and `upper_kwargs` as kwargs for the different types of sub-plots.

* added a bugfix to utils.py

# Checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ plikHM_TTTEEE_lowl_lowE_lensing_NS.tar.gz
 plikHM_TTTEEE_lowl_lowE_lensing_NS/
 .idea/*
 *~
+.pytest_cache/*
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,5 @@ plikHM_TTTEEE_lowl_lowE_lensing.tar.gz
 plikHM_TTTEEE_lowl_lowE_lensing/
 plikHM_TTTEEE_lowl_lowE_lensing_NS.tar.gz
 plikHM_TTTEEE_lowl_lowE_lensing_NS/
+.idea/*
+*~
--- a/anesthetic/plot.py
+++ b/anesthetic/plot.py
@@ -20,7 +20,6 @@ except ImportError:
     from matplotlib.pyplot import hist
 from anesthetic.kde import kde_1d, kde_2d
 from anesthetic.utils import check_bounds, nest_level, unique
-from anesthetic.utils import kwargs_in_func
 from scipy.interpolate import interp1d
 from matplotlib.ticker import MaxNLocator
 from matplotlib.colors import LinearSegmentedColormap
@@ -253,8 +252,7 @@ def plot_1d(ax, data, *args, **kwargs):
     p /= p.max()
     i = (p >= 1e-2)
 
-    plot_kwargs = kwargs_in_func(kwargs, plt.plot)
-    ans = ax.plot(x[i], p[i], *args, **plot_kwargs)
+    ans = ax.plot(x[i], p[i], *args, **kwargs)
     ax.set_xlim(*check_bounds(x[i], xmin, xmax), auto=True)
     return ans
 
@@ -302,10 +300,8 @@ def hist_1d(ax, data, *args, **kwargs):
         xmax = data.max()
     histtype = kwargs.pop('histtype', 'bar')
 
-    hist_kwargs = kwargs_in_func(kwargs, plt.hist)
-    hist_kwargs.update(kwargs_in_func(kwargs, hist))
     h, edges, bars = hist(data, ax=ax, range=(xmin, xmax), histtype=histtype,
-                          *args, **hist_kwargs)
+                          *args, **kwargs)
     # As the y-axis on the diagonal 1D plots of the triangle plot won't
     # be labelled, we set the maximum bar height to 1:
     if histtype == 'bar':
@@ -377,17 +373,15 @@ def contour_plot_2d(ax, data_x, data_y, *args, **kwargs):
     cmap = basic_cmap(color)
     zorder = max([child.zorder for child in ax.get_children()])
 
-    contourf_kwargs = kwargs_in_func(kwargs, plt.contourf)
     cbar = ax.contourf(x[i], y[j], pdf[numpy.ix_(j, i)], contours,
                        vmin=0, vmax=1.0, cmap=cmap, zorder=zorder+1,
-                       *args, **contourf_kwargs)
+                       *args, **kwargs)
     for c in cbar.collections:
         c.set_cmap(cmap)
 
-    contour_kwargs = kwargs_in_func(kwargs, plt.contour)
     ax.contour(x[i], y[j], pdf[numpy.ix_(j, i)], contours,
                vmin=0, vmax=1.2, linewidths=0.5, colors='k', zorder=zorder+2,
-               *args, **contour_kwargs)
+               *args, **kwargs)
     ax.set_xlim(*check_bounds(x[i], xmin, xmax), auto=True)
     ax.set_ylim(*check_bounds(y[j], ymin, ymax), auto=True)
     return cbar
@@ -424,8 +418,7 @@ def scatter_plot_2d(ax, data_x, data_y, *args, **kwargs):
     ymin = kwargs.pop('ymin', None)
     ymax = kwargs.pop('ymax', None)
 
-    plot_kwargs = kwargs_in_func(kwargs, plt.plot)
-    points = ax.plot(data_x, data_y, 'o', markersize=1, *args, **plot_kwargs)
+    points = ax.plot(data_x, data_y, 'o', markersize=1, *args, **kwargs)
     ax.set_xlim(*check_bounds(data_x, xmin, xmax), auto=True)
     ax.set_ylim(*check_bounds(data_y, ymin, ymax), auto=True)
     return points
--- a/anesthetic/plot.py
+++ b/anesthetic/plot.py
@@ -87,7 +87,7 @@ def make_1d_axes(params, **kwargs):
     return fig, axes
 
 
-def make_2d_axes(params, **kwargs):
+def make_2d_axes(params, types={'1d': 'kde', '2d':['kde', 'scatter']}, **kwargs):
     """Create a set of axes for plotting 2D marginalised posteriors.
 
     Parameters
@@ -145,7 +145,6 @@ def make_2d_axes(params, **kwargs):
         grid = GS(*axes.shape, hspace=0, wspace=0)
 
     upper = kwargs.pop('upper', None)
-    diagonal = kwargs.pop('diagonal', True)
 
     if kwargs:
         raise TypeError('Unexpected **kwargs: %r' % kwargs)
@@ -163,10 +162,10 @@ def make_2d_axes(params, **kwargs):
             lower = not (x in axes.index and y in axes.columns
                          and all_params.index(x) > all_params.index(y))
 
-            if x == y and not diagonal:
+            if x == y and '1d' not in types:
                 continue
 
-            if upper == lower and x != y:
+            if upper == lower and x != y and '1d' not in types:
                 continue
 
             if axes[x][y] is None:
@@ -176,6 +175,12 @@ def make_2d_axes(params, **kwargs):
                 sy = sy[0] if sy else None
                 axes[x][y] = fig.add_subplot(grid[j, i],
                                              sharex=sx, sharey=sy)
+
+            if upper == lower and x != y and '1d' in types:
+                axes[x][y].axis('off')
+                axes[x][y]._upper = None
+                continue
+
             if x == y:
                 axes[x][y].twin = axes[x][y].twinx()
                 axes[x][y].twin.set_yticks([])
@@ -252,7 +257,7 @@ def plot_1d(ax, data, *args, **kwargs):
 
 
 def hist_1d(ax, data, *args, **kwargs):
-    """Plot a 1d histogram
+    """Plot a 1d histogram.
 
     This functions is a wrapper around matplotlib.axes.Axes.hist, or
     astropy.visualization.hist if it is available. astropy's hist function
@@ -272,20 +277,6 @@ def hist_1d(ax, data, *args, **kwargs):
         optional, default data.min() and data.max()
         cannot be None (reverts to default in that case)
 
-    bins: int or list or str (optional)
-        str only works with astropy's hist function. Then it must be one of:
-        - 'blocks' : use bayesian blocks for dynamic bin widths
-        - 'knuth' : use Knuth's rule to determine bins
-        - 'scott' : use Scott's rule to determine bins
-        - 'freedman' : use the Freedman-Diaconis rule to determine bins
-
-    max_bins: int (optional)
-        Only used by astropy.visualization.hist().
-        Maximum number of bins allowed. With more than a few thousand bins
-        the performance of matplotlib will not be great. If the number of
-        bins is large *and* the number of input data points is large then
-        it will take a very long time to compute the histogram.
-
     Returns
     -------
     patches : list or list of lists
@@ -293,15 +284,15 @@ def hist_1d(ax, data, *args, **kwargs):
         or list of such list if multiple input datasets.
 
     Other Parameters
-    ---------------
+    ----------------
     **kwargs : `~matplotlib.axes.Axes.hist` properties
 
     """
     if data.max()-data.min() <= 0:
         return
 
-    xmin = kwargs.pop('xmin', data.min())
-    xmax = kwargs.pop('xmax', data.max())
+    xmin = kwargs.pop('xmin', None)
+    xmax = kwargs.pop('xmax', None)
     if xmin is None:
         xmin = data.min()
     if xmax is None:
--- a/anesthetic/plot.py
+++ b/anesthetic/plot.py
@@ -166,7 +166,7 @@ def make_2d_axes(params, **kwargs):
             if x == y and not diagonal:
                 continue
 
-            if upper == lower and x != y and not diagonal:
+            if upper == lower and x != y:
                 continue
 
             if axes[x][y] is None:
@@ -176,12 +176,6 @@ def make_2d_axes(params, **kwargs):
                 sy = sy[0] if sy else None
                 axes[x][y] = fig.add_subplot(grid[j, i],
                                              sharex=sx, sharey=sy)
-
-            if upper == lower and x != y and diagonal:
-                axes[x][y].axis('off')
-                axes[x][y]._upper = None
-                continue
-
             if x == y:
                 axes[x][y].twin = axes[x][y].twinx()
                 axes[x][y].twin.set_yticks([])
--- a/anesthetic/plot.py
+++ b/anesthetic/plot.py
@@ -20,6 +20,7 @@ except ImportError:
     from matplotlib.pyplot import hist
 from anesthetic.kde import kde_1d, kde_2d
 from anesthetic.utils import check_bounds, nest_level, unique
+from anesthetic.utils import kwargs_in_func
 from scipy.interpolate import interp1d
 from matplotlib.ticker import MaxNLocator
 from matplotlib.colors import LinearSegmentedColormap
@@ -252,7 +253,8 @@ def plot_1d(ax, data, *args, **kwargs):
     p /= p.max()
     i = (p >= 1e-2)
 
-    ans = ax.plot(x[i], p[i], *args, **kwargs)
+    plot_kwargs = kwargs_in_func(kwargs, plt.plot)
+    ans = ax.plot(x[i], p[i], *args, **plot_kwargs)
     ax.set_xlim(*check_bounds(x[i], xmin, xmax), auto=True)
     return ans
 
@@ -300,14 +302,10 @@ def hist_1d(ax, data, *args, **kwargs):
         xmax = data.max()
     histtype = kwargs.pop('histtype', 'step')
 
-    # Non-exhaustive (to be expanded with usage...) list of kwargs that might
-    # be desired for the other triangle plots but that would clash with hist:
-    not_hist_kwargs = ['s', 'filled', 'linestyles']
-    for key in not_hist_kwargs:
-        _ = kwargs.pop(key, None)
-
+    hist_kwargs = kwargs_in_func(kwargs, plt.hist)
+    hist_kwargs.update(kwargs_in_func(kwargs, hist))
     h, edges, bars = hist(data, ax=ax, range=(xmin, xmax), histtype=histtype,
-                          *args, **kwargs)
+                          *args, **hist_kwargs)
     # As the y-axis on the diagonal 1D plots of the triangle plot won't
     # be labelled, we set the maximum bar height to 1:
     if histtype == 'bar':
@@ -379,15 +377,17 @@ def contour_plot_2d(ax, data_x, data_y, *args, **kwargs):
     cmap = basic_cmap(color)
     zorder = max([child.zorder for child in ax.get_children()])
 
+    contourf_kwargs = kwargs_in_func(kwargs, plt.contourf)
     cbar = ax.contourf(x[i], y[j], pdf[numpy.ix_(j, i)], contours,
                        vmin=0, vmax=1.0, cmap=cmap, zorder=zorder+1,
-                       *args, **kwargs)
+                       *args, **contourf_kwargs)
     for c in cbar.collections:
         c.set_cmap(cmap)
 
+    contour_kwargs = kwargs_in_func(kwargs, plt.contour)
     ax.contour(x[i], y[j], pdf[numpy.ix_(j, i)], contours,
                vmin=0, vmax=1.2, linewidths=0.5, colors='k', zorder=zorder+2,
-               *args, **kwargs)
+               *args, **contour_kwargs)
     ax.set_xlim(*check_bounds(x[i], xmin, xmax), auto=True)
     ax.set_ylim(*check_bounds(y[j], ymin, ymax), auto=True)
     return cbar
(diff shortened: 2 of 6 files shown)


.gitignore
```diff
@@ -9,3 +9,5 @@ plikHM_TTTEEE_lowl_lowE_lensing.tar.gz
 plikHM_TTTEEE_lowl_lowE_lensing/
 plikHM_TTTEEE_lowl_lowE_lensing_NS.tar.gz
 plikHM_TTTEEE_lowl_lowE_lensing_NS/
+.idea/*
```
williamjameshandley: What are ``.idea/`` files?

anesthetic/plot.py
```diff
@@ -14,12 +14,14 @@
 import pandas
 import matplotlib.pyplot as plt
 from matplotlib.gridspec import GridSpec as GS, GridSpecFromSubplotSpec as SGS
+from astropy.visualization import hist
```
williamjameshandley: ``astropy`` is an excellent package, although it is a pretty heavyweight one. Given that it's only usage here is for using it's more flexible ``hist`` function, it would be good if we could make this dependency optional, i.e. to try importing it, and if it fails to just use the default matplotlib ``hist``

e.g.
```python
try:
    from astropy.visualization import hist
except ImportError:
    from matplotlib.pyplot import hist
```

anesthetic/plot.py
```diff
@@ -246,6 +248,86 @@ def plot_1d(ax, data, *args, **kwargs):
     return ans
 
 
+def hist_1d(ax, data, *args, **kwargs):
+    """Plot a 1d histogram
+
+    This functions is a wrapper around matplotlib.axes.Axes.hist, with the
```
williamjameshandley: This docstring isn't quite correct, as it's a wrapper around astropy's hist function

anesthetic/plot.py
```diff
@@ -246,6 +248,86 @@ def plot_1d(ax, data, *args, **kwargs):
...
+
+    xmin = kwargs.pop('xmin', data.min())
+    xmax = kwargs.pop('xmax', data.max())
+    if xmin is None:
+        xmin = data.min()
+    if xmax is None:
+        xmax = data.max()
+    bins = kwargs.pop('bins', 'knuth')
+    histtype = kwargs.pop('histtype', 'step')
+    density = kwargs.pop('density', True)
```
williamjameshandley: If we're renormalising anyway, do we need the density keyword?

anesthetic/utils.py
```diff
@@ -48,8 +48,8 @@ def quantile(a, q, w):
         w = numpy.ones_like(a)
     i = numpy.argsort(a)
     c = numpy.cumsum(w[i])
-    c /= c[-1]
-    icdf = interp1d(c/c[-1], a[i])
+    c /= c.values[-1]
+    icdf = interp1d(c/c.values[-1], a[i])
```
williamjameshandley: Did you find that there was a bug that needed this fix?

williamjameshandley:
It also doesn't quite resolve #2 , since we'd need 2D histograms for that to be fully complete

.gitignore
lukashergt: Those come with the PyCharm IDE which I use for cleaning up a project (nicely flags pep8 stuff...). Can just keep it locally though, if you don't want that.

williamjameshandley: There shouldn't be any problem with keeping that in the ``.gitignore``

anesthetic/plot.py
lukashergt: Well, astropy's hist function is a wrapper around matplotlib's hist function. I wrote it this way, because I assume people are more familiar with the matplotlib hist function, so I documented astropy's hist function in more detail whereas for the rest I just point to matplotlib's hist function.

anesthetic/plot.py
lukashergt: No, I guess we can drop that now that I've also fixed the ylim.

anesthetic/utils.py
lukashergt: Doesn't work without this for me... Threw some error.

williamjameshandley: Hopefully when I have finally got round to a proper test coverage this kind of issue would be thrown up anyway

anesthetic/plot.py
williamjameshandley: Perhaps a statement saying "This functions as a wrapper around matplotlib.axes.Axes.hist, or astropy.visualization.hist if it is available"

anesthetic/plot.py
lukashergt: That would work. Should specify in the documentation which keywords go with the astropy version then. Won't be able to use 'knuth' as default for 'bins' anymore, though.

anesthetic/plot.py
lukashergt: Guess that would be more complete.

anesthetic/plot.py
williamjameshandley: In general, it's a good idea to not override the defaults unless there's a good reason not to.

It looks like the default for astropy is the same as matplotlib and numpy (``bins=10``). Your extension allows the option of ``knuth`` to be passed through (which you prefer), but you should remove the overriding defaults that you have here.

lukashergt: modified accordingly

anesthetic/samples.py
```diff
@@ -118,9 +118,14 @@ def plot(self, ax, paramname_x, paramname_y=None, *args, **kwargs):
 
         if paramname_y is None or paramname_x == paramname_y:
             xmin, xmax = self._limits(paramname_x)
-            return plot_1d(ax, self[paramname_x].compress(),
-                           xmin=xmin, xmax=xmax,
-                           *args, **kwargs)
+            plot_type_1d = kwargs.pop('plot_type_1d', 'kde')
```
williamjameshandley: I'm also not sure about the introduction of this new keyword argument. It makes a lot of sense for your use-case (i.e. histograms on diagonal, KDEs elsewhere), but isn't very consistent with the other ``MCMCSamples`` functions that call ``MCMCSamples.plot``.

Ideally, what we would have is a single ``plot_type`` kwarg, which can take values 
- ``'kde'``, which in the 2D case would give you kdes on both off-diagonal and diagonals, 
- ``'hist'`` which (when implemented) would give you histograms everywhere. For now, you should put in a ``NotImplementedError`` in the 2d sections of the code when users try to call this
-  A dictionary ``{'1d':'hist', '2d':'kde'}``, which would give your use case

You'd need to have a little more logic in the kwarg parsing that uses ``isinstance`` to check if it's a dictionary

lukashergt: A 2D hist is typically used much less, because it doesn't combine with other models/data... That why I wanted 1D hists separate from 2D, but a dictionary might be more elegant.

williamjameshandley:
The travis CI build is currently failing on account of errors caused by the command:

```bash
pydocstyle --convention=numpy anesthetic
```

Have you checked the output of that command on your local system recently?

anesthetic/samples.py
williamjameshandley: I'm not sure I agree (and indeed corner makes a lot of use of 2D histograms). In general there are quite a few options for histograms/kdes (including plotting both). It's best to try and keep the functionality all inside a single keyword argument if possible.

anesthetic/plot.py
```diff
@@ -246,6 +251,84 @@ def plot_1d(ax, data, *args, **kwargs):
...
+    if data.max()-data.min() <= 0:
+        return
+
+    xmin = kwargs.pop('xmin', data.min())
+    xmax = kwargs.pop('xmax', data.max())
+    if xmin is None:
+        xmin = data.min()
+    if xmax is None:
+        xmax = data.max()
+    histtype = kwargs.pop('histtype', 'step')
```
williamjameshandley: I'm not sure we should override this default either

anesthetic/plot.py
```diff
@@ -246,6 +251,84 @@ def plot_1d(ax, data, *args, **kwargs):
...
+        lower/upper prior bound
+        optional, default data.min() and data.max()
+        cannot be None (reverts to default in that case)
+
+    bins: int or list or str (optional)
+        str only works with astropy's hist function. Then it must be one of:
+        - 'blocks' : use bayesian blocks for dynamic bin widths
+        - 'knuth' : use Knuth's rule to determine bins
+        - 'scott' : use Scott's rule to determine bins
+        - 'freedman' : use the Freedman-Diaconis rule to determine bins
```
williamjameshandley: You probably don't need these docstrings, only a note that kwargs and args are forwarded to the appropriate histogram functionality. Whilst documentation is generally good, if it's liable to decay/outdate it's better to not have it (good documentation >> no documentation >> wrong documentation)

anesthetic/plot.py
```diff
@@ -246,6 +251,84 @@ def plot_1d(ax, data, *args, **kwargs):
...
+    xmax = kwargs.pop('xmax', data.max())
+    if xmin is None:
+        xmin = data.min()
+    if xmax is None:
+        xmax = data.max()
+    histtype = kwargs.pop('histtype', 'step')
+
+    # Non-exhaustive (to be expanded with usage...) list of kwargs that might
+    # be desired for the other triangle plots but that would clash with hist:
+    not_hist_kwargs = ['s', 'filled', 'linestyles']
```
williamjameshandley: This is similar to the first comment above, I'm not sure how well this will work going forward. Is check absolutely necessary?

anesthetic/plot.py
```diff
@@ -246,6 +251,84 @@ def plot_1d(ax, data, *args, **kwargs):
...
+    Other Parameters
+    ---------------
+    **kwargs : `~matplotlib.axes.Axes.hist` properties
+
+    """
+    if data.max()-data.min() <= 0:
+        return
+
+    xmin = kwargs.pop('xmin', data.min())
+    xmax = kwargs.pop('xmax', data.max())
```
williamjameshandley: For consistency with the rest of the code, these should probably have ``None`` as defaults

anesthetic/plot.py
lukashergt: Something of this sort is necessary, especially with all the stripping of defaults. At the moment almost non of the kwargs (such as histtype, bins, etc.) can be used, because the hist kwargs conflict with the plot kwargs and with the contour kwargs...

lukashergt: Is there a way of getting all possible kwargs for a function (say 'hist')?

anesthetic/plot.py
lukashergt: What default would you prefer? I need to pop this for the sake of renormalisation.

anesthetic/plot.py
lukashergt: deleted accordingly

anesthetic/plot.py
lukashergt: changed accordingly

anesthetic/samples.py
lukashergt: I turned types and plot_type into dictionaries. I think this might make the keywords upper and diagonal obsolete, as one can simply check whether the key '1d' exists in the dictionary (if so then diagonal=True) and whether there is just one or two entries for the key '2d'.

What do you think?

anesthetic/plot.py
lukashergt: Added a function to utils.py that lets us single out the kwargs used by a function before passing them on. This should be cleaner than listing them by hand.

anesthetic/plot.py
williamjameshandley: The matplotlib default is ``'bar'``. I agree that there might be aesthetic reasons for choosing ``'step'`` over this type, but for now stick with the default.

anesthetic/samples.py
```diff
@@ -114,25 +115,35 @@ def plot(self, ax, paramname_x, paramname_y=None, *args, **kwargs):
             Pandas array of axes objects
 
         """
-        plot_type = kwargs.pop('plot_type', 'kde')
+        plot_type = kwargs.pop('plot_type', {'1d': 'kde', '2d': 'kde'})
```
williamjameshandley: Currently this code expects ``types`` to be a dictionary, which breaks backward compatibility, since previously all code expects a string (hence the failing tests). You should keep the same default as before, but have a line something along the lines of:
```python
if isinstance(plot_type, str):
    plot_type = {'1d':plot_type, '2d':plot_type}
```
This will allow it to handle both cases

anesthetic/plot.py
```diff
@@ -82,7 +88,8 @@ def make_1d_axes(params, **kwargs):
     return fig, axes
 
 
-def make_2d_axes(params, **kwargs):
+def make_2d_axes(params, types={'1d': 'kde', '2d': ['kde', 'scatter']},
+                 **kwargs):
```
williamjameshandley: I'm not sure this change is an improvement, and does result in different behaviour from before: could you revert it for now (leaving the previous behaviour with a 'diagonal' keyword argument), and submit it as a separate pull request if you think that it's worth including.

anesthetic/utils.py
```diff
@@ -156,3 +157,9 @@ def unique(a):
         if x not in b:
             b.append(x)
     return b
+
+
+def kwargs_in_func(kwargs_dict, func):
+    """Find all kwargs that will be used by func."""
+    aspec = inspect.getargspec(func)
+    return {key: kwargs_dict[key] for key in kwargs_dict if key in aspec.args}
```
williamjameshandley: This could be a really nice solution to the problem, however I'm concerned that matplotlib is not capable of inspecting it's keyword arguments like this (indeed general python code gives no such guarantees). For example, if you tried using this on one of the functions in anesthetic, it wouldn't be able to extract the kwargs that it are expected due to the 'pop' implementation. Does this code actually work for your plotting examples?

I think the only way to address this problem in the current paradigm is to have two additional kwargs to ``MCMCSamples.plot_2d``: ``1d_kwargs``, ``2d_kwargs``. These should be dictionaries, and indicate the arguments that should be passed *only* to the 1d or 2d plotting functions. The remaining *args and **kwargs should get passed to both (e.g. a color-based kwarg).

lukashergt: I tried this with `plot`, `hist`, and `contour` and it seems to work. At least it catches all `args` of those functions. I did not make an exhaustive test of all possible `kwargs` though.

I don't think `1d_kwargs` and `2d_kwargs` would be solution we want, because `plot_1d` and `hist_1d` already have different `kwargs`.

anesthetic/plot.py
lukashergt: How should `diagonal` and `types` or `plot_type` interact? Should `diagonal` turn the `'1d'` keyword on and off?

williamjameshandley: I agree that if ``'1d'`` is not present, then that should imply ``diagonal=False``, but I think that if a user were to provide just a string as a ``plot_type``, then by default both 1d and 2d plots should be produced, and ``diagonal=True`` by default

anesthetic/utils.py
williamjameshandley: > I did not make an exhaustive test of all possible kwargs though.

The issue is that there are a lot of possible keyword arguments to the plotting functions, and an exhaustive test is not really feasible now, or for future changes to matplotlib/additions to anesthetic

> I don't think 1d_kwargs and 2d_kwargs would be the solution we want, because plot_1d and hist_1d already have different kwargs.

I don't think that would matter in practice. The kind of code I was imagining would be:
```python

1d_kwargs_temp = kwargs.pop('1d_kwargs',{})
1d_kwargs = kwargs.copy()
1d_kwargs.update(1d_kwargs_temp)
```
and then passing ``1d_kwargs`` to the relevant 1d function in place of ``kwargs``, independent of whether it's a ``hist`` or ``kde``

anesthetic/samples.py
lukashergt: Do you mean `types` or `plot_type`, here? Or both?

williamjameshandley: Sorry, I meant ``types``, i.e.
```python
if isinstance(types, str):
    types = {'1d':types, '2d':types}
```

williamjameshandley: I'm getting confused, my apologies:

 In ``MCMCSamples.plot_2d`` (where there is no reference to ``types``) you should have the code:

```python
if isinstance(plot_type, str):
    plot_type = {'1d':plot_type, '2d':plot_type}
```

whilst in ``make_2d_axes`` the ``types`` keyword should be removed

anesthetic/utils.py
williamjameshandley: (the above code would occur in ``MCMCSamples.plot_2d``, whilst ``MCMCSamples.plot_1d`` would not need such preprocessing)

anesthetic/samples.py
lukashergt: Do you mean `MCMCSamples.plot()`? Because `plot_2d` does reference `types`. `plot` indeed doesn't and I've reverted things there.

williamjameshandley: Indeed you are correct. I've been looking at a wrong version of the code.

On master, ``MCMCSamples.plot_2d`` expects a keyword argument ``types`` which at the moment can be either a 
- str (in which case it produces a triangle plot),
- list (in which case it produces a rectangle plot with two different types above and below the diagonal)

One way to implement the diagonal behaviour we want would be to allow this to take a length three list (or a dictionary) indicating the upper, lower and diagonal behaviour.   I think that a dictionary would be a lot nicer than how it's currently implemented. We could make it backward compatible by converting lists appropriately, but that may not be essential if the new functionality is a lot clearer.

``MCMCSamples.plot`` should remain unchanged in it's processing of plot_type, i.e. it should always take a string, but now you can pass it ``'hist'`` for the case that it's a 1d plot (and it should throw an ``NotImplementedError`` if you try to do a 2d histogram plot).

anesthetic/utils.py
williamjameshandley: With regands the discussion above on how to implement ``types``, plot_2d should take 'upper_kwargs', 'lower_kwargs' and 'diagonal_kwargs' to pass on to the respective ``plot`` routine

anesthetic/plot.py
lukashergt: done

anesthetic/samples.py
lukashergt: * reverted `MCMCSamples.plot` (see recent commit)
* changed `types` to a dictionary with keywords `'diagonal'`, `'lower'` and `'upper'` and updated docs accordingly. Should be backwards compatible with `str` and `list` though.

Should I add a deprecation warning that now the use of a dictionary is advised?

williamjameshandley: That would be great

lukashergt: done, I made it a FutureWarning as that is what pops up in jupyter notebooks, too.

lukashergt:
Do you know what makes the tests fail? 

It might be related to me creating invisible axes when `'upper'` is not provided. The invisible axes are useful for legend placement and annotations.

anesthetic/utils.py
lukashergt: created `upper_kwargs`, `lower_kwargs` and `diagonal_kwargs` for `MCMCsamples.plot_2d`

anesthetic/samples.py
```diff
@@ -209,21 +242,55 @@ def plot_2d(self, axes, *args, **kwargs):
...
+                         'lower': types[0],
+                         'upper': types[-1]}
+            else:
+                types = {'lower': types[0], 'upper': types[-1]}
+        diagonal_kwargs = kwargs.pop('diagonal_kwargs', {})
+        lower_kwargs = kwargs.pop('lower_kwargs', {})
+        upper_kwargs = kwargs.pop('upper_kwargs', {})
+        diagonal_kwargs.update(kwargs)
+        lower_kwargs.update(kwargs)
+        upper_kwargs.update(kwargs)
```
williamjameshandley: 1. I believe that this implementation means that ``kwargs`` takes precedence over the ``diagonal_kwargs`` etc. I think it might be more intuitive if diagonal_kwargs overrode the kwargs entries
2. We could either have three separate arguments (as you have implemented here), or a dictionary of overrides (which would be more consistent with the types argument), taking ``'diagonal'``, ``'lower'`` and ``'upper'`` as keys. Either works, but we need to make a decision in this PR as to what the default should be. What do you reckon?

anesthetic/samples.py
```diff
@@ -132,7 +140,8 @@ def plot(self, ax, paramname_x, paramname_y=None, *args, **kwargs):
             nsamples = 500
             plot = scatter_plot_2d
         else:
-            raise ValueError("plot_type must be in {'kde', 'scatter'}")
+            raise ValueError("plot_type is '%s', but must be in "
+                             "{'kde', 'scatter'}." % plot_type)
```
williamjameshandley: It's probably more consistent to call this a NotImplementedError

anesthetic/plot.py
```diff
@@ -171,6 +176,12 @@ def make_2d_axes(params, **kwargs):
                 sy = sy[0] if sy else None
                 axes[x][y] = fig.add_subplot(grid[j, i],
                                              sharex=sx, sharey=sy)
+
+            if upper == lower and x != y and diagonal:
+                axes[x][y].axis('off')
+                axes[x][y]._upper = None
+                continue
+
```
williamjameshandley: Why did you need to change this implementation? It could be this that is causing the failing tests

anesthetic/samples.py
```diff
@@ -194,11 +203,35 @@ def plot_2d(self, axes, *args, **kwargs):
...
+            The options for 'diagonal are either:
+
+                - 'kde'
+                - 'hist.
+
+            The options for 'lower' and 'upper' are either:
+
+                - 'kde'
+                - 'scatter'
+
```
williamjameshandley: You don't need the extra blank lines here.

williamjameshandley:
>(How do you link to the changed lines the way you are doing it? Would have liked to reference them, but don't know how.)

Click the 'Files Changed' tab at the top, and press the plus sign that appears on a line

anesthetic/samples.py
lukashergt: 1. This was actually an active decision of mine. I feel like anything directly passed to `plot_2d` should have precedence over something that gets indirectly passed to one subplot. Passing it directly kind of means "this is supposed to apply to all plots, not just one". For example if we passed `color` directly, it would ensure that the lines in the diagonal plots match in colour to the contours in the lower plots. If you really wanted different colours in the different plots, then you should specify it in all sub-kwargs and not pass it directly.

2. Making a dictionary here would turn it into a dictionary of dictionaries and I fear that would look rather ghastly...?

anesthetic/samples.py
lukashergt: Agreed, but was hesitant to overwrite your code...

anesthetic/samples.py
williamjameshandley: 1. OK. I agree with this reasoning. You should add a note in the documentation that kwargs overrides the individual ones.
2. Actual lol. OK, we should prioritise the user over implementation. You might find that you can clean up the code by constructing such a dictionary and then looping.

anesthetic/samples.py
williamjameshandley: Don't be -- PRs are a good opportunity to clean up inconsistencies

anesthetic/samples.py
lukashergt: Was matching it to the itemisation above, but actually agree, I prefer it without the blank lines, too. Changed accordingly.

anesthetic/plot.py
lukashergt: This is what I was referring to earlier (where I didn't know how to reference the changed lines...).

I personally prefer having all axes created and the unused ones turned invisible. This is useful for later placements of legends or annotations. What are your thoughts?

And yes, this is causing the failing tests. Am happy to adapt the tests for this, but only if you'd be happy with the change.

williamjameshandley: That's a difficult one. I can see pros and cons either way, which means it's probably a good idea to make this some kind of optional behaviour (e.g. with an 'invisible' argument). Revert for now, as it's definitely a separate change with regards to this PR, but submit this as an issue for further discussion (we can also use that to discuss how to make ``make_2d_axes`` more consistent with the new ``'upper', 'lower', 'diagonal'`` scheme.

anesthetic/samples.py
lukashergt: 1. done

2. Haha, three is not enough for a loop to justify the ugliness of dict in dict, here. Keeping it as is.

williamjameshandley: OK. If we can get the tests passing and this merged, I'll submit a pull request that you can review that suggests how it might be made neater.

anesthetic/plot.py
lukashergt: reverted.
tests now passing :)

williamjameshandley:
Apologies, I mis-merged the upstream changes with an extra blank line at line 102 of ``anesthetic/gui/plot.py``. Could you remove that and then I'll approve and merge to master.

williamjameshandley:
Never mind, turns out I can push to your pull request (that's new since last I managed a repo)

williamjameshandley:
Congratulations @lukashergt on your first successful PR. Many thanks for providing this functionality.

PR #12, opened 2019-07-08
change plot to scatter for 2D scatter plot option

lukashergt:
* Change `ax.plot` to `ax.scatter` in `scatter_plot_2d` in plot.py
  - This is more true to the intention of the plots and provides the
    additional functionalities of a scatter plot (could be expanded to
    make so-to-speak '3D' plots using a colormap).
  - Change default markersize to s=3 as the matplotlib default is
    typically too large for the amount of points we are dealing with
    here.
  - Documentation adapted accordingly.

Changes to be committed:
* modified:   anesthetic/plot.py

# Checklist:

### DIFF
--- a/anesthetic/plot.py
+++ b/anesthetic/plot.py
@@ -384,7 +384,7 @@ def contour_plot_2d(ax, data_x, data_y, *args, **kwargs):
 def scatter_plot_2d(ax, data_x, data_y, *args, **kwargs):
     """Plot samples from a 2d marginalised distribution.
 
-    This functions as a wrapper around matplotlib.axes.Axes.plot, enforcing any
+    This functions as a wrapper around matplotlib.axes.Axes.scatter, enforcing any
     prior bounds. All remaining keyword arguments are passed onwards.
 
     Parameters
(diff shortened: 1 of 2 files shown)


anesthetic/plot.py
```diff
@@ -402,17 +402,18 @@ def scatter_plot_2d(ax, data_x, data_y, *args, **kwargs):
...
+    matplotlib.collections.PathCollection object
+        A PathCollection object representing the plotted data (same as
+        matplotlib matplotlib.axes.Axes.scatter command)
 
     """
     xmin = kwargs.pop('xmin', None)
     xmax = kwargs.pop('xmax', None)
     ymin = kwargs.pop('ymin', None)
     ymax = kwargs.pop('ymax', None)
+    s = kwargs.pop('s', 3)
```
williamjameshandley: I assume that this default change is necessary because the default size is too large. Does ``markersize=1`` roughly correspond to ``s=3``?

tests/test_plot.py
```diff
@@ -208,26 +209,26 @@ def test_scatter_plot_2d():
...
 
     ax = plt.gca()
-    contour_plot_2d(ax, data_x, data_y, ymin=ymin)
+    scatter_plot_2d(ax, data_x, data_y, ymin=ymin)
     assert(ax.get_ylim()[0] >= ymin)
     plt.close()
 
     ax = plt.gca()
-    contour_plot_2d(ax, data_x, data_y, ymax=ymax)
+    scatter_plot_2d(ax, data_x, data_y, ymax=ymax)
```
williamjameshandley: Good job you found this bug

anesthetic/plot.py
lukashergt: roughly, yes, that's why I picked it.

williamjameshandley:
Many thanks @lukashergt . I think it's definitely more pythonic to have scatter plots be implemented by ``scatter`` rather than ``plot``.

PR #14, opened 2019-07-09
add MontePythonReader and MontePythonSamples classes

lukashergt:
* Created a new file `read/montepythonreader.py` containing the class
  `MontePythonReader` which reads and processes MontePython chain files
  using `montepython.analyze`.
  - Extracts parameter names from MontePython's log.param file.
  - Removes burn-in and non-markovian points from the data.
  - Extracts tex names of the parameters.
  - Extracts parameter limits from MontePython's log.param file.

* The new class `MontePythonSamples` in sampler.py calls
  `MontePythonReader` and turns the data into an `MCMCSamples` instance.

Changes to be committed:
* modified:   anesthetic/__init__.py
* new file:   anesthetic/read/montepythonreader.py
* modified:   anesthetic/samples.py

# Checklist:


williamjameshandley:
I very much approve of a monte python reader.

Could you give me a bit more background on how a ``MontePythonSamples`` class differs from ``MCMCSamples``? I would have thought that these were pretty much the same

lukashergt:
I modelled `MontePythonSamples` to be very similar to `MCMCSamples` and potentially it could be incorporated in it, but then we'd need to decide on how to tell it to use the `MontePythonReader` instead of the `GetDistReader`.

williamjameshandley:
My query is whether one needs an entire new class for MontePythonSamples. At the moment there are MCMCSamples, which covers getdist format (I believe that @Pablo-Lemos may have a cosmosis format, which if so should be offered as a PR when they have time), and NestedSamples, which covers MultiNest and PolyChord formatting.

I had envisaged that the role of the readers was to convert samples into one of these two classes to provide a unified 'inference interface'. NestedSamples are distinct from MCMCSamples, because they provide additional functionality such as evidence calculation and run-time replaying, but it's not clear to me what MontePython adds over GetDist in its samples which couldn't just be incorporated into the MCMCSamples class

lukashergt:
The difference between MontePython and GetDist is similar to the difference between MultiNest and PolyChord I think. So, yes, we probably could have the `MCMCSamples` use the `MontePythonReader`. 

I guess I am not sure how to make `MCMCSamples` differentiate between Montepython and GetDist.

williamjameshandley:
Ah, I see. OK. You'll have to implement an ``MCMCReader``, similar to a ``NestedReader``, which handles recognising whether samples are in montepython format or getdist format. Have a look at ``anesthetic/read/nested.py``, which organises the different cases for ``MultiNestReader`` and ``PolyChordReader``

lukashergt:
Ok, I've managed to merge `MontePythonSamples` into `MCMCSamples`. I have replaced the `NestedReader` by a more general `SampleReader` which distinguishes between MultiNest, PolyChord, GetDist, and MontePython.

What do you think?

williamjameshandley:
This looks good. I like how you're using montepython's native chain reading. Makes me wonder whether a similar strategy should be applied for GetDist. 

Are you able to rename ``montepythonreader.py`` to just ``montepython.py``, or does that introduce a name conflict?

Ideally you should also add a couple of tests to confirm that the code works on a variety of machines. I assume that we want to keep montepython as an optional dependency, so it might be worth implementing an 'optional install' for the .travis system, as is done for [fgivenx](https://github.com/williamjameshandley/fgivenx/blob/master/.travis.yml)

williamjameshandley:
How essential is it to use MontePython's native functionality (in case people just have chains, but don't want to install montepython)?

lukashergt:
* `montepythonreader.py` needs to be named that way else we have a conflict with the module (took me quite some time to figure that one out, rather silent error...)

* How would I test for different machines? And how do you do an 'optional install'?

* Instead of passing the folder as root you can pass the chain files as root which will lead to GetDist reading the files. However then montepython's removal of burn-in and non-markovian points is not performed and no limits are provided.

williamjameshandley:
> montepythonreader.py needs to be named that way else we have a conflict with the module (took me quite some time to figure that one out, rather silent error...)

I thought as much. Would it be worth suffixing the other files with 'reader' for consistency?

> How would I test for different machines? And how do you do an 'optional install'?

Link to fgivenx above should give you the .travis.yml file modifications. Basically you have two different requirements files that it chooses to read from, one 'minimal' and the other with all optional dependencies installed. To do this you'd need to create the second requirements file, and extend the env section of .travis.yml a-la fgivenx.

> Instead of passing the folder as root you can pass the chain files as root which will lead to GetDist reading the files. However then montepython's removal of burn-in and non-markovian points is not performed and no limits are provided.

It's worth remarking that the getdist ones should also potentially have burn-in removed, but anesthetic does not do this.

lukashergt:
> Would it be worth suffixing the other files with 'reader' for consistency?

I think that would be good. I actually generally think it is good when the module name reflects the class name therein.

> It's worth remarking that the getdist ones should also potentially have burn-in removed, but anesthetic does not do this.

What do you mean? Add a comment in the docs of `MCMCSamples` that no burn-in was removed for getdist samples?

williamjameshandley:
>What do you mean? Add a comment in the docs of MCMCSamples that no burn-in was removed for getdist samples?

That would be for another PR. For now you could raise an issue that our system doesn't provide an interface for removing burn in.

lukashergt:
`montepython` unfortunately is not pip installable, so I think we shouldn't put it in the `requirements.txt` file, right? How is this to work with the `.travis.yml` file then?

.travis.yml
```diff
@@ -30,6 +54,14 @@ matrix:
...
           language: generic
           env: CONDA=1 MPLBACKEND=Agg 
+        # OSX montepython_master
+        - os: osx
+          language: generic
+          env: MONTEPYTHONMASTER=1 MPLBACKEND=Agg
+        # OSX montepython_master conda
+        - os: osx
+          language: generic
+          env: MONTEPYTHONMASTER=1 CONDA=1 MPLBACKEND=Agg
```
lukashergt: Is this what you had in mind, Will, when you suggested testing for other platforms? Basically the tests work, however, on OSX `pip install pypolychord` fails. E.g. it works on my linux desktop but fails on my macbook. This seems more of an issue with `pypolychord` though and less with `anesthetic`, right?

.travis.yml
```diff
@@ -52,6 +84,18 @@ install:
...
+          wget https://github.com/brinckmann/montepython_public/archive/3.1.0.tar.gz;
+          tar -xf 3.1.0.tar.gz;
+          pip install montepython_public-3.1.0/;
+      fi;
+    - if [[ "$MONTEPYTHONMASTER" == 1 ]];
+      then
+          wget https://github.com/brinckmann/montepython_public/tarball/master;
+          tar -xf master;
+          pip install brinckmann-montepython_public-*/;
+      fi;
```
lukashergt: Does this look generally sensible?

Where is this executed? Is this done temporarily and then removed or am I missing something and potentially spamming some folder with this...?

anesthetic/read/montepythonreader.py
```diff
@@ -0,0 +1,158 @@
...
+        """
+        # The variable and function names used here correspond to the ones
+        # used in MontePython's analyze.py module.
+        command_line = Namespace(files=[self.root])
+        self.info = analyze.Information(command_line)
+        analyze.prepare(files=command_line.files, info=self.info)
+        analyze.extract_parameter_names(info=self.info)
+        analyze.find_maximum_of_likelihood(info=self.info)
+        data_per_chain = analyze.remove_bad_points(info=self.info)
+        self.data = numpy.concatenate(data_per_chain, axis=0)
```
williamjameshandley: Is there a reason why this code is not in the ``__init__`` function?

tests/test_samples.py
```diff
@@ -63,6 +69,14 @@ def test_read_getdist():
     mcmc.plot_1d(['x0', 'x1', 'x2', 'x3'])
 
 
+@pytest.mark.skipif('montepython' not in sys.modules,
+                    reason="requires montepython package")
```
williamjameshandley: In an ideal world, for users the montepython reader would raise the appropriate exception if users try to read montepython chains when montepython is not installed. Can you see how this might be implemented? Then your test could actually test for this behaviour by checking that the exception is raised when it's not installed, and that the code completes when it is.

.travis.yml
williamjameshandley: This is certainly good enough. I may submit a PR in the future neatening up some of this code, but for now it achieves what we want in a way that's generally consistent with the rest of the travis tests. Well done for getting it to work (the repeated pushes trying to get travis to work are part of the usual workflow of setting up continuous integration)

williamjameshandley: Does anesthetic require ``pypolychord`` in any way? I can't see why we'd need it to be a dependency, and you are correct that pip install [does not currently work on OSX](https://github.com/PolyChord/PolyChordLite/issues/13) for PolyChord.

.travis.yml
williamjameshandley: No, travis will clean everything up on its servers once we're done, so you're fine to leave them as they are, providing they're not in a location that interferes with imports (which I think by default they are not).

.travis.yml
lukashergt: no, `pypolychord` is not necessary here, which is why now montepython is installed without dependencies since commit [117d526](https://github.com/williamjameshandley/anesthetic/pull/14/commits/117d5261c5b430d90cf7c98fc81e1e18c2aa825a) .

anesthetic/read/montepythonreader.py
lukashergt: Yes that was changed in commit [080ab4a ](https://github.com/williamjameshandley/anesthetic/pull/14/commits/080ab4a0d98748a95335b1fae821d53ba4d178f8):

> * Deferred the proper calls to `montepython.analyze` to the `_init`
>   method in `MontePythonReader` (previously already done in `__init__`)
>   such that an instance can be created and its properties used even if
>   the provided root actually is not a MontePython root. This allows the
>   necessary checks in `SampleReader` to be done.

This allows `SampleReader` to use `MontePythonReader`'s properties such as `log_param_file`, `paramnames_file` and `log_file` in order to determine whether the given root belongs to a MontePython folder.

tests/test_samples.py
lukashergt: Good point, changed in commit [79933a6](https://github.com/williamjameshandley/anesthetic/pull/14/commits/79933a66d9fbb5f510b344f5caff11c790a808a0) to work now with `xfail` and not with `skipif`.

lukashergt:
> That would be for another PR. For now you could raise an issue that our system doesn't provide an interface for removing burn in.

Then again, getdist only performs burn-in removal by discarding a certain fraction of the first steps. This would be easy enough to do with a pandas dataframe after the sample has been read with `MCMCSamples`...

williamjameshandley:
>Then again, getdist only performs burn-in removal by discarding a certain fraction of the first steps. This would be easy enough to do with a pandas dataframe after the sample has been read with MCMCSamples... Possibly all that is required here is warning about this in the documentation somewhere?

Not if the chains come from multiple files which are then combined. If anesthetic labelled which chain the points had come from (not a bad idea anyway), then this would be easy enough to do, but it would be most easy if there were an additional keyword argument passed to MCMCSamples and the reader. Definitely worth raising as an issue (or two) to remind somebody to do this at some point.

tests/test_samples.py
```diff
@@ -63,6 +69,15 @@ def test_read_getdist():
...
 
 
+@pytest.mark.xfail('montepython' not in sys.modules,
+                   raises=ImportError,
+                   reason="requires montepython package")
+def test_read_montepython():
+    mcmc = MCMCSamples(root='./tests/example_data/mp')
+    mcmc.plot_2d(['x0', 'x1', 'x2', 'x3'])
+    mcmc.plot_1d(['x0', 'x1', 'x2', 'x3'])
+
```
williamjameshandley: I like the use of this decorator to implement the conditional passing -- not something I've used before.

williamjameshandley:
Many thanks @lukashergt . This kind of PR is essential for increasing the usage of this package

PR #15, opened 2019-07-09
make_2d_plots upgrade

### LINKED ISSUE #13: KeyError when providing `types` only in first call to `plot_2d`
The following bug is introduced through #10:

When using `plot_2d` twice subsequently (i.e. in the same figure), but only providing the argument `types` the first time with `types` including the keyword `'upper'`, then the second call to `plot_2d` the default for `types` will be assumed which does not include `'upper'` and thus we are getting a KeyError.

Example for two `NestedSamples` instaces `ns1` and `ns2`:
```python
fig, ax = ns1.plot_2d(['n_s', 'r'], types={'diagonal': 'kde', 'lower': 'kde', 'upper': 'scatter'})
ns2.plot_2d(axes=ax)
```

This throws the following KeyError:
```python
.virtualenvs/py27env/lib/python2.7/site-packages/anesthetic/samples.pyc in plot_2d(self, axes, *args, **kwargs)
    288                         ax_ = ax.twin
    289                     elif ax._upper:
--> 290                         plot_type = types['upper']
    291                         kwargs = upper_kwargs
    292                     else:

KeyError: 'upper'
```

`types` should be inherited from first to second call to `plot_2d` within the same figure. Maybe make it an attribute of the figure or axes? Or can it be somehow inferred from the existing axes?

Current workaround is to consistently use the same `types` in every call.

williamjameshandley:
# Description

This upgrades ``make_2d_plots`` to use a more user-intuitive definition of ``'upper'``, and adds the keywords ``'lower'`` and ``'diagonal'``. Code is also in theory easier to read.

Fixes #13

# Checklist:


anesthetic/plot.py
```diff
@@ -131,57 +128,66 @@ def make_2d_axes(params, **kwargs):
         xparams, yparams = params
     else:
         xparams = yparams = params
+
+    upper = kwargs.pop('upper', True)
+    lower = kwargs.pop('lower', True)
+    diagonal = kwargs.pop('diagonal', True)
```
lukashergt: Are these defaults inconsistent with the default for `types` in `samples.plot_2d`?
```python
types = kwargs.pop('types', {'diagonal': 'kde', 'lower': 'kde'})
```

tests/test_plot.py
```diff
@@ -103,28 +106,51 @@ def test_make_2d_axes():
...
+        n011 = (~axes011.isna()).sum().sum()
+        n100 = (~axes100.isna()).sum().sum()
+        n001 = (~axes001.isna()).sum().sum()
+        n010 = (~axes010.isna()).sum().sum()
+
+        assert(n111 == nx*ny)
+        assert(n100+n010+n001 == nx*ny)
+        assert(n100+n011 == n111)
+        assert(n010+n101 == n111)
+        assert(n001+n110 == n111)
```
lukashergt: nitpicky, but in my opinion the last three assertions here should also compare to `nx*ny` both in terms of consistency as in terms of testing logic...

tests/test_samples.py
```diff
@@ -86,28 +89,45 @@ def test_read_multinest():
...
 def test_plot_2d_types():
     ns = NestedSamples(root='./tests/example_data/pc')
     params_x = ['x0', 'x1', 'x2', 'x3']
     params_y = ['x0', 'x1', 'x2']
-    fig, axes = ns.plot_2d([params_x, params_y], types=['kde', 'scatter'])
-    assert(axes.isnull().sum().sum() == 0)
+    params = [params_x, params_y]
+    # Test old interface
+    fig, axes = ns.plot_2d(params, types=['kde', 'scatter'])
+    assert((~axes.isnull()).sum().sum() == 12)
```
lukashergt: In `test_make_2d_axes` you define `nx = len(paramnames_x)` and use those to determine the numeric value to test for in the assertions. Would probably be more appropriate here, too, right?

anesthetic/plot.py
williamjameshandley: They are, but the ``make_2d_plots`` is a separate set of code, so they don't need to be the same. I think I might prefer to have the default include the scatter plot above the diagonal, since it wastes less space and does provide orthogonal (and unsmoothed) information.

lukashergt: I agree, but maybe it would make sense to change `types` to behave in the same way? 
At the very least it would make the options of `upper` and of `scatter` known to the user and should they wish to exclude them the docs should be straight forward in telling how to turn them off.

williamjameshandley: I've now changed this default in the latest push

tests/test_plot.py
williamjameshandley: I've updated these tests so they're more logical (as you requested in another comment)

tests/test_samples.py
williamjameshandley: This turned out to be pretty challenging in practice to do, so I've re-written those sections of the tests to be more systematic

anesthetic/samples.py
```diff
@@ -271,6 +273,10 @@ def plot_2d(self, axes, *args, **kwargs):
 
                     self.plot(ax_, x, y, plot_type=types[pos], *args,
                               **local_kwargs[pos])
+                elif ax is not None:
+                    ax_ = ax.twin if x == y else ax
+                    ax_.plot([], [])  # need this to increment color cycle
+                    ax_.scatter([], [])  # need this to increment color cycle
```
williamjameshandley: I really like this solution -- I've cleaned up the code a little in the latest push, but this is exactly what was needed

williamjameshandley:
Resolves #16 
Fixes #19 
Fixes #13

anesthetic/samples.py
lukashergt: Haha, yeah, looks neater now.

williamjameshandley: Why do we need to call ``plot`` and ``scatter`` for every loop?

tests/test_samples.py
lukashergt: Looks great, now, and seems to cover many of cases.

williamjameshandley:
(Since it's my original PR, you should do the honours and squash and merge)

anesthetic/samples.py
lukashergt: They have each their own separate colour cycles, i.e. incrementing the colour for `plot` from `'C0'` to `'C1'` won't affect a call to `scatter` afterwards, `scatter` will use `'C0'`.

lukashergt:
@williamjameshandley  And delete the branch, now?

PR #17, opened 2019-07-09
set matplotlib.use('Agg') in test*.py files to avoide ImportError for Tkinter

lukashergt:
# Description

Using `matplotlib.pyplot` when running pytest will throw 
```python
ImportError: No module named Tkinter
```
for many people (particularly people using jupyter notebooks?). All this requires to fix it is to put the lines
```python
import matplotlib
matplotlib.use('Agg')
```
at the head of the `test/test*.py` files.

# Checklist:

### DIFF
--- a/tests/matplotlib_agg.py
+++ b/tests/matplotlib_agg.py
@@ -1,0 +1,3 @@
+import matplotlib
+matplotlib.use('Agg')
+
--- a/tests/matplotlib_agg.py
+++ b/tests/matplotlib_agg.py
@@ -1,3 +1,2 @@
 import matplotlib
 matplotlib.use('Agg')
-
--- a/tests/test_gui.py
+++ b/tests/test_gui.py
@@ -1,3 +1,5 @@
+import matplotlib
+matplotlib.use('Agg')
 from anesthetic import NestedSamples
 
 
--- a/tests/test_gui.py
+++ b/tests/test_gui.py
@@ -1,4 +1,4 @@
-import matplotlib_agg
+import matplotlib_agg  # noqa: F401
 from anesthetic import NestedSamples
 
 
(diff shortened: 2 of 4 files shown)


lukashergt:
Hi Will,
it was driving me crazy that I wasn't able to run tests locally. Can you check whether this fix is ok, i.e. things still work for you?

lukashergt:
Shoot, the github tests now complain that import statements are not at the top of file... But my local tests don't work unless I set the backend to 'Agg' before importing `matplotlib.pyplot`...

williamjameshandley:
>Shoot, the github tests now complain that import statements are not at the top of file... But my local tests don't work unless I set the backend to 'Agg' before importing matplotlib.pyplot...

You can either [try this experimental method](https://stackoverflow.com/questions/39305810/matplotlib-use-required-before-other-imports-clashes-with-pep8-ignore-or-fix) or set the tests in the travis and circle files to ``pyflake --ignore=E402``, since this PEP8 is only advisory.

lukashergt:
The stackoverflow suggestion doesn't work for me as the backend needs to be changed even before `pyplot` is imported...

How/where can I modify the circle and travis files? Do I even have the rights to do so?

williamjameshandley:
The relevant files in in the main directory of the repository itself ``.travis.yaml`` and ``.circleci/config.yml``. Googling 'travis' and 'circleci setup' should give you a quick introduction to continuous integration.

lukashergt:
Actually never mind, the final suggestion actually works. Seems a bit like cheating, but it works...

lukashergt:
Well, it worked for me, but now it complains about `matplotlib_agg` being unsed...

Which PEP8 would you prefer to keep and which one to ignore?
* imports (not) at head of module
* (un)used imports

lukashergt:
Yay, tests have passed.

Using the "# noqa: F401" comment actually allows for all PEP checks to still be active. F401 is only ignored for those lines not permanently.

Issue #22, opened 2019-07-11
inherit documentation of args and kwargs from sub-functions


lukashergt:
Could [docrep](https://docrep.readthedocs.io/en/latest/) be potentially of interest for inheriting documentation of parameters from one function to another? 

This could be the solution to avoiding copy-pasting of docs and to "good docs are better than no docs, but no docs are better than bad docs".

**Example of a use-case:**
Currently `samples.plot_1d`'s docstring only informs about one argument: `axes`. It might be helpful for the user to also see arguments going into `samples.plot`, though. In particular the `plot_type` argument which is documented in `plot` would be useful to know about in `plot_1d`.

williamjameshandley:
This does sound pretty good. It looks like it should be compatible with our sphinx rtd implementation too, since we follow numpy conventions.

1. Are you able to submit a minor PR with an example of it in action for a subsection of the code?
2. Are there any other similar packages? How do numpy, scipy or matplotlib implement their nested documentation?

lukashergt:
1. done

2. `matplotlib` does some docstring interpolation but the documentation on how to do it is rather opaque.

williamjameshandley:
partially resolved by #239

PR #24, opened 2019-07-11
example of `docrep` applied to `samples.plot` and `samples.plot_1d`

lukashergt:
# Description

This addresses issue #22 and for now provides an example  of `docrep` in action:
* `docrep` gets the parameters from the docstring of `samples.plot` and appends them to the parameters section in the docstring of `samples.plot_1d`.

* `docrep` added to requirements.txt and requirements_minimal.txt for this.

# Checklist:

### DIFF
--- a/anesthetic/samples.py
+++ b/anesthetic/samples.py
@@ -6,15 +6,10 @@
 import numpy
 import pandas
 from scipy.special import logsumexp
-from matplotlib.pyplot import hist as plt_hist
-try:
-    from astropy.visualization import hist as astropy_hist
-except ImportError:
-    pass
-from anesthetic.plot import (make_1d_axes, make_2d_axes, plot_1d,
+from anesthetic.plot import (docstrings, make_1d_axes, make_2d_axes, plot_1d,
                              hist_1d, scatter_plot_2d, contour_plot_2d)
 from anesthetic.read.samplereader import SampleReader
-from anesthetic.utils import compute_nlive, docstrings
+from anesthetic.utils import compute_nlive
 from anesthetic.gui.plot import RunPlotter
 from anesthetic.weighted_pandas import WeightedDataFrame
 
@@ -158,13 +153,6 @@ class MCMCSamples(WeightedDataFrame):
                     *args, **kwargs)
 
     docstrings.keep_params('plot.parameters', 'plot_type')
-    docstrings.get_sections(docstrings.dedent(plt_hist).__doc__,
-                            'plt.hist', sections=['Parameters'])
-    docstrings.delete_params('plt.hist.parameters',
-                             'x', 'orientation', 'log', 'stacked', 'normed')
-    docstrings.get_sections(docstrings.dedent(astropy_hist).__doc__,
-                            'astropy.hist', sections=['Parameters'])
-    docstrings.delete_params('astropy.hist.parameters', 'x', 'ax')
 
     @docstrings.dedent
     def plot_1d(self, axes, *args, **kwargs):
@@ -194,11 +182,7 @@ class MCMCSamples(WeightedDataFrame):
         ----------------
         Parameters for plot_type='hist'
         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-        **Parameters from astropy.hist** (if astropy is installed)
-        %(astropy.hist.parameters.no_x|ax)s
-
-        **Parameters from plt.hist**
-        %(plt.hist.parameters.no_x|orientation|log|stacked|normed)s
+        %(hist_1d.other_parameters)s
 
         """
         if not isinstance(axes, pandas.Series):
--- a/anesthetic/samples.py
+++ b/anesthetic/samples.py
@@ -106,8 +106,14 @@ class MCMCSamples(WeightedDataFrame):
             If not provided, or the same as paramname_x, then 1D plot produced.
 
         plot_type: str, optional
-            Must be in {'kde', 'scatter'} for 2D plots and in {'kde', 'hist'}
-            for 1D plots. (Default: 'kde')
+            What type of plot to produce.
+            The options for 1D plots are:
+                - 'kde'
+                - 'hist'
+            The options for 2D plots are:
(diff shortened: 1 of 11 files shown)


anesthetic/samples.py
```diff
@@ -148,6 +150,10 @@ def plot(self, ax, paramname_x, paramname_y=None, *args, **kwargs):
                     xmin=xmin, xmax=xmax, ymin=ymin, ymax=ymax,
                     *args, **kwargs)
 
+    docstrings.delete_params('plot.parameters',
+                             'ax', 'paramname_x', 'paramname_y')
+
+    @docstrings.dedent
```
lukashergt: Instead of deleting parameters from the docstring there is also the option of specifying which ones to keep. Not sure what in this case would seem more appropriate.
1) Keep only `plot_type` meaning that when parameters are added to `plot` they _won't_ automatically be documented in `plot_1d`.
2) Delete `ax`, `paramname_x`, and `paramname_y` meaning that new parameters to `plot` _will_ show up in the documentation of `plot_1d` unless removed.

williamjameshandley: The first case is definitely safer -- It's always better for documentation to be absent than wrong.

williamjameshandley:
I think the clearest use-case in ``anesthetic/samples.py`` would be to avoid the repetition in the ``MCMCSamples`` and ``NestedSamples`` class docstrings. Do you know if ``docrep`` is able to do this? It may not be possible on python2, but that's not really an issue for a modern package

williamjameshandley:
That does seem to work pretty well, although the additional decorators do add some visual clutter when reading the code.

How many places in the documentation are there where this would need to be applied?

williamjameshandley:
It might be nice to rename the rather cryptic ``@docstrings.dedent`` to something more obvious to users, like ``@get_doc`` or something similar.

lukashergt:
> Do you know if `docrep` is able to do this? It may not be possible on python2, but that's not really an issue for a modern package

Yes, this is possible, see [2d76557](https://github.com/williamjameshandley/anesthetic/pull/24/commits/2d76557e09cb9557215b732e621e09717ea2acaf).

Indeed, this doesn't work out of the box in python2. We could make it work by moving parameter docs into the `__init__` docstring or by explicitly creating the `__doc__` attribute of the class, i.e.:

1. using `__init__`
```python
class A:
    """Class doc."""

    @docstrings.dedent
    def __init__():
        """
        Parameters
        ----------
        %(A.parameters)s
        """

```

2. using `__doc__`
```python
class A:
    __doc__ = docstrings.dedents(
    """Class doc.
    
    Parameters
    ----------
    %(A.parameters)s
    """)

```

lukashergt:
> It might be nice to rename the rather cryptic `@docstrings.dedent` to something more obvious to users, like `@get_doc` or something similar.

`dedent` is handling indentation not getting the code. `get_sectionsf` is getting the `Parameters`, `Results`, `Other Parameters` sections...

But sure, we could rename `docstrings`. E.g. to `doc_inheritance`?

williamjameshandley:
I agree that the syntax makes sense, but it will be confusing for first-time readers. I think docstrings is OK. I was more imagining just a quick shortcut like ``doc_inheritance=docstrings.dedent`` at the top of the file, underneath where ``docstrings`` is first instantiated

williamjameshandley:
I don't like the idea of modifying ``__doc__``, as that's very ugly. I'd be happier just not to support python2, given its [impending deprecation](https://pythonclock.org/)

lukashergt:
> I'd be happier just not to support python2

Thought you might. That's why I did my last commit that way. I think that's fine too. Sucks for MontePython users as they are still forced to use python2 but that's on MontePython...

williamjameshandley:
To be fair to us, we're still supporting python2, it's just that the in-place documentation won't be quite as complete, although it will be on readthedocs.

lukashergt:
Added an example of how to inherit from a function such as `plt.hist`. What do you think?

lukashergt:
> How many places in the documentation are there where this would need to be applied?

Focus should be on the user functions of which many don't need to inherit anything.
The user functions inheritance tree is:
- [x] `MCMCSamples`

- [x] `NestedSamples`
  - [x] inherits from `MCMCSamples`

- [ ] `MCMCSamples.plot_1d`
  - [x] inherits from `MCMCSamples.plot`
  - [ ] inherits from `plot_1d`
    - [ ] inherits from `plt.plot`
  - [x] inherits from `hist_1d`
    - [x] inherits from `astropy.hist`
    - [x] inherits from `plt.hist`

- [ ] `MCMCSamples.plot_2d`
  - [ ] inherits from `contour_plot_2d`
    - [ ] inherits from `plt.contourf`
    - [ ] inherits from `plt.contour`
  - [ ] inherits from `scatter_plot_2d`
    - [ ] inherits from `plt.scatter`

- [x] `make_1d_axes`

- [x] `make_2d_axes`

- [x] `NestedSamples.ns_output`

- [x] `NestedSamples.set_beta`

williamjameshandley:
Incidentally, it does look like matplotlib uses docrep in parts, so it's definitely consistent with the ethos of this anesthetic in trying to keep things as close to matplotlib as possible.

williamjameshandley:
This is quite an old PR, that is now a very long way behind master. @lukashergt do you think the aims of this PR are still feasible/desirable in light of the other updates to anesthetic since this was originally submitted?

lukashergt:
This is something I think I would have found useful as a new user. By now I am too familiar with the usage that I don't open the (quick-)docs anymore, so it's harder to judge the desirability, now.

If I remember correctly this was done and working back then, but it wasn't clear whether we actually want it. It would be another dependency that maybe we don't want.

Now, it would require an extensive overhaul. If we want it, I'd be happy to look through this, but I don't have any problems with this just being closed either.

williamjameshandley:
I think that we have higher-priority things to work on for anesthetic at the moment, so I suggest that @lukashergt closes this PR for now (we can always reopen it), but leaves #22 open as an issue.

Issue #4, opened 2019-04-17
Tricontour plots


williamjameshandley:
Matplotlib now has a [contour plotting on a random grid](https://matplotlib.org/api/_as_gen/matplotlib.pyplot.tricontour.html#matplotlib.pyplot.tricontour).

This is in some sense optimal for our purposes, as then all one needs to compute for 2D probability mass function plots is the *ordering* of the points in probability.

PR #25, opened 2019-07-12
Preliminary tricontour functionality

williamjameshandley:
# Description

In the interests of providing more than one kernel density estimator, I have
now implemented ``scipy.stats.gaussian_kde``

In order to make this usably fast, I have also implemented 
``matplotlib.pyplot.tricontour`` plots, which plots a contour plot on an uneven
grid of points, using the sample values as those points

The resulting kdes are messier than when using fastkde, but are more honest.

Addresses #4

# Checklist:


anesthetic/kde.py
```diff
@@ -104,3 +106,92 @@ def kde_2d(d_x, d_y, xmin=None, xmax=None, ymin=None, ymax=None):
...
+        p = p[x <= xmax]
+        x = x[x <= xmax]
+
+    return x, p
+
+
+def kde_2d(d_x, d_y, xmin=None, xmax=None, ymin=None, ymax=None):
+    """Perform a two-dimensional kernel density estimation.
+
+    Wrapper round scipy.stats.gaussiankde.
```
lukashergt: underscore missing in gaussian_kde?

lukashergt:
The mirroring at the limit edges still works for `'fastkde'` but not for `'kde'`, is that something to do?

williamjameshandley:
```python
fig, axes = make_2d_axes(['x0', 'x1', 'x2', 'x3'])
ns.plot_2d(axes, types={'upper':'kde','diagonal':'kde'})
ns.plot_2d(axes, types={'lower':'fastkde','diagonal':'fastkde'})
proxy = get_legend_proxy(fig)
fig.legend([proxy[0], proxy[3]], ['kde', 'fastkde'])
```

tricontour

This example reveals a few potential bugs:
1. The kde code in the top row doesn't have its y-axis synced with the width of the ``x0`` x-axis
2. The get_legend_proxy now shows far more than two separate lines
3. The ``fastkde`` and ``kde`` implementations have differing widths. This could be because fastkde is obviously a lot more smoothed, and smoothing does tend to broaden a distribution, or it could be down to a bug in my implementation. Either way, it warrants investigation.

lukashergt:
It might be useful to change the example data such that one parameter is truncated, e.g. sample normal distribution with mu=-1 and sigma=1 and truncate at origin. And then make sure that limits are set correctly, but I guess they are all set at 0 and 1 already. 

Maybe actually useful to have one parameter where no limits are set?

Maybe also useful to have another parameter sampled from a uniform distribution?

anesthetic/plot.py
```diff
@@ -319,8 +366,8 @@ def hist_1d(ax, data, *args, **kwargs):
     return bars
 
 
-def contour_plot_2d(ax, data_x, data_y, *args, **kwargs):
-    """Plot a 2d marginalised distribution as contours.
+def grid_kde_plot_2d(ax, data_x, data_y, *args, **kwargs):
+    """Plot a 2d marginalised distribution as contours on a grid.
```
lukashergt: I don't like the name change here. When I'm looking for this function (scrolling or searching) I'm looking for "contour" as this is how I know it from matplotlib. I'd suggest something along the lines `contour_plot_2d_grid_kde` and `contour_plot_2d_sample_kde`.

williamjameshandley:
> The mirroring at the limit edges still works for 'fastkde' but not for 'kde', is that something to do?

Yes. I need to think carefully about this, since the current method has ~O(N^2) scaling, where N is the number of compressed samples. For N~1000, it takes about 30ms per plot, but this can obviously add an overhead if you're creating a lot of reflections.

The reflections bit of the code was always a bit hacky, and it would be better to have a more proper procedure like getdist does

anesthetic/plot.py
```diff
@@ -216,7 +216,7 @@ def make_2d_axes(params, **kwargs):
     return fig, axes
 
 
-def grid_kde_plot_1d(ax, data, *args, **kwargs):
+def contour_plot_1d_grid_kde(ax, data, *args, **kwargs):
```
lukashergt: "contour" only makes sense in the 2D case when it actually plots a contour plot, for 1D I'd leave it at `grid_kde_plot_1d` or flip it around for consistency `plot_1d_grid_kde`. Or `plot_plot_1d_grid_kde` if we go by the matplotlib names, but that sound stupid. Then I'd rather go for `line_plot_1d_grid_kde`.

williamjameshandley:
OK, I've fixed the width issue (3) and #30 has fixed the legend issue (2), Lack of syncing of axes still remains

```python
                                                                                                                                        
fig, axes = make_2d_axes(['x0', 'x1', 'x2', 'x3'])                                                                                      
ns.plot_2d(axes, types={'upper':'kde','diagonal':'kde'}, label='kde')                                                                   
ns.plot_2d(axes, types={'lower':'fastkde','diagonal':'fastkde'}, label='fastkde')                                                       
patch0, label0 = axes['x0']['x1'].get_legend_handles_labels()                                                                           
patch1, label1 = axes['x1']['x0'].get_legend_handles_labels()                                                                           
fig.legend(patch0 + patch1, label0 + label1)                                                                                            
```
![Figure_1](https://user-images.githubusercontent.com/3708486/61373886-ea3bf400-a892-11e9-989d-d2e650de06c6.png)

anesthetic/plot.py
```diff
@@ -488,8 +488,7 @@ def contour_plot_2d_sample_kde(ax, data_x, data_y, *args, **kwargs):
                        ymin=ymin, ymax=ymax)
     pdf /= pdf.max()
     p = sorted(pdf.flatten())
-    m = numpy.cumsum(p)
-    m /= m[-1]
+    m = numpy.arange(1, len(p)+1)/(len(p)+1)
```
lukashergt: Why does this need to be different from `grid_kde`?

williamjameshandley: This was the subtle bug that was causing the difference in 2D plot widths between kde and fastkde. Samples from the distribution (as in `contour_plot_2d_sample_kde`) distribute uniformly across the probability range. Gridded point (as in `contour_plot_2d_grid_kde`) need to be `cumsum`ed across the range.

anesthetic/plot.py
williamjameshandley: Changed to ``line_plot_*_kde`` in 3863a3a

anesthetic/plot.py
williamjameshandley: resolved in ec34419

anesthetic/plot.py
lukashergt: Ahh, thanks, that makes perfect sense.

williamjameshandley:
Replaced by #50

PR #26, opened 2019-07-12
Pip caching and parallelisation to speed up travis builds

williamjameshandley:
The build matrix is large, and testing times are now getting quite long. It
would be nice if we didn't burn a lot of unecessary time installing via pip
over and over again

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,2 @@
+[run]
+omit =anesthetic/gui/_matplotlib.py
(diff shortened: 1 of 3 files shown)


williamjameshandley:
It now spends the same amount of time preparing the build as testing (previously preparation was taking longer). I've also disabled testing branches by default, so travis only runs once for any PR, even if pushed to the main repository.

williamjameshandley:
This now looks to be as fast as its going to get (I've now added parallel testing).

williamjameshandley:
@lukashergt , if we merge this in now, it will speed up any current open pull requests.

williamjameshandley:
It should update if it's out-of-date, but we can always clear it manually if it's clear something is awry with the install.

PR #28, opened 2019-07-12
Fix types

lukashergt:
# Description

* Fixes problem No.1 in #13 but not yet No.2.
  - This means requesting fewer plots with `types` in a second call to
    `plot_2d` is now possible.
  - This fix could be probably further simplified.

# Checklist:

### DIFF
--- a/anesthetic/samples.py
+++ b/anesthetic/samples.py
@@ -145,8 +145,8 @@ class MCMCSamples(WeightedDataFrame):
                 raise NotImplementedError("plot_type is '%s', but must be in "
                                           "{'kde', 'scatter'}." % plot_type)
 
-            if ((paramname_x in self and paramname_y in self)
-               and plot_type is not None):
+            if (paramname_x in self and paramname_y in self
+                    and plot_type is not None):
                 x = self[paramname_x].compress(nsamples)
                 y = self[paramname_y].compress(nsamples)
             else:
--- a/anesthetic/samples.py
+++ b/anesthetic/samples.py
@@ -122,10 +122,10 @@ class MCMCSamples(WeightedDataFrame):
                 plot = plot_1d
             elif plot_type == 'hist':
                 plot = hist_1d
-            else:
+            elif plot_type is not None:
                 raise NotImplementedError("plot_type is '%s', but must be in "
                                           "{'kde', 'hist'}." % plot_type)
-            if paramname_x in self:
+            if paramname_x in self and plot_type is not None:
                 x = self[paramname_x].compress()
             else:
                 x = []
@@ -142,11 +142,12 @@ class MCMCSamples(WeightedDataFrame):
             elif plot_type == 'scatter':
                 nsamples = 500
                 plot = scatter_plot_2d
-            else:
+            elif plot_type is not None:
                 raise NotImplementedError("plot_type is '%s', but must be in "
                                           "{'kde', 'scatter'}." % plot_type)
 
-            if paramname_x in self and paramname_y in self:
+            if ((paramname_x in self and paramname_y in self)
+               and plot_type is not None):
                 x = self[paramname_x].compress(nsamples)
                 y = self[paramname_y].compress(nsamples)
             else:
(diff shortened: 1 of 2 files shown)


williamjameshandley:
Does my latest push achieve the same effect? I really want to avoid the manual skipping of cycles -- it won't be easy to maintain as more types of plots are added, and could have odd side effects.

williamjameshandley:
It would be good if you could add a test to test this behaviour (even if it's just the example code from issue #13).

anesthetic/samples.py
```diff
@@ -142,11 +142,12 @@ def plot(self, ax, paramname_x, paramname_y=None, *args, **kwargs):
...
                 nsamples = 500
                 plot = scatter_plot_2d
-            else:
+            elif plot_type is not None:
                 raise NotImplementedError("plot_type is '%s', but must be in "
                                           "{'kde', 'scatter'}." % plot_type)
 
-            if paramname_x in self and paramname_y in self:
+            if ((paramname_x in self and paramname_y in self)
+               and plot_type is not None):
```
lukashergt: are the extra brackets needed?

williamjameshandley: Yes, because of  the line break.

anesthetic/samples.py
```diff
@@ -271,17 +272,19 @@ def plot_2d(self, axes, *args, **kwargs):
         else:
             fig = axes.values[~axes.isna()][0].figure
 
+        for pos in default_types:
+            if pos not in types:
+                types[pos] = None
+            if pos not in local_kwargs:
+                local_kwargs[pos] = {}
```
lukashergt: If we are doing this change in handling types, would it make sense to do so instantly after `default_types` and adjusting all conditionals from `pos in types` to `types[pos] is not None`?

anesthetic/samples.py
lukashergt: I mean the other set of brackets...

anesthetic/samples.py
williamjameshandley: That could be quite nice. You could also remove the rather clunky  filling in of the dictionary by using a [defaultdict](https://docs.python.org/2/library/collections.html#collections.defaultdict).

anesthetic/samples.py
williamjameshandley: Ah! No they're not. Well spotted.

lukashergt:
> Does my latest push achieve the same effect?

Wow, no it does not. How can the tests pass?! 

At the moment it will raise the error:
```python
UnboundLocalError                         Traceback (most recent call last)
<ipython-input-8-c2b481c7a4a2> in <module>()
      1 ns = NestedSamples(root="/scratch/Projects/AnestheticPrj/anesthetic/tests/example_data/pc")
      2 fig, axes = ns.plot_2d(['x0', 'x1', 'x2'])
----> 3 ns.plot_2d(axes, types={'diagonal': 'hist'});

/scratch/.virtualenvs/py27env/lib/python2.7/site-packages/anesthetic/samples.pyc in plot_2d(self, axes, *args, **kwargs)
    285                     ax_ = ax.twin if x == y else ax
    286                     self.plot(ax_, x, y, plot_type=types[pos], *args,
--> 287                               **local_kwargs[pos])
    288 
    289         return fig, axes

/scratch/.virtualenvs/py27env/lib/python2.7/site-packages/anesthetic/samples.pyc in plot(self, ax, paramname_x, paramname_y, *args, **kwargs)
    155                 y = []
    156 
--> 157             return plot(ax, x, y, xmin=xmin, xmax=xmax, ymin=ymin, ymax=ymax,
    158                         *args, **kwargs)
    159 

UnboundLocalError: local variable 'plot' referenced before assignment
```

Because `plot_type` is set to `None` the return variable `plot` has not been assigned. To get assigned `plot_type` needs to be either `'kde'` or `'scatter'`.

anesthetic/samples.py
lukashergt: not sure how to use defaultdict here

williamjameshandley:
> Wow, no it does not. How can the tests pass?!

It's why it's a good idea to add a test that replicates it, even if it seems trivial to check on your machine.

> not sure how to use defaultdict here

Turns out defaultdict is completely unnecessary here, you can just use a normal dicts `get` function with a default value.

anesthetic/samples.py
```diff
@@ -260,16 +261,12 @@ def plot_2d(self, axes, *args, **kwargs):
 
         local_kwargs = {pos: kwargs.pop('%s_kwargs' % pos, {})
                         for pos in default_types}
-        for pos in default_types:
-            local_kwargs[pos].update(kwargs)
-            if pos not in types:
-                types[pos] = None
```
lukashergt: The `local_kwargs[pos].update` seems missing now, right?

williamjameshandley: Ah yes, I think that was a slightly overzealous edit on my part. That should be restored

williamjameshandley:
@lukashergt does this now fully resolve #13

lukashergt:
> @lukashergt does this now fully resolve #13 ?

It resolves the bug in #13, not the feature. 
But since we probably won't pursue the feature I guess #13 can be closed, right?

williamjameshandley:
Closes #13

PR #30, opened 2019-07-14
Make legend handling more intuitive and consistent with `matplotlib`

lukashergt:
# Description

Legend handling is now more intuitive, i.e. can be done in the usual matplotlib fashion.

* Make the link `axes[x][x] = axes[x][x].twin` such that the user has direct access to the diagonal axes that have actually been drawn in.
  - `axes[x][x].legend()` now automatically makes a legend in the diagonal axes.

* Create the legend proxy already in `contour_plot_2d` and pass it to `ax.patches`.
  - `axes[x][y].legend()` now automatically makes a legend in the 2D 'kde' axes.
  - No need for `get_legend_proxy` anymore, because matplotlib's own functions can handle this, now:
  ```python
  handles, labels = axes[x][y].get_legend_handles_labels()
  leg = fig.legend(handles, labels)
  ```
  - Flagged `get_legend_proxy` as deprecated.

# Checklist:

### DIFF
--- a/anesthetic/plot.py
+++ b/anesthetic/plot.py
@@ -351,6 +351,7 @@ def contour_plot_2d(ax, data_x, data_y, *args, **kwargs):
     xmax = kwargs.pop('xmax', None)
     ymin = kwargs.pop('ymin', None)
     ymax = kwargs.pop('ymax', None)
+    label = kwargs.pop('label')
     color = kwargs.pop('color', next(ax._get_lines.prop_cycler)['color'])
     if len(data_x) == 0 or len(data_y) == 0:
         return numpy.zeros(0), numpy.zeros(0), numpy.zeros((0, 0))
@@ -389,6 +390,9 @@ def contour_plot_2d(ax, data_x, data_y, *args, **kwargs):
     ax.contour(x[i], y[j], pdf[numpy.ix_(j, i)], contours,
                vmin=0, vmax=1.2, linewidths=0.5, colors='k', zorder=zorder+2,
                *args, **kwargs)
+    ax.patches += [plt.Rectangle((0, 0), 1, 1, fc=cmap(0.999), ec=cmap(0.33),
+                                 lw=2, label=label)]
+
     ax.set_xlim(*check_bounds(x[i], xmin, xmax), auto=True)
     ax.set_ylim(*check_bounds(y[j], ymin, ymax), auto=True)
     return cbar
--- a/anesthetic/plot.py
+++ b/anesthetic/plot.py
@@ -351,7 +351,7 @@ def contour_plot_2d(ax, data_x, data_y, *args, **kwargs):
     xmax = kwargs.pop('xmax', None)
     ymin = kwargs.pop('ymin', None)
     ymax = kwargs.pop('ymax', None)
-    label = kwargs.pop('label')
+    label = kwargs.pop('label', None)
     color = kwargs.pop('color', next(ax._get_lines.prop_cycler)['color'])
     if len(data_x) == 0 or len(data_y) == 0:
         return numpy.zeros(0), numpy.zeros(0), numpy.zeros((0, 0))
--- a/anesthetic/plot.py
+++ b/anesthetic/plot.py
@@ -352,6 +352,7 @@ def contour_plot_2d(ax, data_x, data_y, *args, **kwargs):
     ymin = kwargs.pop('ymin', None)
     ymax = kwargs.pop('ymax', None)
     label = kwargs.pop('label', None)
+    zorder = kwargs.pop('zorder', 1)
     color = kwargs.pop('color', next(ax._get_lines.prop_cycler)['color'])
     if len(data_x) == 0 or len(data_y) == 0:
         return numpy.zeros(0), numpy.zeros(0), numpy.zeros((0, 0))
@@ -380,12 +381,12 @@ def contour_plot_2d(ax, data_x, data_y, *args, **kwargs):
 
     cmap = basic_cmap(color)
 
-    cbar = ax.contourf(x[i], y[j], pdf[numpy.ix_(j, i)], contours,
-                       vmin=0, vmax=1.0, cmap=cmap, *args, **kwargs)
+    cbar = ax.contourf(x[i], y[j], pdf[numpy.ix_(j, i)], contours, cmap=cmap,
+                       zorder=zorder, vmin=0, vmax=1.0, *args, **kwargs)
     for c in cbar.collections:
         c.set_cmap(cmap)
 
-    ax.contour(x[i], y[j], pdf[numpy.ix_(j, i)], contours,
+    ax.contour(x[i], y[j], pdf[numpy.ix_(j, i)], contours, zorder=zorder,
                vmin=0, vmax=1.2, linewidths=0.5, colors='k', *args, **kwargs)
     ax.patches += [plt.Rectangle((0, 0), 1, 1, fc=cmap(0.999), ec=cmap(0.33),
                                  lw=2, label=label)]
--- a/anesthetic/plot.py
+++ b/anesthetic/plot.py
@@ -184,6 +184,10 @@ def make_2d_axes(params, **kwargs):
                     axes[x][y].twin.set_yticks([])
                     axes[x][y].twin.set_ylim(0, 1.1)
                     axes[x][y].set_zorder(axes[x][y].twin.get_zorder() + 1)
+                    axes[x][y].lines = axes[x][y].twin.lines
+                    axes[x][y].patches = axes[x][y].twin.patches
+                    axes[x][y].collections = axes[x][y].twin.collections
+                    axes[x][y].containers = axes[x][y].twin.containers
                     axes[x][y].position = 'diagonal'
                 elif position[x][y] == 1:
                     axes[x][y].position = 'upper'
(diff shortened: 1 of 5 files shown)


anesthetic/plot.py
```diff
@@ -388,6 +390,9 @@ def contour_plot_2d(ax, data_x, data_y, *args, **kwargs):
     ax.contour(x[i], y[j], pdf[numpy.ix_(j, i)], contours,
                vmin=0, vmax=1.2, linewidths=0.5, colors='k', zorder=zorder+2,
                *args, **kwargs)
+    ax.patches += [plt.Rectangle((0, 0), 1, 1, fc=cmap(0.999), ec=cmap(0.33),
+                                 lw=2, label=label)]
+
```
williamjameshandley: This is very neat if this works

anesthetic/samples.py
```diff
@@ -281,6 +281,12 @@ def plot_2d(self, axes, *args, **kwargs):
                     lkwargs = local_kwargs.get(pos, {})
                     self.plot(ax_, x, y, plot_type=plot_type, *args, **lkwargs)
 
+        for y, row in axes.iterrows():
+            for x, ax in row.iteritems():
+                if x == y and ax is not None:
+                    axes[x][y] = axes[x][y].twin
+                    axes[x][y].twin = axes[x][y]
+
```
williamjameshandley: This doesn't swap the axes, if that's what you're trying to do. The pythonic standard would be something like 
```python
a, b = b, a
```

williamjameshandley: It might also to do this in the ``make_2d_axes`` code, rather than in the samples code, and to do it at the point of construction. Line 279 in this file could then be removed.

lukashergt: > This doesn't swap the axes, if that's what you're trying to do.

I wasn't. Do we need access to the empty axis?

> do this in the `make_2d_axes` code

Briefly tried that, but something didn't work. Wasn't very persistent, though.

williamjameshandley: > I wasn't. Do we need access to the empty axis?

I'm not sure we do. I don't think we need it to be user-accessible. I can think of use-cases where we use it to ensure that y-axes limits in the top-row are synced with the x-axes limit (see #25), but in that case, we could restore this functionality under a 'dangerous' underscored variable. 

If we're dropping twinning completely, the changes should definitely be in the ``make_2d_axes`` part.

anesthetic/plot.py
```diff
@@ -388,6 +390,9 @@ def contour_plot_2d(ax, data_x, data_y, *args, **kwargs):
     ax.contour(x[i], y[j], pdf[numpy.ix_(j, i)], contours,
                vmin=0, vmax=1.2, linewidths=0.5, colors='k', zorder=zorder+2,
                *args, **kwargs)
+    ax.patches += [plt.Rectangle((0, 0), 1, 1, fc=cmap(0.999), ec=cmap(0.33),
```
lukashergt: What's the reason for the `+1` and `+2` to `zorder` in the contour plots? 

The way they are determined from the axes children they behave strangely when there is also a legend or another plot in that axis.

williamjameshandley: Originally I think it was to make sure that the respective colours were on top (particularly with regards to samples), but it might be more consistent let them be handled automatically, and give the user charge of ``zorder`` vias the ``*args, **kwargs``. Does removing them give any immediately odd behaviour?

anesthetic/samples.py
lukashergt: Hmm, labels and ticks are handled in the untwinned axes, though... We do want user access to them, right?

lukashergt: Changed the behaviour in [9f00b05](https://github.com/williamjameshandley/anesthetic/pull/30/commits/9f00b05ffdb58b46c4b032a07fa396dde0205d60). Keep the twins the way they are, but pass on lines and patches from one to the other.

williamjameshandley: OK. I think this is a nice and consistent solution. This is the reason that I had the 'twin' approach, in that it makes sense for the 1D axis to be distinct from, and plotted over the 2D axis that forms part of the grid. Passing the lines and patches is a nice solution, assuming it works. Is there anything else that should be passed over?

anesthetic/plot.py
lukashergt: added tests to verify

lukashergt:
> Could you also add a test that creates a legend using the new functionality?

Added a test that checks the handles and labels. Is that what you had in mind?

anesthetic/samples.py
lukashergt: The complete set of available handles would be {ax.lines, ax.patches, ax.collections, ax.containers}, but with `plot` and `hist` we only get the first two, so I only passed those. Would you prefer passing the full list?

anesthetic/plot.py
lukashergt: Ok, the odd thing is that the contour lines now always show, even with multiple contours on the same axis. This is because the default `zorder` for `conourf` is 1 and for `contour` is 2. 

I think kwarg popping zorder and setting the default to 1 is the better thing to do here, though.

anesthetic/samples.py
williamjameshandley: Unless it causes issues, it's probably best to pass them all. There may be extensions/use case that we haven't thought of which require the other two.

anesthetic/plot.py
williamjameshandley: Ah, I knew there was a reason I did this. For our case, it is definitely correct to override this default.

williamjameshandley:
>Added a test that checks the handles and labels. Is that what you had in mind?

That looks great.

williamjameshandley:
One further comment -- If I try calling ``fig.legend``, it produces a large legend for every axis. Is this the expected/desired behaviour?

anesthetic/samples.py

lukashergt:
> One further comment -- If I try calling `fig.legend`, it produces a large legend for every axis. Is this the expected/desired behaviour?

Yes, that is the matplotlib behaviour. This happens when you call the _figure_ legend instead of the _axis_ legend. The figure contains _all_ the axes and thus shows _all_ the handles and labels from _all_ the axes.

williamjameshandley:
>Yes, that is the matplotlib behaviour. This happens when you call the figure legend instead of the axis legend. The figure contains all the axes and thus shows all the handles and labels from all the axes.

So how should we advise users to implement a 'figure legend' in the natural way, namely producing a summary that is not a repeat for each sub-plot?

lukashergt:
> So how should we advise users to implement a 'figure legend' in the natural way, namely producing a summary that is not a repeat for each sub-plot?

See description at the top. I also put this in the deprecation warning for `get_legend_proxy`.

anesthetic/plot.py
```diff
@@ -183,6 +183,7 @@ def make_2d_axes(params, **kwargs):
                     axes[x][y].twin = axes[x][y].twinx()
                     axes[x][y].twin.set_yticks([])
                     axes[x][y].twin.set_ylim(0, 1.1)
+                    axes[x][y].set_zorder(axes[x][y].twin.get_zorder() + 1)
```
williamjameshandley: Why is this necessary?

lukashergt: > * Set axes in front of axes.twin.
>   - Since legends are drawn on axes, they got hidden behind the lines and histograms which belonged to the twin axes and thus were in front.

The twin axes get drawn after the original axes, but the legends are drawn on the original axes and will be hidden behind the twin axes. Actually anything added to those axes will dissappear behind the the twin axes, e.g. if someone were to try to annotate something they'd have a hard time without knowing about the twin.

The change just puts the original in front of the twin.

williamjameshandley: Sounds reasonable

anesthetic/samples.py
lukashergt: Ha, maybe...

lukashergt: yes, seems to work, very nice

williamjameshandley:
I'm trying to update the demo.py and demo.ipynb with your changes

```python
from anesthetic import get_legend_proxy                                                          
prior = nested.set_beta(0)                                                                       
fig, axes = prior.plot_2d(['ns','tau'])                                                          
nested.plot_2d(axes=axes)                                                                        
handles, labels = axes['ns']['tau'].get_legend_handles_labels()                                  
leg = fig.legend(handles, labels)
```
example

But no legend is produced (only an empty top-right box). The handles and labels lists are empty. Do you also get this behaviour?

lukashergt:
> I'm trying to update the demo.py and demo.ipynb with your changes
> 
> 
> But no legend is produced (only an empty top-right box). The handles and labels lists are empty. Do you also get this behaviour?

Are you sure they are using this branch?

lukashergt:
Never mind, you are not passing any labels. Do this:
```python
prior = nested.set_beta(0)
fig, axes = prior.plot_2d(['x0','x1'], label='prior')                                                          
nested.plot_2d(axes=axes, label='posterior')
handles, labels = axes['x0']['x1'].get_legend_handles_labels()
leg = fig.legend(handles, labels)
```

williamjameshandley:
Cool. That seems to have worked. I'll update the plot gallery once version 1.1 is out.

lukashergt:
You don't care about PEP8 in the demos? (`flake8 demo.py`)

williamjameshandley:
No, demo.py is for generating the ipython notebook using [py2nb](https://pypi.org/project/py2nb/0.0.4/). flake8ing it is overkill.

williamjameshandley:
(by all means propose a PR that fixes it)

PR #21, opened 2019-07-11
Re-write of colour fix code

williamjameshandley:
@lukashergt: is there any reason why this wouldn't fix #19 in the same manner as #15?

I think this way of doing it is slightly more logical, and will be easier for downstream changes to integrate.

# Checklist:

- [x] I have performed a self-review of my own code
- [x] My code is PEP8 compliant (`flake8 anesthetic tests`)
- [x] My code contains compliant docstrings (`pydocstyle --convention=numpy anesthetic`)
- [x] New and existing unit tests pass locally with my changes (`python -m pytest`)
- [x] I have added tests that prove my fix is effective or that my feature works

PR #32, opened 2019-07-14
Fix colorcycle

### LINKED ISSUE #31: wrong colour when second model plots into empty axes
### Describe the bug

Similar to #19 when plotting 2 models it sometimes happens that the colours for the 2nd model aren't consistent due to the colour cycle not being properly updated in the 1st call.

This was introduced through #21 by switching from direct calls to `ax.plot([], [])` to only passing `x=[]` (but only to _one_ function).

### To Reproduce

```python
ns1 = NestedSamples(root="./tests/example_data/pc")
ns2 = NestedSamples(root="./tests/example_data/pc")
fig, axes = make_2d_axes(['x0', 'x1', 'x2', 'x3'])
ns1.plot_2d(axes=axes, types={'diagonal': None, 'lower': 'kde', 'upper': 'kde'}, label='foo')
ns2.plot_2d(axes=axes, types={'diagonal': 'kde', 'lower': 'kde', 'upper': 'scatter'}, diagonal_kwargs={'ls': '--'}, upper_kwargs={'alpha': 0.3}, label='bar')
```

image

### Expected behavior

In the above image:
* the dashed lines in the diagonal plots should be orange
* the scatter plots in the upper plots should be orange

### Additional context

* In the diagonal case this must be due to somehow nothing being drawn in the 1st call where `diagonal=None`.

* In the 2D case this has to do with `ax.plot` and `ax.scatter` not sharing the same colour cycle, so **both** have to be called **every time**.

lukashergt:
# Description

In #21 the calls of `ax.plot([], [])` and similar were changed to setting `x = []` and passing it to the plotting functions. This does not take care of the different colour cycles of plot and scatter and thus does not work. 
Thus, here I am reverting some of the changes from #21. 

It might be good to write a test for this where it actually checks the colours used. This might be much easier once #30 is merged.

Fixes #31 

# Checklist:

### DIFF
--- a/anesthetic/samples.py
+++ b/anesthetic/samples.py
@@ -118,30 +118,34 @@ class MCMCSamples(WeightedDataFrame):
 
         if do_1d_plot:
             xmin, xmax = self._limits(paramname_x)
-            if plot_type == 'kde' or plot_type is None:
+            if plot_type == 'kde':
                 plot = plot_1d
             elif plot_type == 'hist':
                 plot = hist_1d
+            elif plot_type is None:
+                ax.plot([], [])
             else:
                 raise NotImplementedError("plot_type is '%s', but must be in "
                                           "{'kde', 'hist'}." % plot_type)
             if paramname_x in self and plot_type is not None:
                 x = self[paramname_x].compress()
-            else:
-                x = []
-
-            return plot(ax, x, xmin=xmin, xmax=xmax, *args, **kwargs)
+                return plot(ax, x, xmin=xmin, xmax=xmax, *args, **kwargs)
 
         else:
             xmin, xmax = self._limits(paramname_x)
             ymin, ymax = self._limits(paramname_y)
 
-            if plot_type == 'kde' or plot_type is None:
+            if plot_type == 'kde':
                 nsamples = None
                 plot = contour_plot_2d
+                ax.scatter([], [])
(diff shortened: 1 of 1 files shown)


anesthetic/samples.py
```diff
@@ -118,30 +118,34 @@ def plot(self, ax, paramname_x, paramname_y=None, *args, **kwargs):
 
         if do_1d_plot:
             xmin, xmax = self._limits(paramname_x)
-            if plot_type == 'kde' or plot_type is None:
+            if plot_type == 'kde':
                 plot = plot_1d
             elif plot_type == 'hist':
                 plot = hist_1d
+            elif plot_type is None:
+                ax.plot([], [])
```
williamjameshandley: Does this take care of the colours for both hist and kde, or do you need a case for each like in the 2d code below

lukashergt: No, hist and plot share the same colour cycle, so it must be only one call here. Consequently the `x=[]` method would have more chance to work in the 1D case. However, also the diagonal plots didn't quite work before as issue #31 shows.

williamjameshandley: That does make sense. I think this is too fiddly to fix the way that I was trying to do it in #21. This fix seems sensible enough, and is not any more lines. If it gets nasty further down the line then we can consider updating.

PR #34, opened 2019-07-15
# Non-trivial example for testing

williamjameshandley:
Following the request from #25, a new set of examples are provided which:

- Constrains x0 and x1
- x0 and x1 are correlated
- x2 is upper bounded
- x3 is unconstrained

This example more thoroughly visually tests the functionality of the code

# Checklist:


williamjameshandley:
```python
mcmc = MCMCSamples(root='./tests/example_data/gd')
mcmc.plot_2d(['x0', 'x1', 'x2', 'x3'])
```
mcmc

Seems to produce the desired behaviour.

williamjameshandley:
Although it's worth noting that it seems to be doing something a little funny for the `x3` parameter of the nested sampling example:

```python
ns = MCMCSamples(root='./tests/example_data/pc
ns.plot_2d(['x0', 'x1', 'x2', 'x3'])
```
ns
')

lukashergt:
* A tophat might be another interesting special case when evaluating the performance of the KDE.

* For the bounded Gaussian I would personally shift it further out such that we see what happens at the boundary even when the derivative at the boundary is non-zero. Especially relevant with the mirroring technique which enforces the derivative to be zero on the boundary.

williamjameshandley:
Following #39 I have made the limits less trivial, in that there is now an extra 'slanted top-hat' parameter between 2 and 4, and the first two parameters are zero-centered. I also increased the number of live points to 125, which is the default recommendation for a 5D space

```python
ns.plot_2d(['x0', 'x1', 'x2', 'x3', 'x4'])
```
Figure_1

This demonstrates the issues that need to be resolved
1. It does not resolve the `[2,4]` hard limits on `x4` correctly
2. It has zero-gradient for `x3`, despite the fact that this is now an exponential distribution

williamjameshandley:
OK, I've now added a script which allows one to check the analytical contours for this case

Figure_1

This demonstrates the border issues fairly clearly.

PR #43, opened 2019-07-25
Automatic labelling

### LINKED ISSUE #42: Automatic labelling
@lukashergt latest fix for figure legends #30 means that we now have matplotlib-like handles for creating figure legends. However, these are only applied if you supply a label keyword argument to the plot. It might be better if this was automatically generated/could be set as a member variable for MCMCSamples, so that users didn't need to remember to supply labels to generate plots.

williamjameshandley:
# Description

Added a `label` attribute to `MCMCSamples`, which provides a default for the plots. Labels are defined to be the basename of the root if loaded from file, None otherwise.

Fixes #42

# Checklist:

### DIFF
--- a/anesthetic/samples.py
+++ b/anesthetic/samples.py
@@ -43,10 +43,13 @@ class MCMCSamples(WeightedDataFrame):
         loglikelihoods of samples.
 
     tex: dict
-        mapping from coloumns to tex labels for plotting
+        mapping from columns to tex labels for plotting
 
     limits: dict
-        mapping from coloumns to prior limits
+        mapping from columns to prior limits
+
+    label: str
+        Legend label
 
     """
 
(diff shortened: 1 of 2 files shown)


anesthetic/samples.py
```diff
@@ -43,32 +44,40 @@ class MCMCSamples(WeightedDataFrame):
...
+        mapping from columns to prior limits
+
+    label: str
+        Legend label
 
     """
 
-    _metadata = WeightedDataFrame._metadata + ['tex', 'limits', 'u', 'root']
+    _metadata = WeightedDataFrame._metadata + ['tex', 'limits', 'u',
+                                               'root', 'label']
```
lukashergt: Not really for this PR, but as a side remark: `'u'` is very cryptic. I've gone through the code once to figure out what it does, but I forgot again... Is there maybe a more descriptive name for that variable?

Similar for `w`. Although, not as bad as it often appears with the key `'weight'` (and at least to me it's somewhat familiar/expected)...

tests/test_samples.py
```diff
@@ -187,9 +206,10 @@ def test_plot_2d_legend():
                     assert(all([isinstance(h, Rectangle) for h in handles]))
     plt.close('all')
 
+    # Test label kwarg label kwarg for hist and scatter
```
lukashergt: `label kwarg` repeated

anesthetic/samples.py
williamjameshandley: OK, I've added an issue #44 for further discussion and to remind us to address this at some point

tests/test_samples.py
williamjameshandley: Fixed in 9f457a2

Issue #41, opened 2019-07-22
Setting each diagonal block with different plotting types


xichenamoy:
Currently for all the diagonal blocks of a triangle plot, we could set types={'diagonal': 'hist'} or {'diagonal': 'kde'} . But sometimes one may want to have different type settings for each diagonal blocks. Could we add a feature that enable users to set the type of each diagonal block separately?

williamjameshandley:
I think this should be achievable. The main thing to discuss is what the interface should be  (@lukashergt may want to weigh in here).

Given that `types` is already provided as a dictionary, we could simply allow extra keys in this dictionary that would override the 'diagonal' 'upper' and 'lower' choices, for example

```python
types={'diagonal':'hist', 
       'lower':'kde', 
       ('x0','x1'):'fastkde', 
       'x0':'kde'}
```
would set the `x0` on the diagonal to be kde, and the rest of the diagonals to be `hist`, whilst the lower triangle would be filled with `kde`s, except the `('x0','x1')` block, which would be a `fastkde`

lukashergt:
That sort of addition to the dictionary would be possible, I guess.

Another option that possibly already works (haven't tried it though) could be to set up the `axes` with `make_2d_axes` (say e.g. with lower and diagonal set to true), then call `plot_2d` first on all `axes` with `types={'lower': 'kde'}` (this will leave the diagonal axes empty for now) and then call `plot_2d` with `types={'diagonal': 'hist'}` on the desired subset of axes (e.g. `axes.iloc[0, 0]`) and analogously for `types={'diagonal': 'kde'}`.

lukashergt:
Here an example of how to make it work with the current set-up. The colours will need manual tweaking. This might seem complicated, but extending the dictionary as proposed above would mean also the handling of kwargs would have to be extended and I fear that might complicate things a lot more.

What do you think, @williamjameshandley, @xichenamoy?

```python
ns1 = NestedSamples(root="./tests/example_data/pc")
ns2 = NestedSamples(root="./tests/example_data/pc")
fig = plt.figure(figsize=(10, 10))
fig, axes = make_2d_axes(['x0', 'x1', 'x2', 'x3'], fig=fig, upper=False)
ns1.plot_2d(axes=axes,              types={'lower': 'kde'},     label='2D kde')
ns1.plot_2d(axes=axes.iloc[1:, 1:], types={'diagonal': 'kde'},  label='kde')
[...]

PR #45, opened 2019-07-26
Fix to colouring for missing plots

### LINKED ISSUE #19: wrong colour when second model plots into empty axes
When plotting 2 models one of which has more parameters than the other, e.g. LCDM and rLCDM, and if we then plot the model with less parameters first including the unshared parameter, then the second model will have the wrong colour in the axes with the unshared parameter.

The colour should be dependent on the call to `plot_2d` and thus on the model. It should not depend on whether any axis is drawn in for the first or subsequent time.

Example code:
```python
fig, axes= ns_lcdm.plot_2d(['A_s', 'n_s', 'r'], label='LCDM')
ns_rlcdm.plot_2d(axes=axes, diagonal_kwargs={'ls': '--'}, label='rLCDM')
leg = get_legend_proxy(fig)
axes['n_s']['n_s'].legend(leg, ['LCDM', 'rLCDM'], bbox_to_anchor=(1, 1), loc='lower left')
```

image

Note, how in the above code the second model gets the additional keyword `ls='--'` which shows that rLCDM is drawn in _orange_ in the upper two axes rows and _blue_ in the bottom row.

For now this behaviour can be circumvented by plotting the model with the most parameters first. This won't help in cases where two models each have an additional parameter not shared with the other, though, say e.g. r+LCDM and alpha+LCDM.

williamjameshandley:
# Description

This PR added a failing test that shows the issue noted in #19. It now fixes that test by reordering the 'blank plots' in `MCMCSamples.plot`

Fixes #19 

# Checklist:

### DIFF
--- a/anesthetic/samples.py
+++ b/anesthetic/samples.py
@@ -127,45 +127,46 @@ class MCMCSamples(WeightedDataFrame):
         kwargs['label'] = kwargs.get('label', self.label)
 
         if do_1d_plot:
-            xmin, xmax = self._limits(paramname_x)
-            if plot_type == 'kde':
-                plot = plot_1d
-            elif plot_type == 'hist':
-                plot = hist_1d
-            elif plot_type is None:
-                ax.plot([], [])
-            else:
-                raise NotImplementedError("plot_type is '%s', but must be in "
-                                          "{'kde', 'hist'}." % plot_type)
             if paramname_x in self and plot_type is not None:
                 x = self[paramname_x].compress()
+                xmin, xmax = self._limits(paramname_x)
+                if plot_type == 'kde':
+                    plot = plot_1d
+                elif plot_type == 'hist':
+                    plot = hist_1d
+                else:
+                    raise NotImplementedError("plot_type is '%s', but must be"
+                                              " in {'kde', 'hist'}."
+                                              % plot_type)
                 return plot(ax, x, xmin=xmin, xmax=xmax, *args, **kwargs)
+            else:
+                ax.plot([], [])
 
         else:
-            xmin, xmax = self._limits(paramname_x)
-            ymin, ymax = self._limits(paramname_y)
-
-            if plot_type == 'kde':
-                nsamples = None
-                plot = contour_plot_2d
-                ax.scatter([], [])
-            elif plot_type == 'scatter':
-                nsamples = 500
-                plot = scatter_plot_2d
-                ax.plot([], [])
-            elif plot_type is None:
-                ax.plot([], [])
-                ax.scatter([], [])
-            else:
-                raise NotImplementedError("plot_type is '%s', but must be in "
-                                          "{'kde', 'scatter'}." % plot_type)
-
             if (paramname_x in self and paramname_y in self
                     and plot_type is not None):
+                if plot_type == 'kde':
+                    nsamples = None
+                    plot = contour_plot_2d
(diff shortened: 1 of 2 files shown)


anesthetic/samples.py
```diff
@@ -189,14 +193,15 @@ def plot_1d(self, axes, *args, **kwargs):
             Pandas array of axes objects
 
         """
+        plot_type = kwargs.pop('types', 'kde')
+
```
lukashergt: So far I was using `plot_type` (which is popped in `samples.plot`) to decide what kind of 1D plot I want. Now, providing the kwarg `plot_type` results in 
```
TypeError: plot() got multiple values for keyword argument 'plot_type'
```

Code to reproduce:
```
pc = NestedSamples(root="./tests/example_data/pc")
pc.plot_1d(['x0', 'x1'], label="pc", plot_type='hist');
```

williamjameshandley: Ah, I had noticed that `MCMCSamples.plot_1d` was missing a 'types' argument, in analogy to `MCMCSamples.plot_2d`. In this PR I've made them more consistent in preparation for the finer-grained control suggested in #41 .

The fix is to pass `types='hist'` rather than `plot_type='hist'`. I acknowledge that this is technically a breaking change, although only to undocumented (and, tbh unknown on my part) functionality.

williamjameshandley: One of the things I'm considering is deprecating the cryptic `types` arguments, with the plan to replace them with `subplot_types`, which is more informative.

lukashergt: By the way, the error actually also happens in the 2D case:
```
pc = NestedSamples(root="./tests/example_data/pc")
fig, axes = pc.plot_2d(['x0', 'x1', 'x4', 'x5'], label="pc", plot_type='kde');
```

Might it be more consistent to allow `plot_type` as a kwarg for `plot_2d` that broadcasts the provided string to all subplots?

lukashergt: I'm not sure how to fix this, but having `plot_type` and `types` in parallel like this seems wrong to me. If `types` is the one intended to be passed to `plot_1d` then `plot_type` should not be a valid kwarg.

> I acknowledge that this is technically a breaking change, although only to undocumented (and, tbh unknown on my part) functionality.

Not undocumented with #24. Actually, `plot_type` was one of my major motivations for suggesting #24. 

> In this PR I've made them more consistent in preparation for the finer-grained control suggested in #41 .

The thing is, `plot_type` and `types` don't quite have the same role, so I'm not sure whether this is really more consistent. The need for `types` arises because in the 2D case we distinguish between 3 different families of plots (diagonal, lower and upper) which all need a `plot_type` assignment. This is a 2D specific feature and, thus, `types` is unnecessary in the 1D case.

> One of the things I'm considering is deprecating the cryptic types arguments, with the plan to replace them with subplot_types, which is more informative.

Replacing `types` with a more general `subplot_types` that determines e.g. which plot type for which parameter would be an option. At the same type `plot_type` could be replaced with e.g. `default_plot_type` that broadcasts the plot type across all subplots and `subplot_types` can then be used to pick and alter specific parameters.

lukashergt: In any case, might it be better to address all this in a separate PR?

williamjameshandley: Agreed, reverted in 61e3f6b

PR #46, opened 2019-07-30
Implementation of 2D histograms

### LINKED ISSUE #2: Option for histograms as alternative to KDE

williamjameshandley:
# Description

This aims to provide the functionality requested by #2 by providing 2D histograms.

Fixes #2  (completes #10 by adding 2d histograms)

# Checklist:


anesthetic/plot.py
```diff
@@ -402,6 +392,57 @@ def contour_plot_2d(ax, data_x, data_y, *args, **kwargs):
...
+    if len(data_x) == 0 or len(data_y) == 0:
+        return numpy.zeros(0), numpy.zeros(0), numpy.zeros((0, 0))
+
+    pdf, x, y = numpy.histogram2d(data_x, data_y, density=True)
+    contours = iso_probability_contours(pdf)
+    pdf = numpy.digitize(pdf, contours, right=True)
+    pdf = numpy.array(contours)[pdf]
+    pdf = numpy.ma.masked_array(pdf, pdf < contours[1])
+    cmap = basic_cmap(color)
+    image = ax.pcolormesh(x, y, pdf.T, cmap=cmap, vmin=0, vmax=pdf.max(),
```
lukashergt: Why `pcolormesh` here instead of `hist2d`? 
`hist2d` is the standard 2d version of the 2d `hist` with with mostly the same kwargs (e.g. `bins`, `range`).

williamjameshandley: It depends on whether you want histograms that look like this:

hist1

or like this

hist2

If we want the same contouring effect as we have for kernel density estimation, then we have to split these up. However, the latter simple wrapper around `plt.hist2D` is considerably simpler than the above implementation. We could possibly include an option for both?

williamjameshandley: 9cee366 Addresses this properly, both in providing a complete `hist2d` option, allowing users to provide contours, and also better mimicking the
[source code](https://matplotlib.org/3.1.1/_modules/matplotlib/axes/_axes.html#Axes.hist2d) of `hist2d`. This means that `bins`, `range` etc can now be passed to `MCMCSamples.hist_plot_2d`

anesthetic/plot.py
```diff
@@ -402,6 +392,79 @@ def contour_plot_2d(ax, data_x, data_y, *args, **kwargs):
...
+        density estimator.
+
+    xmin, xmax, ymin, ymax: float
+        lower/upper prior bounds in x/y coordinates
+        optional, default None
+
+    contours: list
+        amount of mass within each iso-probability contour.
+        If None defaults to hist2d.
+        Default [0.05, 0.33]
```
lukashergt: It's not really the amount of mass within the contour, right? I found this confusing in `contour_plot_2d` as well, as I am more used to the numbers 0.68 and 0.95. This is actually more what `plt.contour` refers to as `levels`. The amount of mass within the contours is really one minus the levels.

williamjameshandley: OK, I think I've improved on this now, so that the numbers correspond to the intuitive values of 0.68 and 0.95

williamjameshandley:
@lukashergt Are there any further changes that are required before this is ready for merging?

anesthetic/plot.py
```diff
@@ -402,6 +392,79 @@ def contour_plot_2d(ax, data_x, data_y, *args, **kwargs):
...
+    contours = kwargs.pop('contours', [0.68, 0.95])
+    color = kwargs.pop('color', next(ax._get_lines.prop_cycler)['color'])
+
+    if len(data_x) == 0 or len(data_y) == 0:
+        return numpy.zeros(0), numpy.zeros(0), numpy.zeros((0, 0))
+
+    cmap = basic_cmap(color)
+
+    if contours is None:
+        pdf, x, y, image = ax.hist2d(data_x, data_y, cmap=cmap,
```
lukashergt: Since `contours` is kwarg popped earlier, this option only gets triggered when one expressedly sets `contours` to `None`. From the docs I take it that that is the intended behaviour?

For me the contour-hist would not be what I would go for. I typically go for a hist to play around with the number of bins and get a feel for the data. But the contour-hist seems to me analogous to in the 1D case switching from a KDE to a hist with just 3 bins. But since it's in the docs, I guess it doesn't matter terribly much.

williamjameshandley:
An issue had arisen with the previous implementation that caused problems with the ranges. I've now fixed this, but it does not work with astropy due to a bug in the handling of weights. The following code shows this to not work:

```python
import numpy
import astropy.visualization

N = 1000
dat = numpy.random.rand(N)
w = numpy.exp(-((dat-0.5)/0.01)**2/2)

# matplotlib works
plt.hist(dat, weights=w, range=(0.45,0.55))

# astropy does not
astropy.visualization.hist(dat, weights=w, range=(0.45,0.55))
```

I've removed astropy support for now, and will file an issue for this to be fixed

anesthetic/plot.py
williamjameshandley: ￼8fd14ca adds the contours kwarg to ``contour_plot_2d``, so that it's consistent with ``hist_2d``. I have two questions
1. Should the default be `[0.68, 0.95]`, or `None`, for either `hist_2d` or `contour_plot_2d`
2. Should we use the keyword `contours`, or `levels`. The latter is the argument that is passed into `contour` and `contourf`, so would avoid conflicts.

lukashergt: These questions address two objectives of this package that sometimes point in different directions: 
On the one hand we want to provide a nice tool for plotting MCMC and Neste Sampling output. On the other hand we want a package that nicely integrates with matplotlib and thus making it intuitively usable for people familiar with matplotlib.

1. Here, I feel like the defaults would indeed be different for `hist_2d` and `contour_plot_2d`.
    - `contour_plot_2d`: For contour plots we definitely want the 68 and 95% contours as default. This is the standard which people will want to plot. Hence, I'd go for either `[0.68, 0.95]` as default for `contours` or `[0.32, 0.05]` as default for `levels`.
    - `hist_2d`: I would swap to 2D histograms when I want a more fine grained view of the data and play around with how different bins might affect the picture. For that I wouldn't want to see only 2 quantiles, I would want a proper histogram. Also for flat priors 2D histograms might be useful, but also in this case I wouldn't want quantiles but a proper histogram that shows that the data is evenly spread across all bins. Hence, for `hist_2d` I would prefer the default set to `None`.

2. A case can be made for either one here. Do we want to stick close to `matplotlib` or change things to be closer to the usage.
    - `contours` is the term typically used in the context of these plots ("68 and 95% contours"). It would also indicate that these are not the `matplotlib` typical contour plots, but are designed to be blobs/bananas/... with about 2 levels.
    - `levels` is the term that's more conform with `matplotlib`. This would prevent having two terms for essentially the same thing which would simplify the kwarg handling. I think I personally am leaning to this option, but if we use this we probably should document that levels `[0.32, 0.05]` correspond to 68 and 95% contours. Or should we invert the levels internally, i.e. `levels = 1. - levels`?

williamjameshandley: I agree with point 1. Now that `contour_plot_2d` has a `contours`/`levels` kwarg, it's more obvious that you can supply these to hist_2d if desired. Changes pushed

I've opted for levels, with the defaults at the sensible values [0.68, 0.95].

anesthetic/plot.py
```diff
@@ -343,6 +337,11 @@ def contour_plot_2d(ax, data_x, data_y, *args, **kwargs):
         x and y coordinates of uniformly weighted samples to generate kernel
         density estimator.
 
+    levels: list
+        amount of mass within each iso-probability contour.
+        If None defaults to hist2d.
+        optional, default [0.68, 0.95]
```
lukashergt: > `If None defaults to hist2d`.

Is this a copy-paste error or is that really how it works for `contour_plot_2d`?

williamjameshandley: copy-paste error -- corrected.

anesthetic/plot.py
```diff
@@ -359,49 +358,111 @@ def contour_plot_2d(ax, data_x, data_y, *args, **kwargs):
...
+        density estimator.
+
+    xmin, xmax, ymin, ymax: float
+        lower/upper prior bounds in x/y coordinates
+        optional, default None
+
+    levels: list
+        amount of mass within each iso-probability contour.
+        If None defaults to hist2d.
+        optional, default None
```
lukashergt: Maybe good to specify that it defaults to `matplotlib`'s `hist2d`?

williamjameshandley: Made this a little more descriptive in the latest push.

anesthetic/samples.py
```diff
@@ -129,45 +129,51 @@ def plot(self, ax, paramname_x, paramname_y=None, *args, **kwargs):
...
                 xmin, xmax = self._limits(paramname_x)
                 if plot_type == 'kde':
-                    plot = plot_1d
+                    x = self[paramname_x].compress()
+                    return plot_1d(ax, x, xmin=xmin, xmax=xmax,
+                                   *args, **kwargs)
                 elif plot_type == 'hist':
-                    plot = hist_1d
+                    return hist_1d(ax, self[paramname_x], weights=self.w,
+                                   xmin=xmin, xmax=xmax, *args, **kwargs)
```
lukashergt: It is the weights passed here that breaks the interface with astropy and that also prevents the `str` options for `bins` in numpy and matplotlib's histogram functions. I agree that it might be simpler to drop any interface with astropy, however, I find it a pity if none of the `str` options work anymore.

Could we turn the data into unweighted data before passing it to `hist_1d`?

williamjameshandley: Weighted samples are going to be quite important for future development, so I'm keen to have hist_1d accepting weights. 

I have another PR almost ready for pushing that follows from this one, implementing scipy.stats.gaussian_kde as an alternative to fastkde. fastkde uses unweighted samples, whilst gaussian_kde uses weighted ones. 

A consistent way of reinstating astropy would be to have a separate 'astropyhist' option in addition to 'hist'. Would that do as a compromise?

williamjameshandley: OK, Pushed the above. I think it's better anyway to explicitly state that you're using astropy, rather than letting the system quietly replace it. Resolves #47

lukashergt: Is there a reason why weighted samples are important in the context of histograms? Otherwise, wouldn't it be easy enough to turn a weighted sample into an unweighted one just before passing it to `hist_1d`? That wouldn't interfere with weighted samples at all...

williamjameshandley: Ah, well spotted. I had forgotten to put in that compression step before the astropyhist.

In general, it's always better to use the weights if you can. Compressing is not lossless, and it's better to let the underlying algorithms use the weights as best they see fit.

lukashergt: Hmm, I guess you are thinking more in terms of nested sampling whereas I've been thinking more in terms of MCMC chains where the weights are integers. With integer weights there would be no problem turning weighted into unweighted samples, right? Or am I missing something?

williamjameshandley: That's right. It's trivial for MCMC, but for nested sampling because the weights are continuous (and a large fraction are close to zero) making this compression is not trivial, and not lossless. The best you can do is provided by the compress functionality of WeightedDataFrame.

Note that even for MCMC, you could run into memory issues if your chain is already large, and you intend to increase its size by several factors by simply repeating samples.

lukashergt: Hmm, ok. 

How accurate is the `compress()` function? Could this cause noticeable differences in the appearance of a histogram (for a sufficiently large sample size)?

williamjameshandley: For large sample sizes the difference should  be negligible. Why?

lukashergt: `compress()`, `channel_capacity()` etc. are not super intuitive in how they work. Just making sure that they don't have a visible effect on histograms that I wasn't aware of.

It's nice to have `'astropyhist'` as an option.

williamjameshandley:
OK, Travis and coverage are now passing (minus the python 3.7 conda errors). There are no functional changes since the last review, so if @lukashergt approves, this can me merged into master.

lukashergt:
> OK, Travis and coverage are now passing (minus the python 3.7 conda errors). There are no functional changes since the last review, so if @lukashergt approves, this can be merged into master.

Haha! Wow, reading these last commit messages in my postbox made me crack up.

williamjameshandley:
>Haha! Wow, reading these last commit messages in my postbox made me crack up.

It's always the same with trying to fix any CI software.

Fixes #47

anesthetic/plot.py
```diff
@@ -304,23 +303,34 @@ def hist_1d(ax, data, *args, **kwargs):
...
-    h, edges, bars = hist(data, range=(xmin, xmax), histtype=histtype,
-                          *args, **kwargs)
-    # As the y-axis on the diagonal 1D plots of the triangle plot won't
-    # be labelled, we set the maximum bar height to 1:
+    if plotter == 'astropyhist':
+        try:
+            print(xmin, xmax)
+            print(data)
+            print(weights)
+            h, edges, bars = hist(data, ax=ax, range=(xmin, xmax),
```
lukashergt: A bit late, but I just noticed this now. We don't need those print statements any longer, right? 

Should this be addressed in a new PR?

williamjameshandley: Opened #51

Issue #35, opened 2019-07-16
Boundary corrections


williamjameshandley:
Currently boundary corrections are implemented via 'reflecting boundary conditions'. The industry standard is to use a local linear approximation at the edges. A good review of a simple strategy [can be found here](http://www3.stat.sinica.edu.tw/statistica/oldpdf/A6n414.pdf)

This is very effectively implemented in [GetDist](https://cosmologist.info/notes/GetDist.pdf), although the code there is specific to the custom KDE that is used.

In an ideal world, these boundary corrections would be submitted as a PR for `scipy.stats.gaussian_kde`, but for now we will battle-test it in `anesthetic`.

@williamjameshandley is happy to take charge of a PR in the near future, but would welcome comments/suggestions.

PR #50, opened 2019-08-10
scipy.stats.gaussian_kde contour plots

### LINKED ISSUE #4: Tricontour plots

williamjameshandley:
# Description

- This is a relatively large addition, offering an alternative kde to `fastkde` using `scipy.stats.gaussian_kde`
- `fastkde` is therefore now an optional requirement, with CI updated accordingly).
- It is a necessary step in order to implement boundary corrections #35 

The main innovation here is to extensively use matplotlib's triangulation functionality to first dynamically bin, with bins defined by a random subsample. This means that you put bins where they are needed, and typically only 1000 points are required. Triangulation is also used to plot the contours, so once again only 1000 samples are required. This means that despite the fact that `gaussian_kde` is slower than `fastkde`, it is still snappy enough to be used dynamically for large plots.

The final result is reasonable:

Figure_1

Although it clearly needs the boundary data correction to remove the edge effects.

One additional change that is mixed in here is to revert `ax.scatter` to `ax.plot` in `scatter_plot_2d`, since I have observed scatter to make axis limits misbehave in non-trivial ways when applied to practical examples where data have very different scales. This has been [documented](https://github.com/matplotlib/matplotlib/issues/9979) and will be [fixed in matplotlib v3.2](https://github.com/matplotlib/matplotlib/pull/13642), but until that is merged, we should revert to `ax.plot`. Whilst `ax.scatter` is more consistent with the name of the function, moving away from that actually neatens up a lot of the colouring issues that we have seen before (#32 #21 #45 #19 #31)

Fixes #4 
Replaces #25 

# Checklist:


williamjameshandley:
> Renaming `kde.py` to `fastkde.py` causes an import error even when fastkde is installed because of overloading the term fastkde.

Well spotted. I think I should have fixed this now. There weren't actually that many places that anesthetic.fastkde was being imported, so it's a relatively easy change.

anesthetic/plot.py
```diff
@@ -402,6 +472,77 @@ def contour_plot_2d(ax, data_x, data_y, *args, **kwargs):
...
+    c: matplotlib.contour.QuadContourSet
+        A set of contourlines or filled regions
+
+    """
+    xmin = kwargs.pop('xmin', None)
+    xmax = kwargs.pop('xmax', None)
+    ymin = kwargs.pop('ymin', None)
+    ymax = kwargs.pop('ymax', None)
+    weights = kwargs.pop('weights', None)
+    n = kwargs.pop('ncompress', 1000)
```
lukashergt: `n` is used just once in this function, so I don't see the need for the abbreviation from `ncompress`. I'd stick with the more descriptive name (also for 1D case).

anesthetic/utils.py
```diff
@@ -182,3 +183,131 @@ def iso_probability_contours(pdf, contours=[0.68, 0.95]):
...
+
+    """
+    if w is None:
+        w = numpy.ones_like(x)
+
+    # Select samples for triangulation
+    if sum(w != 0) < n:
+        i = numpy.arange(len(w))
+    else:
+        i = numpy.random.choice(len(w), size=n, replace=False, p=w/w.sum())
```
lukashergt: Should `replace` really be `False` here? 
How does that affect the weighting?

anesthetic/utils.py
```diff
@@ -182,3 +183,131 @@ def iso_probability_contours(pdf, contours=[0.68, 0.95]):
...
+    x_, y_ = Linv.dot([x[i], y[i]])
+    tri = Triangulation(x_, y_)
+
+    # Mask out triangles with unreasonably large perimeters
+    vec = numpy.array([x_[tri.triangles], y_[tri.triangles]]).transpose()
+    s = (numpy.linalg.norm(vec[1, :, :] - vec[0, :, :], axis=1) +
+         numpy.linalg.norm(vec[2, :, :] - vec[1, :, :], axis=1) +
+         numpy.linalg.norm(vec[0, :, :] - vec[2, :, :], axis=1))
+    A = s**2
+    tri.set_mask(A > 25*A.mean())
```
lukashergt: Where is the number 25 here coming from? How do you quantify "unreasonably" in "unreasonably large perimeters"?

anesthetic/samples.py
```diff
@@ -154,34 +160,39 @@ def plot(self, ax, paramname_x, paramname_y=None, *args, **kwargs):
...
                 if plot_type == 'kde':
+                    x = self[paramname_x]
+                    y = self[paramname_y]
+                    return kde_contour_plot_2d(ax, x, y, weights=self.w,
+                                               xmin=xmin, xmax=xmax,
+                                               ymin=ymin, ymax=ymax,
+                                               *args, **kwargs)
+                elif plot_type == 'fastkde':
                     x = self[paramname_x].compress()
                     y = self[paramname_y].compress()
```
lukashergt: This concerns fastkde, so is actually less relevant considering that kde is taking over, but still. For large amounts of data, the contour plots with fastkde  have too much fine grained structure:

image

Is `.compress()` geared towards the 1D case? Should we set a maximum `nsamples`? Any other idea of improving the compression in this case?

williamjameshandley: I'm not sure why fastkde would be showing this behaviour, since it's supposed to be optimised for the case of lots of samples.

However, the `compress()` function does end up having repeated rows, for which fastkde may not be optimised for. What happens to the above plot if you change `compress()` for `compress(-1)` in `MCMCSamples.plot_2d` ?

anesthetic/utils.py
williamjameshandley: This was very much done by trial and error. 

It combats the convex case (rings or bananas), since then the triangulation has ~O(a few) large perimeter triangles that tile outside the distribution. In the convex case, this line has no effect.

anesthetic/utils.py
williamjameshandley: Excellent question. It all depends on what you do with the weights after. For concreteness, consider this code:

```python
x = numpy.array([1,2,3])
w = numpy.array([1e10,1,1])
print(numpy.random.choice(x, size=2, replace=False, p = w/w.sum()))
```

You are right that `replace=False` would definitely be wrong if you were using it to 'de-weight' a set of samples (as in the `MCMCSamples.compress`) function). However, if you keep the weights, then this acts as an alternative form of compression . You can actually think of the strategy performed in `MCMCSamples.compress` as doing this, but with the optimal `size` chosen by the channel capacity.

In actual fact, since we are only using this to bin the weights into ~O(1000) samples, it doesn't even matter that i is chosen as a compression strategy. For example, if you wanted to ensure better coverage in the tails, the `p` in the above line need not be proportional to the weights. Points chosen in the tails would aggregate to have very low weight, but that may be what you want in practice if you wanted to go beyond two-sigma contours.

anesthetic/samples.py
lukashergt: Not as chaotic but super wobbly:

image

lukashergt: It gets a bit better by manually reducing the number of samples. Here for `nsamples=10000`:

image

Maybe worth turning `nsamples` into a kwarg, then again maybe not that needed as we have `gaussian_kde` soon.

anesthetic/utils.py
lukashergt: This seems a bit arbitrary, could using a z-score or the IQR improve this? 

E.g. crop 1.5 IQR above the 3rd quartile: 
```python
tri.set_mask(s > s.quantile(0.75) + 1.5 * stats.iqr(s))
```

anesthetic/samples.py
williamjameshandley: OK. The latest push adds ncompress as a kwarg, which should give the control you desire for any of the plotting functions.

anesthetic/utils.py
williamjameshandley: Here is a typical histogram of s:
Figure_1
We basically want to trim off those points way out in the tail. The quantile iqr approach you suggest would end up trimming too much of the main distribution

williamjameshandley: OK. It seems that taking the log of s, zscoring and removing those points greater than 3 standard deviations seems to do the trick. 3 standard deviations makes sense for 1000 samples, since that's 99.9% of the distribution. Easily generalises to nsamples with scipy.stats.norm.ppf. Changes pushed
Figure_1

anesthetic/plot.py
```diff
@@ -510,7 +511,7 @@ def kde_contour_plot_2d(ax, data_x, data_y, *args, **kwargs):
     ymin = kwargs.pop('ymin', None)
     ymax = kwargs.pop('ymax', None)
     weights = kwargs.pop('weights', None)
-    ncompress = kwargs.pop('ncompress', 1000)
+    ncompress = kwargs.pop('ncompress', None)
```
lukashergt: Why change this to `None` when docs still state `Default 1000`?

williamjameshandley: I've pushed the alternative, which makes `MCMCSamples.plot` a little more ugly, but the lower functions more intuitive.

anesthetic/utils.py
lukashergt: This looks good. If this also works as desired on banana cases then this is great.

williamjameshandley: Without and with the mask:
Figure_1

lukashergt: Oh dear, well with the mask is clearly better... but it doesn't look optimal either...?

williamjameshandley: It's actually quite a difficult one for it to plot. Increasing the number of samples with ncompress improves matters (asking it to plot this one accurately with only 1000 samples is a bit of a tall order) 
Figure_1

lukashergt: Yeah, that looks even better. Is there a way to find out or estimate the minimum amount of `nsamples` needed for neat plots? The `channel_capacity` returns too large numbers which take too long to compute, right?

anesthetic/utils.py
```diff
@@ -182,3 +184,132 @@ def iso_probability_contours(pdf, contours=[0.68, 0.95]):
...
+    tri = Triangulation(x_, y_)
+
+    # Mask out triangles with unreasonably large perimeters
+    vec = numpy.array([x_[tri.triangles], y_[tri.triangles]]).transpose()
+    s = (numpy.linalg.norm(vec[1, :, :] - vec[0, :, :], axis=1) +
+         numpy.linalg.norm(vec[2, :, :] - vec[1, :, :], axis=1) +
+         numpy.linalg.norm(vec[0, :, :] - vec[2, :, :], axis=1))
+
+    # Mask out triangles with a perimeter zscore smaller than expected
+    tri.set_mask(zscore(numpy.log(s)) > -norm.ppf(1/len(s)))
```
lukashergt: Have you looked into [`matplotlib.tri.TriAnalyzer`](https://matplotlib.org/3.1.1/api/tri_api.html#matplotlib.tri.TriAnalyzer) and in particular [`get_flat_tri_mask`](https://matplotlib.org/3.1.1/api/tri_api.html#matplotlib.tri.TriAnalyzer.get_flat_tri_mask)? Could this be even better than our manual masking?

williamjameshandley: >Have you looked into matplotlib.tri.TriAnalyzer and in particular get_flat_tri_mask? Could this be even better than our manual masking?

I did try this first. The trouble is that it's not just flat triangles we want to be rid of, but over-large ones as well.

anesthetic/utils.py
williamjameshandley: I think this likely requires manual tuning. The main point is that the default should be fast, and good enough. One can always dial up ncompress to maximum once you're happy with the general layout.

PR #52, opened 2019-08-14
First-order boundary correction

### LINKED ISSUE #35: Boundary corrections

williamjameshandley:
# Description

This implements a 'cut and normalise' boundary correction. Whilst [far better corrections are available](https://cosmologist.info/notes/GetDist.pdf), this is manifestly better than no correction, and has the advantage of being an extremely minor modification.

The below plots show that it slightly improves x2 and x4, and clearly improves x3.

Without boundary correction:
Figure_1

With boundary correction:
Figure_2

Fixes #35 

# Checklist:

### DIFF
--- a/anesthetic/boundary.py
+++ b/anesthetic/boundary.py
@@ -1,0 +1,37 @@
+"""Boundary correction utilities."""
+
+import numpy
+from scipy.special import erf
+
+
+def cut_and_normalise_gaussian(x, p, sigma, xmin=None, xmax=None):
+    """Cut and normalise boundary correction for a Gaussian kernel.
+
+    Parameters
+    ----------
+    x: numpy.array
+        locations for normalisation correction
+
+    p: numpy.array
+        probability densities for normalisation correction
+
+    sigma: float
+        bandwidth of KDE
+
+    xmin, xmax: float
+        lower/upper prior bound
+        optional, default None
+
+    Returns
+    -------
+    p: numpy.array
+        corrected probabilities
+
+    """
+    correction = numpy.ones_like(x)
+
+    if xmin is not None and (x >= xmin).all():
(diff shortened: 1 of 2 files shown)


williamjameshandley:
@lukashergt Do you have any input into this implementation?

lukashergt:
Sorry, had a terrible internet connection this past week...
Looks indeed better than no correction, but in the long run `x2` and `x4` will need better boundary corrections. 

Is the "cut and normalise" method following a particular paper or book? If so, it would be nice to have the reference in the docs.

williamjameshandley:
> Sorry, had a terrible internet connection this past week...
> Looks indeed better than no correction, but in the long run `x2` and `x4` will need better boundary corrections.
> 
> Is the "cut and normalise" method following a particular paper or book? If so, it would be nice to have the reference in the docs.

We'll document properly when we've settled on a boundary implementation, but we should make a note to do this as part of #22 and #24. For now [here is an open source paper on cut and normalise methods](https://arxiv.org/pdf/1511.08271.pdf), and if we want to improve it to the state-of-the-art in cut-and-normalise, then we should do so with [this method](https://www.sciencedirect.com/science/article/pii/S0378375897000827)

Issue #44, opened 2019-07-25
Consider renaming u and w


williamjameshandley:
As indicated in #43, the attribute `MCMCSamples.u` and `WeightedSeries.w` is pretty opaquely named. We should either

1. Hide this implementation further
2. Create a clearer variable name

(For reference, the variable `u` is a set of fixed random numbers used to make a consistent trimming procedure when compressing the number of samples)

PR #55, opened 2019-09-09
Weighted dataframe slicing

### LINKED ISSUE #53: Best strategy for selecting a certain parameter range? Allow mask?
I would like to select a certain parameter range and then plot the contours for this selection, e.g. say I'd like to look at only positive x0 with the example data:
```python
ns = NestedSamples(root="/scratch/Projects/AnestheticPrj/anesthetic/tests/example_data/pc")
fig = plt.figure(figsize=(7, 7))
fig, axes = make_2d_axes(['x0', 'x1'], fig=fig, upper=False)
ns.plot_2d(axes=axes, label='foo')
axes.iloc[-1, 0].legend(bbox_to_anchor=(len(axes), len(axes)), loc='upper right')
```
image

It is easy to create a mask for a specific parameter range:
```python
mask = ns['x0'] > 0
```
A table with the selected data can be displayed simply by:
```python
ns[mask]
```

However, the anesthetic specific functions can't be applied to this, i.e. exchanging `ns` for `ns[mask]` in the above code doesn't work but fails with:
```bash
ValueError: weights should have the same shape as x
```

This means while the mask is applied to the parameters (e.g. self['x0'], but also self['weight']!) it is not applied to the array `self.w` (and `self.u`?). 

I see two possible solutions here:

1) Might it possibly be better to be working internally with `self['weight']` instead of `self.w`? (And analogously for `self.logL`, `self.u` etc.) Could that be enough to make masking work as described above?

2) We could introduce a `mask` kwarg to the plotting functions?

williamjameshandley:
# Description

This fixes #53 by ensuring that the invisible weights array is sliced whenever the array is.

# Checklist:

### DIFF
--- a/anesthetic/samples.py
+++ b/anesthetic/samples.py
@@ -56,7 +56,7 @@ class MCMCSamples(WeightedDataFrame):
 
     """
 
-    _metadata = WeightedDataFrame._metadata + ['tex', 'limits', 'u',
+    _metadata = WeightedDataFrame._metadata + ['tex', 'limits',
                                                'root', 'label']
 
     @property
@@ -80,15 +80,7 @@ class MCMCSamples(WeightedDataFrame):
             self.limits = kwargs.pop('limits', {})
             self.label = kwargs.pop('label', None)
             self.root = None
-            self.u = None
-
             super(MCMCSamples, self).__init__(*args, **kwargs)
-
-            self.u = numpy.random.rand(len(self))
-
-            if self.w is not None:
-                self['weight'] = self.w
-                self.tex['weight'] = r'MCMC weight'
 
             if logL is not None:
                 self['logL'] = logL
@@ -425,10 +417,6 @@ class NestedSamples(MCMCSamples):
             if 'nlive' in self:
                 self.beta = self._beta
 
-            if self.w is not None:
-                self['weight'] = self.w
-                self.tex['weight'] = r'MCMC weight'
-
     @property
     def beta(self):
         """Thermodynamic inverse temperature."""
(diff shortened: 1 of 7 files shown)


anesthetic/samples.py
```diff
@@ -80,16 +80,8 @@ def __init__(self, *args, **kwargs):
...
             self.root = None
-            self.u = None
-
             super(MCMCSamples, self).__init__(*args, **kwargs)
 
-            self.u = numpy.random.rand(len(self))
-
-            if self.w is not None:
-                self['weight'] = self.w
-                self.tex['weight'] = r'MCMC weight'
```
lukashergt: I think it's worth keeping this, it's nice to have this in the table when printing `samples`...

anesthetic/samples.py
```diff
@@ -425,10 +417,6 @@ def __init__(self, *args, **kwargs):
             if 'nlive' in self:
                 self.beta = self._beta
 
-            if self.w is not None:
-                self['weight'] = self.w
-                self.tex['weight'] = r'MCMC weight'
```
lukashergt: idem

anesthetic/weighted_pandas.py
```diff
@@ -62,29 +79,48 @@ def compress(self, nsamples=None):
...
+            return self._w[self.index]
+        else:
+            return None
+
+    @property
+    def _u(self):
+        """Random number for consistent compression."""
+        return self._u_[self.index]
+
     def __init__(self, *args, **kwargs):
```
lukashergt: I think the `__init__` function is conventionally placed first in a class. Is there a reason you put it further down? I was a bit confused when I was looking for it and didn't find it right away...

williamjameshandley: The latest push should resolve this

williamjameshandley:
Closes #54

williamjameshandley:
Closes #44

PR #56, opened 2019-09-10
changing beta now changes the weight column

williamjameshandley:
# Description

Following from #55, in order to maintain synchronisation of the `'weight'` column the setting of the column has been moved to `set_beta`. Tests have also been added to check that setting beta has the correct behaviour

# Checklist:


anesthetic/samples.py
```diff
@@ -432,6 +428,9 @@ def beta(self, beta):
         self._beta = beta
         logw = self._dlogX() + self.beta*self.logL
         self._weight = numpy.exp(logw - logw.max())
+        if self.weight is not None:
+            self['weight'] = self.weight
+            self.tex['weight'] = r'MCMC weight'
 
     def set_beta(self, beta, inplace=False):
         """Change the inverse temperature.
```
lukashergt: Using the setter decorator and a method `set_beta` is slightly unpythonic. We need `set_beta` to get a copy for the prior, right?

Would it maybe make more sense to define `__copy__(beta=1)` to make a copy with any given beta?

williamjameshandley: Can `__copy__` take arguments like that? 

I see what you're saying. The two use cases I envisage are
```python
prior = samples.set_beta(0)
# or
prior = samples.copy()
prior.beta = 0
```
Perhaps a different name to `set_beta`? Perhaps `change_temperature` or something more pithy?

lukashergt: yeah, doesn't work with `__copy__`...

It would be possible to overwrite `NDFrame`'s version of `copy`:
```python
def copy(self, beta=1):
    data = super(NestedSamples, self).copy()
    data.beta = beta
    return data
```

Then one could copy the data as `prior = samples.copy(beta=0)`.

lukashergt: maybe also keep `NDFrame`'s kwarg `deep=True`

williamjameshandley: I'm not sure having beta as an argument to copy is very clear at all in terms of an API. I think I definitely want a 'function that explicitly changes the temperature and returns a new one'. There are plenty of these `set_...` in the pandas dataframe, but I think you're right it's arguably a little confusing to have both a `@setter` for `beta` and a `set_beta`.

PR #57, opened 2019-09-11
Added tests to weighted pandas

williamjameshandley:
# Description

Weighted pandas has been missing tests for a while, so this adds some rudimentary tests to check that the outputs have the correct format and the means are roughly correct.

The organisation of the classes has also been improved to reduce code repetition by using multiple inheritance for the functions that are similar between `WeightedSeries` and `WeightedDataFrame`.

# Checklist:


tests/test_weighted_pandas.py
```diff
@@ -1,12 +1,28 @@
...
 import numpy
-from numpy.testing import assert_array_equal
+from numpy.testing import assert_array_equal, assert_allclose
+
+
+def test_WeightedSeries_constructor():
+    numpy.random.seed(0)
+    N = 100000
+    data = numpy.random.rand(N)
+    weights = numpy.random.rand(N)
```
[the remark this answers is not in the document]
williamjameshandley: 56edff6 I believe implements this

tests/test_weighted_pandas.py
```diff
@@ -23,7 +61,106 @@ def test_WeightedDataFrame_constructor():
...
+        assert isinstance(quantile, Series)
+        assert_allclose(quantile, q, atol=1e-2)
+
+
+def test_WeightedDataFrame_neff():
+    df = test_WeightedDataFrame_constructor()
+    neff = df.neff()
+    assert isinstance(neff, float)
+    assert neff < len(df)
+    assert neff > len(df) * numpy.exp(-0.25)
```
lukashergt: Is this last line generally true? Where is the `-0.25` coming from?
If we were to reduce `len(df)` to 100 this would be false quite frequently, right?

williamjameshandley: It's definitely more of a sanity check that neff is roughly the right amount, and it would only be true for large dataframes like the one in this case.

lukashergt: ok

PR #61, opened 2019-09-19
Evidence and merging functionality

williamjameshandley:
# Description

There is now a function `MCMCSamples.logZ` that explicitly computes the
evidence (rather than relying on the rather cryptically named `ns_output`),
either as a mean value or samples.

Also added `merge_nested_samples` which consistently merges nested samples.

# Checklist:


anesthetic/samples.py
```diff
@@ -470,22 +457,63 @@ def ns_output(self, nsamples=200):
...
+        samples['D'] = numpy.exp(logsumexp(logw, b=S, axis=0))
+        samples['d'] = numpy.exp(logsumexp(logw, b=(S-samples.D)**2, axis=0))*2
+
+        samples.tex = {'logZ': r'$\log\mathcal{Z}$',
+                       'D': r'$\mathcal{D}$',
+                       'd': r'$d$'}
+        samples.label = self.label
+        return samples
+
+    def logZ(self, nsamples=None):
```
lukashergt: Do the methods `self.logZ()`, `self.D()` and `self.d()` now conflict with the corresponding attributes `self.logZ` etc. created by `ns_output()`? Do we accept that or would it be better to rename the methods, e.g. `self.get_logZ()` or `self.calc_logZ()`?

williamjameshandley: Fortunately they don't conflict. The object returned from `ns_output` is an `MCMCSamples` object, not a `NestedSamples` one, so there shouldn't be a conflict.

tests/test_samples.py
```diff
@@ -424,6 +434,21 @@ def test_masking():
...
 
+def test_merging():
+    numpy.random.seed(3)
+    samples_1 = NestedSamples(root='./tests/example_data/pc')
+    samples_2 = NestedSamples(root='./tests/example_data/pc_250')
+    samples = merge_nested_samples([samples_1, samples_2])
+    nlive_1 = samples_1.nlive.mode()[0]
+    nlive_2 = samples_2.nlive.mode()[0]
+    nlive = samples.nlive.mode()[0]
+    assert nlive == nlive_1 + nlive_2
```
lukashergt: I know there is the test `test_utils.test_compute_nlive`, however, there the test data is generated by `numpy.random.rand` and that seems somewhat simplified compared to PolyChord output. 

This might be just me not as familiar with the code, but maybe it would be worth to **add a test that checks whether `samples_1.nlive.mode()[0]` is indeed what we expect from the example data?**

lukashergt: I'm also unsure at a glance, why `.mode()` is the right thing to use to get `nlive`. Is that because throughout a long PC run the number of live points will almost always be `nlive`?

williamjameshandley: Does [the latest commit](https://github.com/williamjameshandley/anesthetic/pull/61/commits/70842b1538bfffaf3a6ef02e3bf8d2e473b83645) give you the functionality you were expecting?

williamjameshandley: > I'm also unsure at a glance, why .mode() is the right thing to use to get nlive. Is that because throughout a long PC run the number of live points will almost always be nlive?

For these runs, nlive starts at e.g. 125, remains constant until the very end, where then it drops linearly due to the impact of the final set of live points:

Figure_1

The mode is just a quick-and-easy way to get the nlive of the main part of the run

lukashergt: Nice plot, I like how you can see the different drop offs in the combined sample. 

Ok, this makes sense then and I do like the hard coded numbers in there. Though not great in the code itself they do have a place in unit testing.

williamjameshandley: Pretty easy to do as well:
```python
    samples_1 = NestedSamples(root='./tests/example_data/pc')
    samples_2 = NestedSamples(root='./tests/example_data/pc_250')
    samples = merge_nested_samples([samples_1, samples_2])

    samples_1.nlive.plot(label='samples_1')
    samples_2.nlive.plot(label='samples_2')
    ax = samples.nlive.plot(label='samples')
    ax.set_ylabel('nlive')
    ax.legend()

```

williamjameshandley: Here's an alternative view on it, plotting against compression rather than iteration

```python
    samples_1 = NestedSamples(root='./tests/example_data/pc')
    samples_2 = NestedSamples(root='./tests/example_data/pc_250')
    samples = merge_nested_samples([samples_1, samples_2])

    plt.plot((1/samples_1.nlive).cumsum(),samples_1.nlive, label='samples_1')
    plt.plot((1/samples_2.nlive).cumsum(),samples_2.nlive, label='samples_2')
    plt.plot((1/samples.nlive).cumsum(),samples.nlive, label='samples')
    plt.xlabel('compression')
    plt.ylabel('nlive')
    plt.legend()
```

Figure_1

Issue #38, opened 2019-07-17
MWE - NestedSamples() argument 'columns'


xichenamoy:
Adding the argument 'columns = [str]' in function NestedSamples() seems doesn't work well to assign new string/names to the column parameters.  

I tried 
sample = NestedSamples(root=xx)
sample.columns = ['theta', 'beta', 'logL', 'nlive', 'weight']
This would work.

lukashergt:
This is a bit dated, now, but I find that renaming works. One needs to watch out that when providing a list the length of the list needs to match the length of the data.

```python
ns = NestedSamples(root="/scratch/Projects/AnestheticPrj/anesthetic/tests/example_data/pc")
ns.columns = ['y0', 'y1', 'y2', 'y3', 'y4', 'y5', 'y6', 'y7', 'y8']
ns.columns.values[1] = 'z1'
```

This will change the column names from 
```python
['x0', 'x1', 'x2', 'x3', 'x4', 'logL', 'logL_birth', 'nlive', 'weight']
```
as read from the header in the example file to 
```python
['y0', 'z1', 'y2', 'y3', 'y4', 'y5', 'y6', 'y7', 'y8']
```

Or did I misunderstand the problem?

williamjameshandley:
I've encountered this issue in the past when loading chains that lack a `.paramnames` file. The code I'd ideally write would be
```python
ns = NestedSamples(root="<file root>", columns=['here', 'are', 'the', 'parameters'])
```
but at the moment the `root` keyword overrides all the others.

>```python
>ns.columns.values[1] = 'z1'
>```

This is a neat trick, although far from obvious for people who are only just starting out with pandas.

lukashergt:
> The code I'd ideally write would be
> 
> ```python
> ns = NestedSamples(root="<file root>", columns=['here', 'are', 'the', 'parameters'])
> ```

I see, so could this solve it?
```diff
             params, tex = reader.paramnames()
+            columns = kwargs.pop('columns', params)
             limits = reader.limits()
             kwargs['label'] = kwargs.get('label', os.path.basename(root))
-            self.__init__(data=samples, columns=params, w=w, logL=logL,
+            self.__init__(data=samples, columns=columns, w=w, logL=logL,
```

williamjameshandley:
That looks like a pretty good solution. Should be a pretty quick PR if you want to submit one (ideally with a test)

lukashergt:
Fixed in #64

Issue #37, opened 2019-07-17
MCMCSamples cannot read nested sampling output files correctly.


xichenamoy:
Happened when execute this line:

sample = MCMCSamples(root=fileName)

Then get this error message:

ValueError: DataFrame constructor not properly called!

One thing worth to note is that 1.1.0 works well with the provided lensing example. So the error only happens when I change the 'root' equals to my own file name in the chains folder.

I also tried the standard PolyChord output file using the default gaussian example, it would report:

ValueError: Wrong number of items passed 5, placement implies 1

Previously when I used it with anesthetic 1.03, it can read the file ok. Now I upgraded it to version 1.1.0, run the same code it would report this.  The output files are the standard MultiNest/PolyChord output files.

williamjameshandley:
Hi @xichenamoy . Many thanks for spotting this. 

Can you confirm that this only occurs if you try to use read nested sampling chains with `MCMCSamples`, and that reading them with `NestedSamples` does not have this issue?

@lukashergt, I think this was introduced in #14 when we restructured in terms of `SampleReader`. It might be worth just raising a more informative `ValueError` if one tries to read ns chains with `MCMCSamples`

xichenamoy:
@williamjameshandley 
Yes, I forgot to mention this. It works well with NestedSamples(), only occurs in MCMCSamples().

lukashergt:
> I think this was introduced in #14 when we restructured in terms of `SampleReader`

Probably, I never checked for the option of reading nested sampling files with `MCMCSamples`. Is that what we want?

williamjameshandley:
I'm not sure. In theory NS files should be readable as MCMCSamples, but I don't know why you'd want to/if we should encourage it. Given that NestedSamples is a subclass of MCMCSamples you would get no benefit from loading them as MCMCSamples, so at the very least if we want to allow people to do this we should throw a warning. For now, the easiest solution would be to not support reading MN/PC files with MCMCSamples, but to throw a more informative error.

lukashergt:
Fixed in #64

PR #64, opened 2019-11-22
Add columns kwarg

lukashergt:
# Description

This allows column names to be passed manually to both `MCMCSamples` and `NestedSamples` but behaves as usual if only the root is passed.

Addresses #37 and #38.

# Checklist:

### DIFF
--- a/anesthetic/samples.py
+++ b/anesthetic/samples.py
@@ -60,11 +60,18 @@ class MCMCSamples(WeightedDataFrame):
         root = kwargs.pop('root', None)
         if root is not None:
             reader = SampleReader(root)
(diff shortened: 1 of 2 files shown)


lukashergt:
> I see that you've chosen to raise an error rather than a warning if you try to read nested sampling chains with MCMCSamples. This is arguably a little strong, but given that it's manifestly better than the current state of affairs, I'm happy to accept this.

would you prefer if it were a warning?

williamjameshandley:
Many thanks @lukashergt -- that was an impressively efficient contribution with regards to to resolving issues.

PR #65, opened 2019-12-11
Removed vertical slider code

williamjameshandley:
# Description

The [vertical option for slider widgets](https://github.com/matplotlib/matplotlib/pull/10746) 
is now in the stable version of matplotlib, so we no longer need to replicate
it. 

This PR was prompted (via a cron job test) by an update to pydocstyle that
flagged that the now unnecessary code did not have compliant docstrings. I
propose to resolve the issue by removing the code

# Checklist:


williamjameshandley:
The tests pass on python 3.6, 3.7 and 3.8. I propose to take this opportunity to drop support for 3.5 and 2.7, to both speed up the test cycle, and to recognise the [imminent deprecation](https://pythonclock.org/).

.travis.yml
```diff
@@ -5,8 +5,6 @@ python:
     - "3.8"
     - "3.7"
     - "3.6"
-    - "3.5"
-    - "2.7"
```
lukashergt: These python versions should also be taken out of the `setup.py` file and the `README.rst` should be updated accordingly.

williamjameshandley: Well spotted.

lukashergt: stumbled over a python35 environment which didn't work with anesthetic anymore...

williamjameshandley: Can you remember what it was that didn't work?

lukashergt: The newest matplotlib (v3.1.2) which incorporates the vertical slider has dependency python>=3.6. 

Maybe worth incorporating in the requirements? i.e.
```
matplotlib>=3.1.2
```

williamjameshandley: Added as part of #73

PR #67, opened 2019-12-14
logZ computation now includes beta

williamjameshandley:
# Description

The computations provided by #61 don't take into account the thermodynamic temperature. This PR addresses this

# Checklist:


williamjameshandley:
OK, I've added some simple tests that check that beta is behaving as one would expect.

PR #71, opened 2020-01-14
set values <-1e150 to -1e150 to fix overflow issues which happen beca…

### LINKED ISSUE #70: Overflow error because of Cobaya's definition of `logzero=-1e+300`
Cobaya's definition of `logzero=-1e+300` causes Anesthetic to fail with an overflow error upon the calculation of the Bayesian dimensionality. The problematic lines are:

```python
S = (dlogX*0).add(self.logL, axis=0) - samples.logZ
samples['d'] = numpy.exp(logsumexp(logw, b=(S-samples.D)**2, axis=0))*2
```
When `S = -1e+300` then squaring `S` will cause an overflow error.

An easy fix would be to insert:
```python
S = S.where(S > -1e150, -1e150)
```

This can be easily done for the calculation of `d`, but I don't know whether we want to handle these extreme values for `logzero` further upstream?

lukashergt:
# Description

Sets values `<-1e+150` to `-1e+150` to fix overflow issues which happen because of Cobaya's `logzero=-1e+300` definition.

This can be easily done for the calculation of `d`, but I don't know whether we would want to handle these extreme values for `logzero` further upstream?

Fixes #70 

# Checklist:


anesthetic/samples.py
```diff
@@ -396,6 +396,9 @@ def __init__(self, *args, **kwargs):
         if root is not None:
             reader = SampleReader(root)
             samples, logL, logL_birth = reader.samples()
+            logzero = kwargs.pop('logzero', -numpy.inf)
+            logL = numpy.where(logL <= logzero, -numpy.inf, logL)
+            logL_birth = numpy.where(logL_birth <= logzero, -numpy.inf, logL_birth)
```
lukashergt: In #70 you're saying 

> In principle, there shouldn't be any -numpy.inf values in the logL column, since they should be excluded in prior generation 

Why shouldn't there be `-inf` in `logL` and why do I find `-1e300` then? 

Is this what you had in mind with `logzero`?

lukashergt:
@williamjameshandley: I get this weird recursion error when I try to add an attribute to MCMCSamples. Any idea where this might be coming from?

MWE:
In MCMCSamples (using master branch) add a line
```diff
        else:
            logL = kwargs.pop('logL', None)
            self.tex = kwargs.pop('tex', {})
+           self.new_attribute = 1
```
and try running `ns = NestedSamples(root="example_data/pc")` 

---

This gives me a RecursionError:

```python
RecursionError                            Traceback (most recent call last)
<ipython-input-4-dd8c878f4931> in <module>
----> 1 ns = NestedSamples(root="/scratch/Projects/AnestheticPrj/anesthetic/tests/example_data/pc")

/scratch/Projects/AnestheticPrj/anesthetic/anesthetic/samples.py in __init__(self, *args, **kwargs)
    404             self.__init__(data=samples, columns=columns,
    405                           logL=logL, logL_birth=logL_birth,
--> 406                           tex=tex, limits=limits, *args, **kwargs)
    407             self.root = root
    408         else:

/scratch/Projects/AnestheticPrj/anesthetic/anesthetic/samples.py in __init__(self, *args, **kwargs)
    409             self._beta = kwargs.pop('beta', 1.)
    410             logL_birth = kwargs.pop('logL_birth', None)
--> 411             super(NestedSamples, self).__init__(*args, **kwargs)
    412             if logL_birth is not None:
    413                 self._compute_nlive(logL_birth)

/scratch/Projects/AnestheticPrj/anesthetic/anesthetic/samples.py in __init__(self, *args, **kwargs)
     78             logL = kwargs.pop('logL', None)
     79             self.tex = kwargs.pop('tex', {})
---> 80             self.new_attribute = 1
     81             self.limits = kwargs.pop('limits', {})
     82             self.label = kwargs.pop('label', None)

/scratch/.virtualenvs/anestheticenv/lib/python3.7/site-packages/pandas/core/generic.py in __setattr__(self, name, value)
   5202         else:
   5203             try:
-> 5204                 existing = getattr(self, name)
   5205                 if isinstance(existing, Index):
   5206                     object.__setattr__(self, name, value)

/scratch/.virtualenvs/anestheticenv/lib/python3.7/site-packages/pandas/core/generic.py in __getattr__(self, name)
   5175             return object.__getattribute__(self, name)
   5176         else:
-> 5177             if self._info_axis._can_hold_identifiers_and_holds_name(name):
   5178                 return self[name]
   5179             return object.__getattribute__(self, name)

... last 1 frames repeated, from the frame below ...

/scratch/.virtualenvs/anestheticenv/lib/python3.7/site-packages/pandas/core/generic.py in __getattr__(self, name)
   5175             return object.__getattribute__(self, name)
   5176         else:
-> 5177             if self._info_axis._can_hold_identifiers_and_holds_name(name):
   5178                 return self[name]
   5179             return object.__getattribute__(self, name)

RecursionError: maximum recursion depth exceeded
```

anesthetic/utils.py
```diff
@@ -1,11 +1,24 @@
...
+def logsumexpinf(a, b, **kwargs):
+    """Handles -inf in weighted logsumexp.
+
+    If a=-inf in `log(sum(b * exp(a))` then we can set b=0, using:
+
+    limit x->-oo: x * exp(x) = 0
+
+    """
+    b = numpy.where(a == -numpy.inf, 0, b)
+    return logsumexp(a, b=b, **kwargs)
```
lukashergt: Does this function make sense? I'm effectively setting the terms in the sum of logsumexp with `a=-inf` to zero.

williamjameshandley:
>@williamjameshandley: I get this weird recursion error when I try to add an attribute to MCMCSamples. Any idea where this might be coming from?

This is why [it's not recommended](https://pandas.pydata.org/pandas-docs/stable/development/extending.html#subclassing-pandas-data-structures) to subclass pandas (as `anesthetic does`), as it's quite a challenge to not get it wrong. You need to add your new attribute to the metadata list (have a look at `_beta` for `NestedSamples`, or `limits` for `MCMCSamples`)

lukashergt:
> This is why [it's not recommended](https://pandas.pydata.org/pandas-docs/stable/development/extending.html#subclassing-pandas-data-structures) to subclass pandas (as `anesthetic does`), as it's quite a challenge to not get it wrong. You need to add your new attribute to the metadata list (have a look at `_beta` for `NestedSamples`, or `limits` for `MCMCSamples`)

What would you prefer?
* current setup (works): 
  - take kwarg `logzero`
  - set `logL` and `logL_birth` to `-np.inf` where they are `<= logzero` (e.g. convert all `-1e300` to `-np.inf`)
  - test for `a == -numpy.inf` in `logsumexpinf`

or
* alternative:
  - leave `logL` and `logL_birth` as they are
  - create attribute `self.logzero` in `MCMCSamples`
  - test for `a <= self.logzero` in `logsumexpinf`

anesthetic/samples.py
williamjameshandley: >Is this what you had in mind with logzero?

Yes

>Why shouldn't there be -inf in logL and why do I find -1e300 then?

In theory, samples with logzero loglikelihood should be rejected at prior generation. How were these samples generated?

anesthetic/utils.py
williamjameshandley: I'm pretty sure it does. This is similar to the idea that `0*log(0)==0`, i.e. `(anything)*exp(-inf)=0`. Ideally this function should have the same call signature as scipy.logsumexp, and I'd probably just call it logsumexp, and note that `anesthetic.utils.logsumexp` has a slightly more sensible default.

williamjameshandley:
>* current setup (works): 
>  - take kwarg `logzero`
>  - set `logL` and `logL_birth` to `-np.inf` where they are `<= logzero` (e.g. convert all `-1e300` to `-np.inf`)
>  - test for `a == -numpy.inf` in `logsumexpinf`

This is how I would do it personally, but am open to be being persuaded otherwise. I would also have `-1e30` as the default.

anesthetic/samples.py
lukashergt: > In theory, samples with logzero loglikelihood should be rejected at prior generation. How were these samples generated?

This is output from a Cobaya+PC run...

anesthetic/utils.py
lukashergt: > Ideally this function should have the same call signature as scipy.logsumexp, and I'd probably just call it logsumexp, and note that `anesthetic.utils.logsumexp` has a slightly more sensible default.

Adapted accordingly in [bfe6cfb](https://github.com/williamjameshandley/anesthetic/pull/71/commits/bfe6cfbadface8b1ed56142df2fc0a51490efde0).

lukashergt:
I've added a test for `utils.logsumexp`. For further tests, we'd need data with `-inf` or `-1e300` in the `logL` column. Should we add that?

Otherwise, ready for review once tests have passed.

anesthetic/samples.py
```diff
@@ -389,13 +388,24 @@ class NestedSamples(MCMCSamples):
...
+    logzero: float
+        The threshold for `log(0)` values assigned to rejected sample points.
+        Anything equal or below this value is set to `-numpy.inf`.
+        default: -numpy.inf
+
     """
 
     def __init__(self, *args, **kwargs):
         root = kwargs.pop('root', None)
+        logzero = kwargs.pop('logzero', -numpy.inf)
```
williamjameshandley: I would probably have this as ``-1e30`` as a default, since that's what both MultiNest and PolyChord use.

anesthetic/samples.py
```diff
@@ -389,13 +388,24 @@ class NestedSamples(MCMCSamples):
...
         if root is not None:
             reader = SampleReader(root)
             samples, logL, logL_birth = reader.samples()
+            logL = numpy.where(logL <= logzero, -numpy.inf, logL)
+            if not isinstance(logL_birth, int):
+                logL_birth = numpy.where(logL_birth <= logzero,
+                                         -numpy.inf,
+                                         logL_birth)
```
williamjameshandley: Do we need the newline here?

anesthetic/utils.py
```diff
@@ -1,10 +1,31 @@
 """Data-processing utility functions."""
 import numpy
 import pandas
+from scipy import special as sp
```
williamjameshandley: I would just use ``import scipy.special`` and avoid the `sp` abbreviation. I'm not sure how standard this is, and it's more in keeping with the rest of `anesthetic`

anesthetic/samples.py
williamjameshandley: Better still would be to use the lowest value in `NestedSamples.logL_birth`

lukashergt: > I would probably have this as `-1e30` as a default, since that's what both MultiNest and PolyChord use.

Would `<=-1e29` be safer to handle potential conversion issues?
I have spotted some `-0.99e30` values in the first column of a `_dead.txt` file. What does that column correspond to? Such values would be missed by a `-1e30` threshold but caught by a `-1e29` threshold...

> Better still would be to use the lowest value in `NestedSamples.logL_birth`

Is it certain that there is a `logzero` value in `logL_birth`? Wouldn't want to declare a physically meaningful value to be logzero...

anesthetic/samples.py
lukashergt: flake8 requires one new line, but not necessarily two, i.e. we could lift either `-numpy.inf` or `logL_birth` up by one line. Preference?

anesthetic/utils.py
lukashergt: Good point, `sp` is totally unnecessary here. Taken care of in [7b893da](https://github.com/williamjameshandley/anesthetic/pull/71/commits/7b893dab10e1be9410ec4ff814d9ac663455c47d).

On another note, though, I personally would be very much in favour of using `numpy as np`. Have often stumbled over writing `np` first and then having to correct it. I think that `np` is a very widely used abbreviation and this would also shorten the code and make it easier to keep things on one line and hence more readable.

lukashergt:
> It might be good to have a direct test inside `test_build_mcmc` in the `tests/test_samples.py`that checks that it's doing the right thing (no need to generate new files).

I have only introduced `logzero` for `NestedSamples` so far. Should I also introduce it for `MCMCSamples`? The original problem is mostly connected to calculations of `D` and `d`...

While there is a `test_build_mcmc` there is no `test_build_ns`. Is that deliberate?

anesthetic/utils.py
williamjameshandley: This isn't quite the same signature, since you've converted the positional `*args` to `**kwargs`, which means that if somebody tried calling `anesthetic.utils.logsumexp(a, axis)` they would get unexpected behaviour

anesthetic/samples.py
williamjameshandley: >I have spotted some -0.99e30 values in the first column of a _dead.txt file. What does that column correspond to? Such values would be missed by a -1e30 threshold but caught by a -1e29 threshold...

That's annoying. You could use a `numpy.isclose` test?

>Is it certain that there is a logzero value in logL_birth? Wouldn't want to declare a physically meaningful value to be logzero...

￼I would have thought so, although it seems like things can get garbled by the wrapper that's calling nested sampling (e.g. Cobaya)

I'm starting to wonder how useful this conversion is, and whether or not the modified logsumexp would be enough to fix the original issue in #71

anesthetic/samples.py
williamjameshandley: I would have lifted `logL_birth`

anesthetic/utils.py
williamjameshandley: I know. It's really just my idiosyncrasy to avoid using `np`. You should raise it as a separate issue to remind us to possibly make this wholesale conversion (which shouldn't be too difficult to do).

williamjameshandley:
>I have only introduced logzero for NestedSamples so far. Should I also introduce it for MCMCSamples? The original problem is mostly connected to calculations of D and d...

I don't think logzero would affect any of the `MCMCSamples` functionality, so I'd keep it confined to `NestedSamples`

> While there is a test_build_mcmc there is no test_build_ns. Is that deliberate?

Only in that everything that tests is equally valid for `NestedSamples` since it's a subclassed type. Really it should probably loop `for typ in [MCMCSamples, NestedSamples]:`

anesthetic/samples.py
lukashergt: > That's annoying. You could use a `numpy.isclose` test?

I think `<=threshold` is more practical. That way I could specify `logzero=-1e30` as the threshold and it would work for multiple definitions, e.g. for both `-1e30` and `-1e300`. With `numpy.isclose` I would also have to pick some tolerance. I could simply apply that to the lower bound and test for `<=-1e29` or `<=-9e29`. Just a question of which tolerance to go for...?


> I'm starting to wonder how useful this conversion is, and whether or not the modified logsumexp would be enough to fix the original issue in #71

The new logsumexp currently tests against `-np.inf`. This only works if all logzero values have been converted. Otherwise `-1e300` values can still cause errors.

Also, part of the original problem was that `(-1e300)**2` was causing overflow errors completely independent of logsumexp.

anesthetic/samples.py
lukashergt: Done in [4faf61e](https://github.com/williamjameshandley/anesthetic/pull/71/commits/4faf61e4a4f600986f6dfbefaab81810b0accbbe).

anesthetic/utils.py
lukashergt: Corrected signature in [5ef79fa](https://github.com/williamjameshandley/anesthetic/pull/71/commits/5ef79fa69a8276d1f68fa0c8d2f39cb8fc3d654e).

anesthetic/utils.py
lukashergt: > You should raise it as a separate issue to remind us to possibly make this wholesale conversion

Done in #76.

lukashergt:
> I don't think logzero would affect any of the `MCMCSamples` functionality, so I'd keep it confined to `NestedSamples`

Added test to `test_build_mcmc` for using logzero kwarg in NestedSamples.

tests/test_samples.py
```diff
@@ -59,6 +59,10 @@ def test_build_mcmc():
     for p in params:
         assert(mcmc.limits[p] == limits[p])
 
+    ns = NestedSamples(data=samples, logzero=-1e25)
+    assert(len(ns) == nsamps)
+    assert_array_equal(ns.columns, numpy.array([0, 1, 2], dtype=object))
+
```
williamjameshandley: Could you perhaps also add some tests here with a `logL_birth` and `logL` argument, with a check that the logzeros are being modified correctly?

anesthetic/samples.py
williamjameshandley: OK, agreed that we do need this functionality. The thresholding should be through an rtol, since this should only really occur with a floating-point type copying error. Perhaps at single precision (i.e. an rtol of 1e-8)?

anesthetic/samples.py
williamjameshandley: OK, we should keep on the lookout for any issues this might potentially cause, although it doesn't seem to have reared its head yet.

anesthetic/samples.py
lukashergt: `rtol=1e-8` wouldn't even catch `-0.99e30`, so in that case I wouldn't bother with a `numpy.isclose`. 

Should I just set the default threshold to `-1e30` then and if needed it can be adjusted manually (athough as long as the really extreme values `-1e300` are taken care of, there shouldn't even be any problems anymore).

williamjameshandley: Are the values `-0.99e30` or `-0.9999999999e30`? If it's the former, then it might be worth investigating if this is something that Cobaya is doing deliberately (say by increasing from logzero a little).

lukashergt: It's `-0.99e30`.

tests/test_samples.py
lukashergt: Added test in [1e85d65](https://github.com/williamjameshandley/anesthetic/pull/71/commits/1e85d65efa54307bd4bafc368d78b91c5302885f) where I modified `logL` such that it has some `-1e300` values, which get turned into `-numpy.inf`.

Good call. Had to move the conversion, because it was previously only affecting calls to NestedSamples with root passing.

williamjameshandley: Excellent. Ideally we should also check that if you pass ``logzero=-1e301`` it doesn't change anything, and also that the rest of the logL column is unchanged.

anesthetic/samples.py
williamjameshandley: Ah, then this quite a specific cobaya issue. In ``cobaya.samplers.polychord.polychord.polychord.run`` it's return of `logpost` (equivalent to `loglike`) has the statement:

```python
 return max(logposterior + self.logvolume, 0.99 * self.pc_settings.logzero), derived
```

i.e. for some reason it's instructed to never return below logzero, which is actually quite counterproductive. @JesusTorrado could you shed some light on why cobaya puts in this factor of 0.99 ?

williamjameshandley: I'm also not sure what the logvolume part is about. Nested sampling expects a likelihood, and the prior transformation in ``pc_prior`` should handle any normalisations like that. As written at the moment I think the log evidence would be incorrect.

williamjameshandley: Ah, or perhaps it's undoing a volume transformation that is already done in the `model.logposterior` call, so that a likelihood is supplied (in which case this is the correct thing to do). That would make more sense. @JesusTorrado can you confirm?

tests/test_samples.py
lukashergt: Done in [2a16556](https://github.com/williamjameshandley/anesthetic/pull/71/commits/2a1655660982d74f9682076ddff53d882841c384).

anesthetic/samples.py
williamjameshandley: OK, I've opened an issue on [cobaya](https://github.com/CobayaSampler/cobaya/issues/77). As far as I can tell, Cobaya is effectively using -0.99e30 as its logzero value, so I believe the correct thing to do here is just to pass `logzero=-0.99e30` as the definition if the chains have come from cobaya.

anesthetic/samples.py
```diff
@@ -406,8 +411,14 @@ def __init__(self, *args, **kwargs):
...
             self._beta = kwargs.pop('beta', 1.)
+            logL = kwargs.pop('logL', None)
             logL_birth = kwargs.pop('logL_birth', None)
-            super(NestedSamples, self).__init__(*args, **kwargs)
+            if logL is not None:
+                logL = numpy.where(logL <= logzero, -numpy.inf, logL)
+            if not isinstance(logL_birth, int) and logL_birth is not None:
+                logL_birth = numpy.where(logL_birth <= logzero,
+                                         -numpy.inf, logL_birth)
+            super(NestedSamples, self).__init__(logL=logL, *args, **kwargs)
```
williamjameshandley, line 421: This code looks great. Is there any reason that this functionality for `logL` should be `NestedSamples` specific? It could easily be moved up into the parent `MCMCSamples` class. Is there a reason it shouldn't?

anesthetic/utils.py
```diff
@@ -1,10 +1,32 @@
...
+    Thus, if a=-inf in `log(sum(b * exp(a))` then we can set b=0 such that
+    that term is ignored in the sum.
+    """
+    if b is None:
+        b = numpy.ones_like(a)
+    b = numpy.where(a == -numpy.inf, 0, b)
+    return special.logsumexp(a, axis=axis, b=b, keepdims=keepdims,
+                             return_sign=return_sign)
+
+
```
williamjameshandley, line 29: Nice.

tests/test_utils.py
```diff
@@ -60,3 +63,19 @@ def test_is_int():
...
+    assert logsumexp(-numpy.inf, b=[-numpy.inf]) == -numpy.inf
+    assert logsumexp(a, b=b) == sp.logsumexp(a, b=b)
+    a[0] = -numpy.inf
+    assert logsumexp(a, b=b) == sp.logsumexp(a, b=b)
+    b[0] = -numpy.inf
+    with warnings.catch_warnings():
+        warnings.filterwarnings('ignore',
+                                'invalid value encountered in multiply',
+                                RuntimeWarning)
+        assert numpy.isnan(sp.logsumexp(a, b=b))
```
williamjameshandley, line 80: This is an excellent test.

anesthetic/samples.py
lukashergt: It's been a while, so not 100% sure, but I think that should work just as well for `MCMCSamples`. Probably not needed there, since it doesn't go to as extreme values...?

williamjameshandley: I guess that a metropolis hastings sampler would never accept 'logzero' values. However, First, these are `MCMCSamples`, not `MetropolisHastingsSamples`, and are therefore meant to be a bit more general than this. Second, I think this functionality could be useful for some analyses that like to increase the 'unphysical' cutoff artificially (such as capped likelihoods). I think it would be worth pushing this further up the hierarchy.

lukashergt: Actually we had that discussion [above](https://github.com/williamjameshandley/anesthetic/pull/71#issuecomment-579385104):

>> I have only introduced logzero for NestedSamples so far. Should I also introduce it for MCMCSamples? The original problem is mostly connected to calculations of D and d...
> 
> I don't think logzero would affect any of the MCMCSamples functionality, so I'd keep it confined to NestedSamples

Do you still prefer moving it up in the hierarchy, now?

williamjameshandley: I think I do prefer moving it up now, although if you disagree you can have the deciding vote between the three of us (Me, you and past me).

lukashergt: moved up in hierarchy

lukashergt: Do we need to add tests for using `logzero` with MCMCSamples?

lukashergt: done anyhow

williamjameshandley:
OK, Travis has been misbehaving on account of its migration from travis-ci.org to travis-ci.com. Is it OK if we approve and merge #77 first, that way we'll get an updated travis build when we merge master into this branch, and hopefully a correct codecov.

lukashergt:
> Excellent. This looks ready for @lukashergt to squash and merge

Should there be an update to the version number?

williamjameshandley:
We can treat #77 and this one to be part of the same [patch](https://semver.org/). Current latest version on PyPi is 1.2.4. After this is merged, I'll release it as version 1.2.5 (which is what it says in the README), and upload to PyPi,

williamjameshandley:
The updated version number will be [1.3.0](https://github.com/williamjameshandley/anesthetic/milestone/3) once I've updated the README and plot gallery.

Issue #69, opened 2020-01-13
Plotting n-sigma contours


Pablo-Lemos:
It would be useful to plot n-sigma contours in the 2D marginalised plots, instead of only 1 and 2 sigma.

lukashergt:
With #216 this should now be possible (see corresponding plots in #216 discussion).

PR #73, opened 2020-01-23
Partial fix to KDE narrow contours

### LINKED ISSUE #68: 2D KDE contours perform poorly on highly correlated, narrow samples
The 2D `'kde'` plots are very bad for highly correlated, narrow samples. "Bananas" are particularly bad. 

Even when smoothing the second parameter with `  + np.random.randn(n) * 0.05`, the contours are still rather bad.

```python
n = 1000
dim = 2
samples = np.empty((n, dim))
samples[:, 0] = np.random.exponential(1, n)
samples[:, 1] = np.log(samples[:, 0])# + np.random.randn(n) * 0.05
limits = {'0': (0, samples[:, 0].max()), '1': (None, None)}

ns = NestedSamples(data=samples, limits=limits)

fig, axes = make_2d_axes([0, 1])
ns.plot_2d(axes)
```

2D KDE contours perform poorly on highly correlated, narrow samples

williamjameshandley:
# Description

This is not a full fix to the issue, since a simple Gaussian KDE as provided by
``scipy.stats.gaussian_kde`` will never be able to tackle these kind of problems. However, the resulting plots now look over-fattened, rather than wrong. I am the upcoming PR that looks at #69 should address some more of these issues.

Fixes #68

# Checklist:


williamjameshandley:
```python
import numpy as np                                                                               
n = 1000                                                                                         
dim = 2                                                                                          
samples = np.empty((n, dim))                                                                     
samples[:, 0] = np.random.exponential(1, n)                                                      
samples[:, 1] = np.log(samples[:, 0])                                                            
limits = {0: (0, samples[:, 0].max())}                                                           
                                                                                                 
ns = NestedSamples(data=samples, limits=limits)                                                  
                                                                                                 
fig, axes = make_2d_axes([0, 1])                                                                 
ns.plot_2d(axes)                                                                                 
```

Figure_1

anesthetic/utils.py
```diff
@@ -210,17 +209,43 @@ def iso_probability_contours_from_samples(pdf, contours=[0.68, 0.95],
...
+    return Triangulation(x, y, tri.triangles)
+
+
+def triangular_sample_compression_2d(x, y, cov, w=None, n=1000):
     """Histogram a 2D set of weighted samples via triangulation.
 
-    This defines bins via a triangulation of the subsamples, sums weights
-    within triangles, and computes weighted centroids of triangles.
+    This defines bins via a triangulation of the subsamples and sums weights
+    within triangles surrounding point each point
```
lukashergt: `point each point`

typo?

williamjameshandley: Fixed, and also added test to check this function behaves as appropriate

williamjameshandley:
> Should there be tests added?

This is a difficult one. I'm not sure if there is a reasonable way to create tests for these kind of plots (particularly given that we know it doesn't produce a very good answer in these cases).

lukashergt:
> This is a difficult one. I'm not sure if there is a reasonable way to create tests for these kind of plots (particularly given that we know it doesn't produce a very good answer in these cases).

What was the major change? The adding of the covariance matrix? Can we come up with a test case for `cov`?

williamjameshandley:
So the covariance matrix is just a re-factoring (previously that was computed in `triangular_sample_compression_2d` but now is computed further out, since it needs to be reused again).

The main alteration is:

```python
    x, y = kde.resample(ncompress)
    x = numpy.concatenate([tri.x, x])
    y = numpy.concatenate([tri.y, y])
    w = numpy.concatenate([w, numpy.zeros(ncompress)])
    tri = scaled_triangulation(x, y, cov)

    p = kde([tri.x, tri.y])
```

i.e. you generate extra plotting points with zero weight from the kde itself. This means that you don't get the large triangles, and ensure that the area where the kde is non-zero is well represented.

PR #75, opened 2020-01-28
Version 1.3

williamjameshandley:
Version bump PR for the upcoming [1.3 release](https://github.com/williamjameshandley/anesthetic/milestone/3)


williamjameshandley:
Given we're about to have some more significant changes from #83 #82 #81 #86, I think it's past time that we signify all of the updates since 1.2 with a version bump

PR #77, opened 2020-04-01
Bug fix for histogram plotting

williamjameshandley:
# Description

Recently I've been using the 2d-histogram plotting, and found a couple of bugs:

- When plotting with samples drawn from a wide range but with low weight (i.e. a typical nested sampling run), the axes resized in a way that was annoying, and iconsistent with the way that ``plot_hist_1d works``. This PR fixes this, and supplies a test that fails with the old code, but passes with the update
- There is a 'non-zero edge' correction in `iso_probability_contours` that apparently doesn't do anything except break in certain difficult to define cases. Removing it makes it more consistent with the behaviour of `iso_probability_contours_from_samples` and does not break. Will reinstate this code if it does become clear that it was serving a specific purpose.

Fixes # (issue)

# Checklist:

### DIFF
--- a/anesthetic/plot.py
+++ b/anesthetic/plot.py
@@ -600,6 +600,18 @@ def hist_plot_2d(ax, data_x, data_y, *args, **kwargs):
     label = kwargs.pop('label', None)
     levels = kwargs.pop('levels', None)
     color = kwargs.pop('color', next(ax._get_lines.prop_cycler)['color'])
+    weights = kwargs.pop('weights', None)
+
+    if xmin is None:
+        xmin = quantile(data_x, 0.01, weights)
+    if xmax is None:
+        xmax = quantile(data_x, 0.99, weights)
+    if ymin is None:
+        ymin = quantile(data_y, 0.01, weights)
+    if ymax is None:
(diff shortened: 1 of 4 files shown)


williamjameshandley:
Fixes #78

anesthetic/plot.py
```diff
@@ -600,20 +600,31 @@ def hist_plot_2d(ax, data_x, data_y, *args, **kwargs):
...
+    weights = kwargs.pop('weights', None)
+
+    if xmin is None or not numpy.isfinite(xmin):
+        xmin = quantile(data_x, 0.01, weights)
+    if xmax is None or not numpy.isfinite(xmax):
+        xmax = quantile(data_x, 0.99, weights)
+    if ymin is None or not numpy.isfinite(ymin):
+        ymin = quantile(data_y, 0.01, weights)
+    if ymax is None or not numpy.isfinite(ymax):
+        ymax = quantile(data_y, 0.99, weights)
```
lukashergt, line 612: We'd need this for `hist_plot_1d`, too.

williamjameshandley: FIxed by 02da2b6

PR #80, opened 2020-05-21
automatic minmax limits

### LINKED ISSUE #36: Do we need .ranges files?
Currently we apply boundary corrections (#35) based on user-specified edges in the `MCMCSamples.limits` dictionary, often input via a `.ranges` file. This has the advantage of being unambiguous, but is limited in that it does not control for edge effects that occur for "top-hat distributions."

One alternative that has occurred to me is to use the outermost samples as the definition of the edge of the distribution. We would then never plot the KDE in a region for which there are no samples, which doesn't seem particularly unreasonable, and the edge corrections may be able to handle top-hats as well as hard prior bounds. 

This would be very easy to implement for the 1D case, and would obviate the need for xmin and xmax (since they could be read from the samples used to construct the KDE). 2D would require a lot more thought/research, but two-dimensional KDE is generally less damaged by incorrect edge treatments.

Discussion welcome

lukashergt:
Automatically set all limits not specified by a `.ranges` file to their respective minimum and maximum values in the sample.

Fixes #36 

# Checklist:

### DIFF
--- a/anesthetic/samples.py
+++ b/anesthetic/samples.py
@@ -97,9 +97,9 @@ class MCMCSamples(WeightedDataFrame):
                 self['weight'] = self.weight
                 self.tex['weight'] = r'MCMC weight'
 
-            self.set_automatic_limits()
+            self._set_automatic_limits()
 
-    def set_automatic_limits(self):
+    def _set_automatic_limits(self):
         """Set all unassigned limits to min and max of sample."""
         for param in self.columns:
             if param not in self.limits:
@@ -437,7 +437,7 @@ class NestedSamples(MCMCSamples):
             if logL_birth is not None:
                 self._compute_nlive(logL_birth)
 
-            self.set_automatic_limits()
+            self._set_automatic_limits()
 
     @property
     def beta(self):
--- a/anesthetic/samples.py
+++ b/anesthetic/samples.py
@@ -89,10 +89,6 @@ class MCMCSamples(WeightedDataFrame):
             self.root = None
             super(MCMCSamples, self).__init__(*args, **kwargs)
 
-            for param in self.columns:
-                if param not in self.limits:
-                    self.limits[param] = (self[param].min(), self[param].max())
-
             if logL is not None:
                 self['logL'] = logL
                 self.tex['logL'] = r'$\log\mathcal{L}$'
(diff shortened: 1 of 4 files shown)


anesthetic/samples.py
```diff
@@ -89,6 +89,10 @@ def __init__(self, *args, **kwargs):
             self.root = None
             super(MCMCSamples, self).__init__(*args, **kwargs)
 
+            for param in self.columns:
+                if param not in self.limits:
+                    self.limits[param] = (self[param].min(), self[param].max())
```
lukashergt, line 94: Fairly simple implementation. Too simple? (Am going to write tests, now)

I first was wondering whether these lines should go into `ChainReader` or `GetDistReader`, but ultimately decided that this doesn't actually belong in the readers. The readers _read_ (duh!) from the files, whereas this here is a post reading process. 

Am open to suggestions for a better placement in the project structure, though.

williamjameshandley:
This looks great. I've had a quick look and it seems like it works well. Could you iterate the version number in the README to 1.2.6 to signify a change, and I'll formalise this as a new release.

anesthetic/samples.py
williamjameshandley: This does indeed seem too easy, but as far as I can tell does seem to do what we want.

lukashergt:
I noticed one thing:
In the current setup, it doesn't assign limits to `logL` or `weight`. Would we prefer to have limits set on them, too? That might be better for cases where a PC run is ended prematurely and `logL` will have a sharp cutoff which could get washed out by the KDE without limits.

williamjameshandley:
I think it makes sense to automate this too. It would be harder to communicate in the case that you don't want a hard limit, but you could always do that manually at the python level if you had to. You should probably also do this for the nlive column as well.

anesthetic/samples.py
```diff
@@ -429,6 +433,10 @@ def __init__(self, *args, **kwargs):
             if logL_birth is not None:
                 self._compute_nlive(logL_birth)
 
+            for param in self.columns:
+                if param not in self.limits:
+                    self.limits[param] = (self[param].min(), self[param].max())
```
lukashergt, line 438: Exact copy as for MCMCSamples does the trick to get limits for `logL`, `weight` and `nlive`.

It is a repetition of the same lines, though. Would you prefer if this was in its own utils function?

lukashergt: Nevermind, moved into its own method. Much better this way.

anesthetic/samples.py
```diff
@@ -97,6 +97,14 @@ def __init__(self, *args, **kwargs):
                 self['weight'] = self.weight
                 self.tex['weight'] = r'MCMC weight'
 
+            self.set_automatic_limits()
+
+    def set_automatic_limits(self):
+        """Set all unassigned limits to min and max of sample."""
```
lukashergt, line 103: Happy with name and docstring? Open for suggestions of improvement.

williamjameshandley: Perhaps make it _set_automatic_limits (i.e. with an extra leading underscore), as it's not something that a user is expected to call by themselves, but could do if they wanted to.

PR #85, opened 2020-05-30
-infinity handling in set_beta

### LINKED ISSUE #84: `.set_beta(0)` assigns `weight=NaN` to parameters with zero weight causing plot errors
**Describe the bug**
When the initial data contains samples with zero weight, these will turn into NaNs when running `.set_beta(0)`. This will then result in errors when trying to plot the prior.

**MWE**

```python
ns = NestedSamples(root="/path/to/anesthetic/tests/example_data/pc")
for i in range(10):
    ns['logL'][i] = -np.inf
    ns['weight'][i] = 0
prior = ns.set_beta(0)
prior.plot_1d(['x0', 'x1'])
```

**Error output**

```python
ValueError                                Traceback (most recent call last)
<ipython-input-24-31aeaee3a714> in <module>
----> 1 prior.plot_1d(['x0', 'x1'])

~/.virtualenvs/py38env/lib/python3.8/site-packages/anesthetic/samples.py in plot_1d(self, axes, *args, **kwargs)
    256 
    257         for x, ax in axes.iteritems():
--> 258             self.plot(ax, x, *args, **kwargs)
    259 
    260         return fig, axes

~/.virtualenvs/py38env/lib/python3.8/site-packages/anesthetic/samples.py in plot(self, ax, paramname_x, paramname_y, *args, **kwargs)
    157                     if ncompress is None:
    158                         ncompress = 1000
--> 159                     return kde_plot_1d(ax, self[paramname_x],
    160                                        weights=self.weight,
    161                                        ncompress=ncompress,

~/.virtualenvs/py38env/lib/python3.8/site-packages/anesthetic/plot.py in kde_plot_1d(ax, data, *args, **kwargs)
    322     weights = kwargs.pop('weights', None)
    323     ncompress = kwargs.pop('ncompress', 1000)
--> 324     x, w = sample_compression_1d(data, weights, ncompress)
    325     kde = gaussian_kde(x, weights=w)
    326     p = kde(x)

~/.virtualenvs/py38env/lib/python3.8/site-packages/anesthetic/utils.py in sample_compression_1d(x, w, n)
    344         i = w.index
    345     else:
--> 346         i = numpy.random.choice(w.index, size=n, replace=False, p=w/w.sum())
    347 
    348     # Define sub-samples

mtrand.pyx in numpy.random.mtrand.RandomState.choice()

ValueError: probabilities contain NaN
```

**Expected behavior**

I'd say that `.set_beta` should carry over the zero weight instead of assigning NaNs. Then the `w/w.sum()` in the sample compression wouldn't produce the plotting error. Thoughts?

**Additional context**

Cobaya allows to specify external priors with which you can exclude parts of a multi-parameter space. This will create entries with `log(0)` likelihood and with `0` weight

lukashergt:
Implements minimal change suggested in #84 .

Fixes #84 

# Checklist:

### DIFF
--- a/anesthetic/samples.py
+++ b/anesthetic/samples.py
@@ -447,7 +447,8 @@ class NestedSamples(MCMCSamples):
     @beta.setter
     def beta(self, beta):
         self._beta = beta
-        logw = self.dlogX() + self.beta*self.logL
+        logw = self.dlogX() + numpy.where(self.logL == -numpy.inf,
+                                          -numpy.inf, self.beta * self.logL)
         self._weight = numpy.exp(logw - logw.max())
 
         if self._weight is not None:
--- a/tests/test_samples.py
+++ b/tests/test_samples.py
@@ -508,6 +508,16 @@ def test_beta():
         assert not numpy.array_equal(pc['weight'], weight)
 
 
+def test_beta_with_logL_infinities():
+    ns = NestedSamples(root="./tests/example_data/pc")
+    for i in range(10):
+        ns['logL'][i] = -numpy.inf
+    prior = ns.set_beta(0)
+    assert numpy.all(prior.logL[:10] == -numpy.inf)
+    assert numpy.all(prior.weight[:10] == 0)
+    ns.plot_1d(['x0', 'x1'])
+
+
 def test_live_points():
     numpy.random.seed(4)
     pc = NestedSamples(root="./tests/example_data/pc")


tests/test_samples.py
```diff
@@ -508,6 +508,16 @@ def test_beta():
         assert not numpy.array_equal(pc['weight'], weight)
 
 
+def test_beta_with_logL_infinities():
+    ns = NestedSamples(root="./tests/example_data/pc")
+    for i in range(10):
+        ns['logL'][i] = -numpy.inf
+    prior = ns.set_beta(0)
+    assert numpy.all(prior.logL[:10] == -numpy.inf)
+    assert numpy.all(prior.weight[:10] == 0)
```
williamjameshandley, line 517: Do you need to slice to 10 for the weights? shouldn't all the weights be zero if beta is zero?

lukashergt: No, should they be? All weights zero? That would be like having no samples, wouldn't it?

lukashergt:
As mentioned in #84 this is a cobaya related issue, where an external prior that serves to exclude part of the parameter space will result in -inf entries for the logL column.

If we were to say that -inf in logL should never be allowed, then this would need to be checked for and enforced when loading the samples.

lukashergt:
Note that the modified line calculates `logw` not `w`, which is why np.where needs to return -inf instead of 0

anesthetic/samples.py
```diff
@@ -447,7 +447,8 @@ def beta(self):
     @beta.setter
     def beta(self, beta):
         self._beta = beta
-        logw = self.dlogX() + self.beta*self.logL
+        logw = self.dlogX() + numpy.where(self.logL == -numpy.inf,
+                                          -numpy.inf, self.beta * self.logL)
         self._weight = numpy.exp(logw - logw.max())
```
lukashergt, line 452: Note that the modified line calculates `logw` not `w`, which is why `np.where` needs to return `-np.inf` instead of `0`.

PR #87, opened 2020-05-31
Convert numpy to np

### LINKED ISSUE #76: convert `numpy` to `np`
`np` is the widely used abbreviation for the `numpy` package. So much so that most people (not only me...) will blindly type `np` without looking up whether the abbreviation is used.

Using the abbreviation comes with the benefit of shortening the code, making it easier to keep things on one line and hence more readable.

Maybe worth making the conversion sometime. Usage stays unaffected.

williamjameshandley:
# Description

Following #76, this replaces all instances of `numpy` with `np`.

This obviously shortens some lines of code, meaning that they no longer needed to be broken over lines. I took this opportunity to do so where possible whilst remaining PEP8 compliant.

I made one minor change to a test which previously used `pdarray[column][index]`. It is more future-proof to use `pdarray.loc[index, column]`, since this will give a modifiable view onto an array in future versions of pandas, whilst the other will not (and was currently raising a warning)

Fixes #76

# Checklist:

### DIFF
--- a/anesthetic/boundary.py
+++ b/anesthetic/boundary.py
@@ -1,6 +1,6 @@
 """Boundary correction utilities."""
 
-import numpy
+import numpy as np
 from scipy.special import erf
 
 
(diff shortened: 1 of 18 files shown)


williamjameshandley:
@appetrosyan and @igomezv  You should watch out for this in your PRs if you're currently midway through.  The merge should be seamless/easy to resolve conflicts, but wanted to flag so there aren't any nasty surprises.

Issue #79, opened 2020-04-30
matplotlib uses `ms` as alias for `markersize`


lukashergt:
**Describe the bug**

`ms` is an alias for `markersize` in matplotlib. Would be nice to allow passing either one in anesthetic as well. At the moment it throws a warning when trying to pass `ms` and keeps `markersize` (the one that is `kwargs.pop`-ed internally) instead:
```
MatplotlibDeprecationWarning: Saw kwargs ['ms', 'markersize'] which are all aliases 
for 'markersize'. Kept value from 'markersize'.  Passing multiple aliases for the 
same property will raise a TypeError in 3.3.
```

**To Reproduce**

```python
fig, axes = make_2d_axes(['a', 'b'])
samples.plot_2d(axes, upper_kwargs={'ms': 5});
```

**Where in source code**

`anesthetic.plot.scatter_plot_2d` line 680.

PR #97, opened 2020-06-08
make plotting kwargs user accessible

lukashergt:
There are some plotting kwargs that are currently not user accessible such as the kwarg `range` for 1d histograms. That kwarg is important when faced with a big dynamic range in values such as for `logL` where only a small range is typically relevant for plotting though, which the user should be able to restrict.

More generally, I think we should try to keep as many kwargs passed to `plt.plot`, `plt.hist`, `plt.contour` etc. user accessible as possible. For most kwargs that is already the case, but a few have slipped us by.

# Checklist:

### DIFF
--- a/anesthetic/plot.py
+++ b/anesthetic/plot.py
@@ -460,7 +460,7 @@ def fastkde_contour_plot_2d(ax, data_x, data_y, *args, **kwargs):
         raise ImportError("You need to install fastkde to use fastkde")
 
     levels = iso_probability_contours(pdf, contours=levels)
-    cmap = basic_cmap(color)
+    cmap = kwargs.pop('cmap', basic_cmap(color))
 
     i = (pdf >= levels[0]*0.5).any(axis=0)
     j = (pdf >= levels[0]*0.5).any(axis=1)
@@ -549,7 +549,7 @@ def kde_contour_plot_2d(ax, data_x, data_y, *args, **kwargs):
 
     contours = iso_probability_contours_from_samples(p, weights=w)
 
-    cmap = basic_cmap(color)
+    cmap = kwargs.pop('cmap', basic_cmap(color))
 
     cbar = ax.tricontourf(tri, p, contours, cmap=cmap,
                           zorder=zorder, vmin=0, vmax=p.max(), *args, **kwargs)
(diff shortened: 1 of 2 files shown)


williamjameshandley:
I guess it does make sense to change any other hard-coded choices in this file at the same time. I think there is another cmap statement in `kde_contour_plot_2d`

anesthetic/plot.py
```diff
@@ -677,6 +680,7 @@ def scatter_plot_2d(ax, data_x, data_y, *args, **kwargs):
     xmax = kwargs.pop('xmax', None)
     ymin = kwargs.pop('ymin', None)
     ymax = kwargs.pop('ymax', None)
+    kwargs = cbook.normalize_kwargs(kwargs, mlines.Line2D)
```
lukashergt, line 683: This is the way `matplotlib` handles aliases such as `ms` and `markersize`, so this fixes #79.

As long as we `pop` all needed anesthetic specific kwargs before applying `cbook.normalize_kwargs`, this shouldn't cause any conflicts.

Might be worth applying before all the major matplotlib calls such as `plt.plot`, `plt.contour`, `plt.hist` etc. What do you think, @williamjameshandley?

williamjameshandley: This sounds like an excellent idea. You could either have a go at doing it as part of this PR, or raise it as a new issue. I will approve and leave it up to you whether to merge or continue with `normalise_kwargs`

lukashergt: > Might be worth applying before all the major matplotlib calls such as `plt.plot`, `plt.contour`, `plt.hist` etc.

Actually, `markersize` was the only one so far, where this applies.

lukashergt: Also, `normalize_kwargs` is not as straight forward for contourplots or histograms. 

For `plt.plot` you need to pass `matplotlib.lines.Line2D` as a dictionary for all possible kwargs. This case was easy, as this is done at the start of [`plt.plot`](https://matplotlib.org/3.2.1/_modules/matplotlib/axes/_axes.html#Axes.plot) itself.

I couldn't find what we'd need to use in place of `Line2D` to make this work for `plt.contour`.

lukashergt:
Fixes #79

PR #98, opened 2020-06-16
Read cobaya mcmc

### LINKED ISSUE #18: getdist samples have no burn-in removal
Currently when loading **getdist samples** there is **no burn-in removed**. If there is only one chain file a burn-in removal can be easily performed _after_ the samples are loaded by `MCMCSamples` by discarding a fraction of the first steps. However, if there are multiple chain files that are combined that does not work anymore.

If anesthetic labelled which chain files the steps have come from, then removing a fraction of the steps would be an option again, but it might be nicer to have a keyword argument passed to `MCMCSamples` and `GetDistReader` to take care of this.

lukashergt:
* Provides functionality to read MCMC chains from Cobaya which use `.` instead of `_` for their chains and which have a `.yaml` file instead of a `.paramnames` file.
* Provides functionality to discard burn-in samples, which is useful e.g. when comparing PolyChord results to Planck MCMC chains.

Fixes #18 

# Checklist:

### DIFF
--- a/anesthetic/samples.py
+++ b/anesthetic/samples.py
@@ -59,8 +59,9 @@ class MCMCSamples(WeightedDataFrame):
         default: -1e30
 
     discard_burnin: int or float
-        discard the first integer number of nsamples (int)
+        Discard the first integer number of nsamples (int)
         or the first fraction of nsamples (float).
+        Only works if `root` provided and if chains are GetDist compatible.
         default: False
 
     """
@@ -75,7 +76,8 @@ class MCMCSamples(WeightedDataFrame):
                                  "instead which has the same features as "
                                  "MCMCSamples and more. MCMCSamples should be "
                                  "used for MCMC chains only." % root)
-            w, logL, samples = reader.samples()
+            discard_burnin = kwargs.pop('discard_burnin', False)
+            w, logL, samples = reader.samples(discard_burnin=discard_burnin)
             params, tex = reader.paramnames()
             columns = kwargs.pop('columns', params)
             limits = reader.limits()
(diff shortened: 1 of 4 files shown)


anesthetic/samples.py
```diff
@@ -58,6 +58,12 @@ class MCMCSamples(WeightedDataFrame):
         Anything equal or below this value is set to `-np.inf`.
         default: -1e30
 
+    discard_burnin: int or float
```
lukashergt, line 61: Any strong feelings about the naming of the variable? 
Happy to change if something else would be more appropriate.

lukashergt: Same goes more generally for the whole docstring.

anesthetic/samples.py
```diff
@@ -70,7 +76,8 @@ def __init__(self, *args, **kwargs):
                                  "instead which has the same features as "
                                  "MCMCSamples and more. MCMCSamples should be "
                                  "used for MCMC chains only." % root)
-            w, logL, samples = reader.samples()
+            discard_burnin = kwargs.pop('discard_burnin', False)
+            w, logL, samples = reader.samples(discard_burnin=discard_burnin)
```
williamjameshandley, line 80: Is there any way one could accidentally pass discard_burnin from a NestedSamples object? Given that this is something that people can do if they are not used to working with nested sampling chains, if it is possible we should probably put in a guard against this that throws a ValueError if discard_burnin is passed to NestedSamples.

anesthetic/samples.py
lukashergt: changed to `burn_in`

anesthetic/samples.py
lukashergt: A few lines above this, there already is a guard against using `MCMCSamples` on nested sampling data, so no danger on that front.

When passing to `NestedSamples` it won't reach where `discard_burnin`/`burn_in` is beeing pop-ed, as that happens in the `if root is not None` block of `MCMCSamples` which doesn't get executed by `NestedSamples`.

williamjameshandley:
Many thanks @lukashergt. Please squash and merge.

Issue #95, opened 2020-06-08
incorrect weights in scatter plots when slicing or masking


lukashergt:
When plotting a scatter plot of a masked or sliced `NestedSamples` instance, the weights seem to be incorrect. Could this be a normalisation issue (if so, worth to keep in mind for #90)? 

The data used in the MWE contains some `-inf` entries in `logL`, which are excluded from any plotting. Hence, masking all these `-inf` entries should have no effect on a resulting plot. This seems to be correct for `'kde'` plots, but wrong for `'scatter'` plots. In the MWE below, blue refers to the masked data and orange to the full data, which are not the same in the scatter plots.

This issue becomes very clear when looking at how slicing and masking affects the axes limit behaviour. Compare for example the axis range for `r` in the MWE here to the MWE in #94.

### MWE

data: [unilog_r.log](https://github.com/williamjameshandley/anesthetic/files/4746501/unilog_r.log) (same data as in #93 and #94)

```python
data = pandas.read_csv("/home/hergtl/data/AnestheticData/unilog_r.log", sep='\t')
ns = NestedSamples(data=data[['n_s', 'r', 'logr']], logL=data.logL, logL_birth=data.logL_birth, w=data.weight)
mask = np.isfinite(ns.logL)
fig = plt.figure(figsize=(5, 5))
fig, axes = make_2d_axes(['n_s', 'r', 'logr'], fig=fig)
ns[mask].plot_2d(axes, label='masked')
ns.plot_2d(axes, label='full')
axes.iloc[-1, 0].legend(bbox_to_anchor=(len(axes), len(axes)), loc='upper left');
```

To look at slicing instead, simply change `[mask]` to e.g. `[412:]`:
```python
ns[412:].plot_2d(axes, label='sliced')
```

![unilog_full_vs_masked](https://user-images.githubusercontent.com/18258042/84041286-8a16da80-a99b-11ea-9d15-c995e9154235.png)

**Additional context**

Note that this issue goes beyond `-inf` entries in `logL` but applies more generally to masking and slicing.

lukashergt:
### Simpler MWE:

```python
ns = NestedSamples(root="anesthetic/tests/example_data/pc")
fig, axes = make_2d_axes(['x0', 'x1'])
ns.plot_2d(axes)
ns[1:].plot_2d(axes);
```

![full_vs_sliced](https://user-images.githubusercontent.com/18258042/84073700-33c09080-a9c9-11ea-8210-62844510e519.png)

[...]

Issue #94, opened 2020-06-08
ragged 2d kde boundaries


lukashergt:
The contours of 2d `'kde'` plots are ragged when they reach up to the boudary as can be seen in the `[logr][n_s]` plot in the MWE below. It would be nicer (and more correct strictly speaking) for them to reach up to the boundary.

If I remember correctly this was introduced with #73, which aimed at handling highly correlated samples (such as in the `[logr][r]` plot below) better. #73 already states that it's only a partial fix and should be improved upon in the future. Hence, I thought it might be good to keep this issue in mind in future improvement attempts (e.g. when working on #69).

### MWE:

data: [unilog_r.log](https://github.com/williamjameshandley/anesthetic/files/4746305/unilog_r.log) (same data as in #93)

```python
data = pandas.read_csv("unilog_r.log", sep='\t')
ns = NestedSamples(data=data[['n_s', 'r', 'logr']], logL=data.logL, logL_birth=data.logL_birth, w=data.weight)
fig = plt.figure(figsize=(10, 10))
fig, axes = make_2d_axes(['n_s', 'r', 'logr'], fig=fig)
ns.plot_2d(axes);
```

![unilog_triangle](https://user-images.githubusercontent.com/18258042/84037600-b9771880-a996-11ea-81ad-81681491ffec.png)

lukashergt:
For anybody with a particularly ugly case, one way to improve on this issue is by cranking up the value for `ncompress` which by default is set to 1000. In the OP example the last line should hence be changed, e.g. to:
```python
ns.plot_2d(axes, ncompress=5000);
```

williamjameshandley:
For reference, could you provide an example in this discussion? 

I would leave the issue open, since hopefully future work on alternative kde schemes (e.g. #131) may resolve this.

williamjameshandley:
This is partially resolved by cranking up ncompress and decreasing the bandwidth #193. However, this could also likely be resolved by adjusting the grid of points to include a set of additional _plotting points_ (rather than kde points) on the boundary.

lukashergt:
Fixed by #216.

PR #99, opened 2020-06-16
Improvements for 1D kde plotting (quantiles and compression)

### LINKED ISSUE #93: 1d 'kde' plot cutoff too much
1d `'kde'` plots often seem cut off earlier than the data warrants. In the below example (data file provided, see MWE for how to load) the data range for `logr` actually goes up to the red line but the blue `'kde'` plot stops much earlier. It would be nice to be able to see how the plot drops to zero there.

This might be related to a change in #73 which was introduced to better deal with highly correlated 2d plots (see #68). However, I think it might be better to apply that functionality to 2d `'kde'` plots only, not 1d. A good starting point might be to try and relax the function `cut_and_normalise_gaussian` in `boundary.py` for the 1d case.

### MWE:

data: [unilog_r.log](https://github.com/williamjameshandley/anesthetic/files/4746148/unilog_r.log) (same as in #94)

```python
data = pandas.read_csv("unilog_r.log", sep='\t')
ns = NestedSamples(data=data[['n_s', 'r', 'logr']], logL=data.logL, logL_birth=data.logL_birth, w=data.weight)
fig, axes = make_1d_axes(['logr'], fig=fig)
ns.plot_1d(axes)
axes['logr'].set_xlim(-5, 0)
axes['logr'].set_ylim(0, 1.1)
axes['logr'].hlines(1, -5, 0)
axes['logr'].axvline(ns.logr.max(), lw=5, c='r')
print("max(logr) = %g" % ns.logr.max())
```

unilog_logr

lukashergt:
This addresses #93 that shows how 1d kde contours are sometimes cut off earlier than desirable.

This introduces an additional kwarg `ncompress_1d` that overwrites `ncompress` for 1d kde plots and by setting `ncompress_1d=0` allows to prevent any compression of the data for 1d kde plots. This can come with some heavy computation time (~o(1min) as opposed to ~o(1s)). Hence, I keep the default at `ncompress_1d=1000`, but it is nice to have the option of investing more computation time to get nicer plots.

The plot below shows a comparison between `ncompress_1d=1000` (the default) and `ncompress_1d=0` (no compression). 

no_1d_kde_compression

Fixes #93 

# Checklist:

### DIFF
--- a/anesthetic/plot.py
+++ b/anesthetic/plot.py
@@ -804,6 +804,7 @@ else:
 
 
 def quantile_plot_interval(q):
+    """Interpret quantile q input to quantile plot range tuple."""
     if isinstance(q, str):
         sigmas = {'1sigma': 0.682689492137086,
                   '2sigma': 0.954499736103642,
--- a/anesthetic/plot.py
+++ b/anesthetic/plot.py
@@ -529,6 +529,11 @@ def kde_contour_plot_2d(ax, data_x, data_y, *args, **kwargs):
     if len(data_x) == 0 or len(data_y) == 0:
         return np.zeros(0), np.zeros(0), np.zeros((0, 0))
 
+    if weights is not None:
+        data_x = data_x[weights != 0]
+        data_y = data_y[weights != 0]
+        weights = weights[weights != 0]
+
     cov = np.cov(data_x, data_y, aweights=weights)
     tri, w = triangular_sample_compression_2d(data_x, data_y, cov,
                                               weights, ncompress)
--- a/anesthetic/plot.py
+++ b/anesthetic/plot.py
@@ -323,6 +323,11 @@ def kde_plot_1d(ax, data, *args, **kwargs):
     xmax = kwargs.pop('xmax', None)
     weights = kwargs.pop('weights', None)
     ncompress = kwargs.pop('ncompress', 1000)
+
+    if weights is not None:
+        data = data[weights != 0]
+        weights = weights[weights != 0]
+
     x, w = sample_compression_1d(data, weights, ncompress)
     kde = gaussian_kde(x, weights=w)
     p = kde(x)
--- a/anesthetic/plot.py
+++ b/anesthetic/plot.py
@@ -268,13 +268,15 @@ def fastkde_plot_1d(ax, data, *args, **kwargs):
 
     xmin = kwargs.pop('xmin', None)
     xmax = kwargs.pop('xmax', None)
+    q = kwargs.pop('q', '5sigma')
+    q = quantile_plot_interval(q=q)
 
     try:
         x, p = fastkde_1d(data, xmin, xmax)
     except NameError:
         raise ImportError("You need to install fastkde to use fastkde")
     p /= p.max()
-    i = (x < quantile(x, 0.99, p)) & (x > quantile(x, 0.01, p)) | (p > 0.1)
+    i = ((x > quantile(x, q[0], p)) & (x < quantile(x, q[1], p)))
 
     ans = ax.plot(x[i], p[i], *args, **kwargs)
     ax.set_xlim(*check_bounds(x[i], xmin, xmax), auto=True)
@@ -324,17 +326,7 @@ def kde_plot_1d(ax, data, *args, **kwargs):
     weights = kwargs.pop('weights', None)
     ncompress = kwargs.pop('ncompress', 1000)
     q = kwargs.pop('q', '5sigma')
-    if isinstance(q, str):
-        sigmas = {'1sigma': 0.682689492137086,
-                  '2sigma': 0.954499736103642,
-                  '3sigma': 0.997300203936740,
-                  '4sigma': 0.999936657516334,
-                  '5sigma': 0.999999426696856}
-        q = (1 - sigmas[q]) / 2
-    if isinstance(q, float) or isinstance(q, int):
-        if q > 0.5:
-            q = 1 - q
-        q = (q, 1-q)
+    q = quantile_plot_interval(q=q)
 
     if weights is not None:
         data = data[weights != 0]
@@ -809,3 +801,18 @@ else:
                                                       emit=emit, auto=auto,
                                                       **kwargs)
         ax.__class__ = DiagonalAxes
+
+
+def quantile_plot_interval(q):
+    if isinstance(q, str):
+        sigmas = {'1sigma': 0.682689492137086,
+                  '2sigma': 0.954499736103642,
+                  '3sigma': 0.997300203936740,
+                  '4sigma': 0.999936657516334,
+                  '5sigma': 0.999999426696856}
+        q = (1 - sigmas[q]) / 2
+    if isinstance(q, float) or isinstance(q, int):
+        if q > 0.5:
+            q = 1 - q
+        q = (q, 1-q)
+    return q
--- a/anesthetic/plot.py
+++ b/anesthetic/plot.py
@@ -327,7 +327,7 @@ def kde_plot_1d(ax, data, *args, **kwargs):
     kde = gaussian_kde(x, weights=w)
     p = kde(x)
     p /= p.max()
-    i = ((x < quantile(x, 0.999, w)) & (x > quantile(x, 0.001, w))) | (p > 0.1)
+    i = (p > 0)
     if xmin is not None:
         i = i & (x > xmin)
     if xmax is not None:
--- a/anesthetic/plot.py
+++ b/anesthetic/plot.py
@@ -323,19 +323,18 @@ def kde_plot_1d(ax, data, *args, **kwargs):
     xmax = kwargs.pop('xmax', None)
     weights = kwargs.pop('weights', None)
     ncompress = kwargs.pop('ncompress', 1000)
-    percentile = kwargs.pop('percentile', '5sigma')
-    if isinstance(percentile, str):
+    q = kwargs.pop('q', '5sigma')
+    if isinstance(q, str):
         sigmas = {'1sigma': 0.682689492137086,
                   '2sigma': 0.954499736103642,
                   '3sigma': 0.997300203936740,
                   '4sigma': 0.999936657516334,
                   '5sigma': 0.999999426696856}
-        percentile = sigmas[percentile]
-    if isinstance(percentile, float) or isinstance((percentile, int)):
-        if percentile < 0.5:
-            percentile = 1 - percentile
-        percentile = (1-percentile, percentile)
-        print(percentile)
+        q = (1 - sigmas[q]) / 2
+    if isinstance(q, float) or isinstance(q, int):
+        if q > 0.5:
+            q = 1 - q
+        q = (q, 1-q)
 
     if weights is not None:
         data = data[weights != 0]
@@ -345,8 +344,7 @@ def kde_plot_1d(ax, data, *args, **kwargs):
     kde = gaussian_kde(x, weights=w)
     p = kde(x)
     p /= p.max()
-    i = ((x > quantile(x, percentile[0], w)) &
-         (x < quantile(x, percentile[1], w))) | (p > 0.1)
+    i = ((x > quantile(x, q[0], w)) & (x < quantile(x, q[1], w)))
     if xmin is not None:
         i = i & (x > xmin)
     if xmax is not None:
--- a/anesthetic/plot.py
+++ b/anesthetic/plot.py
@@ -345,7 +345,8 @@ def kde_plot_1d(ax, data, *args, **kwargs):
     kde = gaussian_kde(x, weights=w)
     p = kde(x)
     p /= p.max()
-    i = ((x > quantile(x, percentile[0], w)) & (x < quantile(x, percentile[1], w))) | (p > 0.1)
+    i = ((x > quantile(x, percentile[0], w)) &
+         (x < quantile(x, percentile[1], w))) | (p > 0.1)
     if xmin is not None:
         i = i & (x > xmin)
     if xmax is not None:
--- a/anesthetic/plot.py
+++ b/anesthetic/plot.py
@@ -323,6 +323,19 @@ def kde_plot_1d(ax, data, *args, **kwargs):
     xmax = kwargs.pop('xmax', None)
     weights = kwargs.pop('weights', None)
     ncompress = kwargs.pop('ncompress', 1000)
+    percentile = kwargs.pop('percentile', '5sigma')
+    if isinstance(percentile, str):
+        sigmas = {'1sigma': 0.682689492137086,
+                  '2sigma': 0.954499736103642,
+                  '3sigma': 0.997300203936740,
+                  '4sigma': 0.999936657516334,
+                  '5sigma': 0.999999426696856}
+        percentile = sigmas[percentile]
+    if isinstance(percentile, float) or isinstance((percentile, int)):
+        if percentile < 0.5:
+            percentile = 1 - percentile
+        percentile = (1-percentile, percentile)
+        print(percentile)
 
     if weights is not None:
         data = data[weights != 0]
@@ -332,7 +345,7 @@ def kde_plot_1d(ax, data, *args, **kwargs):
     kde = gaussian_kde(x, weights=w)
     p = kde(x)
     p /= p.max()
-    i = (p > 0)
+    i = ((x > quantile(x, percentile[0], w)) & (x < quantile(x, percentile[1], w))) | (p > 0.1)
     if xmin is not None:
         i = i & (x > xmin)
     if xmax is not None:
--- a/anesthetic/samples.py
+++ b/anesthetic/samples.py
@@ -152,16 +152,17 @@ class MCMCSamples(WeightedDataFrame):
         do_1d_plot = paramname_y is None or paramname_x == paramname_y
         kwargs['label'] = kwargs.get('label', self.label)
         ncompress = kwargs.pop('ncompress', None)
+        ncompress_1d = kwargs.pop('ncompress_1d', ncompress)
 
         if do_1d_plot:
             if paramname_x in self and plot_type is not None:
                 xmin, xmax = self._limits(paramname_x)
                 if plot_type == 'kde':
-                    if ncompress is None:
-                        ncompress = 1000
+                    if ncompress_1d is None:
+                        ncompress_1d = 1000
                     return kde_plot_1d(ax, self[paramname_x],
                                        weights=self.weight,
-                                       ncompress=ncompress,
+                                       ncompress=ncompress_1d,
                                        xmin=xmin, xmax=xmax,
                                        *args, **kwargs)
                 elif plot_type == 'fastkde':
--- a/anesthetic/samples.py
+++ b/anesthetic/samples.py
@@ -152,17 +152,16 @@ class MCMCSamples(WeightedDataFrame):
         do_1d_plot = paramname_y is None or paramname_x == paramname_y
         kwargs['label'] = kwargs.get('label', self.label)
         ncompress = kwargs.pop('ncompress', None)
-        ncompress_1d = kwargs.pop('ncompress_1d', ncompress)
 
         if do_1d_plot:
             if paramname_x in self and plot_type is not None:
                 xmin, xmax = self._limits(paramname_x)
                 if plot_type == 'kde':
-                    if ncompress_1d is None:
-                        ncompress_1d = 1000
+                    if ncompress is None:
+                        ncompress = 1000
                     return kde_plot_1d(ax, self[paramname_x],
                                        weights=self.weight,
-                                       ncompress=ncompress_1d,
+                                       ncompress=ncompress,
                                        xmin=xmin, xmax=xmax,
                                        *args, **kwargs)
                 elif plot_type == 'fastkde':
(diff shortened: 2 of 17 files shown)


anesthetic/plot.py
```diff
@@ -327,7 +327,7 @@ def kde_plot_1d(ax, data, *args, **kwargs):
     kde = gaussian_kde(x, weights=w)
     p = kde(x)
     p /= p.max()
-    i = ((x < quantile(x, 0.999, w)) & (x > quantile(x, 0.001, w))) | (p > 0.1)
+    i = (p > 0)
```
lukashergt, line 330: Was there an important reason why the data excluded the outer 0.2% ? If yes, I could bring this back in, but without an option of manually adjusting this, this is not ideal...

anesthetic/samples.py
```diff
@@ -152,16 +152,17 @@ def plot(self, ax, paramname_x, paramname_y=None, *args, **kwargs):
         do_1d_plot = paramname_y is None or paramname_x == paramname_y
         kwargs['label'] = kwargs.get('label', self.label)
         ncompress = kwargs.pop('ncompress', None)
+        ncompress_1d = kwargs.pop('ncompress_1d', ncompress)
```
lukashergt, line 155: I'm not sure whether introducing `ncompress_1d` is the best way of implementing this feature. If you have a better idea, please let me know, @williamjameshandley.

Problem is, we need to be able to adjust the 1d kde compression independently from the 2d compression...

anesthetic/plot.py
williamjameshandley: I'm not 100% certain, but I'm pretty sure that this was associated with fastkde, which is now effectively deprecated with all the other changes that have been made since then. Does removing this have any obvious bad effects (either with or without fastkde)?

lukashergt: It does seem to have been copied from fastkde. Removing it causes no obvious problems for kde, so I'd say removing it would be nice as it gets rid of some hard coded numbers that hinder customisation.

anesthetic/samples.py
williamjameshandley: OK, this is actually quite subtle, but I think we can fix this with only very small changes to the code.

The issue is that ncompress is used to downsample the kde (to speed things up) to the points `x`, but that those same points are then the ones that are used to evaluate the kde for plotting. There is no reason that these points need to be the same (and in fact for `kde_contour_plot_2d` they are not). It would be better practice anyway to separate them into `x_compress` and `x_plot`.

In 1D, one option would be to evaluate the kde on `x_plot` a uniform grid from xmin to xmax. This however may create issues for heavily compressed posteriors as 
1. you may not get resolution within the posterior bulk
2. this will mess around with the plotting limits, defaulting them to xmin and xmax

Point 1 can be fixed by using something like
 ```python
x_plot = sorted(np.concatenate([x_compress,np.linspace(xmin,xmax,ncompress]))
```
and this is in keeping with how `kde_contour_plot_2d` does it as well.

Point 2 is a little harder, and may require experimentation, but I imagine that following `kde_contour_plot_2d`'s lead again is the way to go, which is to base the bounds on `x_compress`, rather than `x_plot`

Comment (lukashergt) [reply anesthetic/samples.py:155]: > It would be better practice anyway to separate them into `x_compress` and `x_plot`.

This actually also applies to `xmin` and `xmax`, doesn't it? This could be related to the issue that I raised in #104 , where changing the limits in a way that the posterior bulk is cut messes with the plot.

Comment (williamjameshandley) [reply anesthetic/samples.py:155]: It does. However, provided xmin and xmax are larger than the posterior bulk, it should behave consistently with respect to the rest of the code. I think it would be nice to complete this PR first, and then for all of #104 #95 #93 to be addressed at the same time.

Comment (lukashergt) [reply anesthetic/samples.py]: The envisaged split into `x_compress` and `x_plot` does not work as hoped. It is fine in the 2d case, where we don't try to plot the tails. However, if `x_compress` does not contain the tail region, then it also won't plot anything in that region even if `x_plot` requests it.

The latest commit [bc0b9e7](https://github.com/williamjameshandley/anesthetic/pull/99/commits/bc0b9e73c43e16554b73bf4f5d59852191fda9f7) seems to work, needs a thorough check, whether that is allowed, though.

File: anesthetic/utils.py [hunk 3]
```diff
@@ -343,7 +343,7 @@ def sample_compression_1d(x, w=None, n=1000):
     if sum(w != 0) < n:
         i = w.index
     else:
-        i = np.random.choice(w.index, size=n, replace=False, p=w/w.sum())
+        i = np.random.choice(w.index, size=n, replace=False)
```
lukashergt, line 346: This change seems to be doing exactly what I wanted. By turning this weighted subsample into a uniformly distributed subsample, it actually retains some points from the tails.

From what I can tell, this still works with the weight redistribution in the following lines. However, you are more familiar with this. Is there any danger to discarding the weights in this subsample generation?

lukashergt: This is what this change looks like:
```python
ns = NestedSamples(data=data[['n_s', 'r', 'logr']], logL=data.logL, logL_birth=data.logL_birth, w=data.weight)
fig = plt.figure(figsize=(15, 5))
fig, axes = make_1d_axes(['n_s', 'r', 'logr'], fig=fig, ncols=3)
ns.plot_1d(axes, alpha=0.7, ncompress=1000 , label="ncompress=1000")
ns.plot_1d(axes, alpha=0.7, ncompress=50000, label="no compression")
axes['r'].set_xlim(0, 0.2)
axes['r'].legend()
```
better_1d_compression

lukashergt:
The test from #102 now fails with a ValueError...

```python
    ns.loc[:9, 'x0'] = np.nan
    with pytest.raises((np.linalg.LinAlgError, RuntimeError)):
        ns.plot_2d(['x0', 'x1'])
```

Should we change this to a more general test for `Exception`, indicating that we expect things to fail when there are nans with nonzero weight?

lukashergt:
> Should we change this to a more general test for `Exception`, indicating that we expect things to fail when there are nans with nonzero weight?

This actually required an analogous treatment to the 2d case if there are nans, i.e. zero weights should be discarded first.

anesthetic/utils.py
williamjameshandley: This is actually a very interesting and possibly extremely neat solution, as you are basing the binning on the nested sampling 'scan' rather than the posterior sample locations (and therefore get good tail coverage). The weights are still used and compressed correctly, but the location of the bins is far more sensible.

It's clear that this is better than what is currently being done, but I'm concerned that we're still seeing the compression level still affects some posteriors.

Taking a look at the code in sample_compression_1d, it looks like there is an edge-effect error which may be causing the behaviour we see in the middle plot, and may also be contributing to #93. I have just pushed a fix to this branch, and would appreciate your reproducing this plot on your system to see how things change.

If this does provide the correct behaviour, then there may be a corresponding better way of doing plot_2d.

anesthetic/utils.py
```diff
@@ -335,28 +335,22 @@ def sample_compression_1d(x, w=None, n=1000):
...
-        i = w.index
-    else:
-        i = np.random.choice(w.index, size=n, replace=False, p=w/w.sum())
-
-    # Define sub-samples
-    x_ = np.sort(x[i])
+    x_ = x.copy()
+    if len(x_) > n:
+        x_ = np.random.choice(x_, size=n, replace=False)
+    x_ = np.sort(x_)
```
lukashergt, line 347: Do we need to make a copy of `x` here? Couldn't we just do `np.random.choice(x, ...)`?

williamjameshandley: Yes. We need x_ (which represents the 'bins' that we round to) and x (to round in np.digitize).

lukashergt: I meant to do `x_ = np.random.choice(x, ...)` and only do `x_ = x.copy()` if necessary.

Granted, this might be evil premature optimisation...

williamjameshandley: I think in this case the sort is much more expensive than the copy, so its definitely a case of unecessary optimisation. The new version is an extra line of code, so the better question is which is clearer pythonically?

williamjameshandley: I think yours, so let's go with that.

anesthetic/utils.py
williamjameshandley: More importantly, what is the update to this plot with the new compression scheme?

lukashergt: Here the updated plot:

```python
ns = NestedSamples(data=data[['n_s', 'r', 'logr']], logL=data.logL, logL_birth=data.logL_birth, weight=data.weight)
fig = plt.figure(figsize=(15, 5))
fig, axes = make_1d_axes(['n_s', 'r', 'logr'], fig=fig, ncols=3)
ns.plot_1d(axes, alpha=0.7, ncompress=1000 , label="ncompress=1000")
ns.plot_1d(axes, alpha=0.7, ncompress=5000 , label="ncompress=5000")
ns.plot_1d(axes, alpha=0.7, ncompress=50000, label="no compression")
axes['r'].set_xlim(0, 0.2)
axes['r'].legend()
```

better_1d_compression2

It doesn't change much, but the difference to the full compression was decreased a little.

williamjameshandley: If we were to be scientific, it would be good to have the plot with the new compression scheme, but with posterior rather than nested compression (this amounts to putting a `p=w/w.sum()` in the`np.random.choice`). I imagine the tail behaviour will be worse, but would like to see what the posterior bulk portion looks like. (It may be that the optimum is to have a combination of both).

lukashergt: Here with `p=w/w.sum()` put back in:

better_1d_compression_posteriorcompression

williamjameshandley: OK, so the nested version looks like it's definitely better, and doesn't compromise things a great deal in the tails.

The only remaining question is whether for example in the ns and r case we want to plot that much of the tail (it's clear that we want it in the logr plot), but since the other two have high posterior compression I think the limits are now too wide. The role of the quantile portion in the original code (still in master) was to try and only plot the regions of interest. If you rein-state the code 
```python
i = ((x < quantile(x, 0.999, w)) & (x > quantile(x, 0.001, w))) | (p > 0.1)
```
 how do the plots above change?

williamjameshandley: (but keep the nested binning rather than posterior binning)

williamjameshandley: I would also be interested to know how this update affects #93.

lukashergt: Updated plot using 
```python
i = ((x < quantile(x, 0.999, w)) & (x > quantile(x, 0.001, w))) | (p > 0.1)
```

better_1d_compression_percentiles

lukashergt: I don't like this because with this I have no chance of manually improving things.

On the other hand, it is easy to manually change limits, e.g.:
```python
axes['r'].set_xlim(0, 0.2)
```
as already done above. Note that the `r` plot is extreme because it's from a `logr` sampling run.

lukashergt: > I would also be interested to know how this update affects #93.

Without the percentiles this would fix #93.

williamjameshandley: > Without the percentiles this would fix #93.

￼This is very good.

> I don't like this because with this I have no chance of manually improving things.

I agree that there is definitely a hard-coded element to this, and one might often want to adjust the limits in other cases. 

However, as it stands this does not produce a very satisfactory plot by default, and would require manual adjustment (which would be impractical for large plots).

Is there a way that we could have the default bounds of the plot focusing on the posterior bulk, but still having the line in theory plotting the entire range?

lukashergt: > Is there a way that we could have the default bounds of the plot focusing on the posterior bulk, but still having the line in theory plotting the entire range?

Introduce a `percentile` kwarg?

williamjameshandley: I think the issue is that we currently use xmin and xmax in two ways, to both trim the plot bounds, and to adjust the kde at the edges. This is the source of #105, and possibly #94 and #95.

I imagine that there is a single PR that fixes all of these. The question is whether we make it this one, or find a reasonable half-way house to merge and close this.

lukashergt: I already looked into a `percentile` kwarg, so I pushed it. Happy to revert if you want to leave that to another PR.

Tackling all of #94, #95 and #105 is more than I'd go for in this PR.

anesthetic/plot.py
```diff
@@ -323,11 +323,29 @@ def kde_plot_1d(ax, data, *args, **kwargs):
...
+    if weights is not None:
+        data = data[weights != 0]
+        weights = weights[weights != 0]
+
     x, w = sample_compression_1d(data, weights, ncompress)
     kde = gaussian_kde(x, weights=w)
     p = kde(x)
     p /= p.max()
-    i = ((x < quantile(x, 0.999, w)) & (x > quantile(x, 0.001, w))) | (p > 0.1)
+    i = ((x > quantile(x, percentile[0], w)) & (x < quantile(x, percentile[1], w))) | (p > 0.1)
```
lukashergt, line 348: `percentile` can now be either a `str`, a `float`, or a `tuple of float`.

`str`: specify desired sigma level, gets converted to corresponding percentile
`float`: percentile (symmetric)
`tuple`: left and right percentile (can be asymmetric)

anesthetic/plot.py
```diff
@@ -323,11 +323,29 @@ def kde_plot_1d(ax, data, *args, **kwargs):
...
+                  '2sigma': 0.954499736103642,
+                  '3sigma': 0.997300203936740,
+                  '4sigma': 0.999936657516334,
+                  '5sigma': 0.999999426696856}
+        percentile = sigmas[percentile]
+    if isinstance(percentile, float) or isinstance((percentile, int)):
+        if percentile < 0.5:
+            percentile = 1 - percentile
+        percentile = (1-percentile, percentile)
+        print(percentile)
```
williamjameshandley, line 338: Cool! I would default to 3 sigma, as I think this is closest to the existing behaviour, or perhaps 0.001.

You should remove the print statement.

 `if isinstance(percentile, float) or isinstance((percentile, int)):` might be better using `np.isscalar`.

anesthetic/utils.py
williamjameshandley: Agreed

anesthetic/plot.py
williamjameshandley: Quantile is probably more accurate (as it's a number between 0 and 1).

anesthetic/plot.py
lukashergt: > Cool! I would default to 3 sigma, as I think this is closest to the existing behaviour, or perhaps 0.001.

I think 5 sigma quite consistently gives prettier default results. Here a comparison of 3 sigma (top) and 5 sigma (bottom):

better_1d_3sigma

better_1d_5sigma

The tails nicely hit the floor in the 5 sigma case without extending too far out, but your call.

williamjameshandley:
Many thanks @lukashergt . Please squash and merge.

anesthetic/plot.py
lukashergt: `quantile` is already taken by the function... Change the function name to `get_quantile`?

lukashergt:
> Many thanks @lukashergt . Please squash and merge.

Still tests to write...

anesthetic/plot.py
lukashergt: > Quantile is probably more accurate (as it's a number between 0 and 1).

Actually, aren't both quantile and percentile referring to the x-values and not the p-values? Should I go for `pvalue` or `p_value` or `plot_p_value`  or something like that?  (`plot_p_interval`, `plot_p_range`, ...?)

williamjameshandley: None of these are great. We could just be consistent with [scipy](https://numpy.org/doc/stable/reference/generated/numpy.quantile.html), and call it q, as we are plotting the inner 'q'-quantiles.

anesthetic/samples.py
```diff
@@ -139,6 +139,17 @@ def plot(self, ax, paramname_x, paramname_y=None, *args, **kwargs):
...
+        q: str, float, (float, float)
+            Plot the `q` inner posterior quantiles in 1d 'kde' plots. To plot
+            the full range, set `q=0` or `q=1`.
+            * if str: any of {'1sigma', '2sigma', '3sigma', '4sigma', '5sigma'}
+                Plot within mean +/- #sigma of posterior.
+            * if float: Plot within the symmetric confidence interval
+                `(1-q, q)`  or `(q, 1-q)`.
+            * if tuple:  Plot within the (possibly asymmetric) confidence
+                interval `q`.
+            optional, (Default: '5sigma')
```
lukashergt, line 151: Is this docstring understandable?

anesthetic/plot.py
```diff
@@ -323,11 +323,28 @@ def kde_plot_1d(ax, data, *args, **kwargs):
...
+    if weights is not None:
+        data = data[weights != 0]
+        weights = weights[weights != 0]
+
     x, w = sample_compression_1d(data, weights, ncompress)
     kde = gaussian_kde(x, weights=w)
     p = kde(x)
     p /= p.max()
-    i = ((x < quantile(x, 0.999, w)) & (x > quantile(x, 0.001, w))) | (p > 0.1)
+    i = ((x > quantile(x, q[0], w)) & (x < quantile(x, q[1], w)))
```
lukashergt, line 347: Should the same thing be done for `fastkde`?

anesthetic/samples.py
williamjameshandley: Looks good to me.

anesthetic/plot.py
williamjameshandley: Quantile-wise yes, although the compression does not apply to fastkde.

Are you currently adding tests, or should I review this in toto?

lukashergt: writing tests

tests/test_plot.py
```diff
@@ -187,40 +187,49 @@ def test_2d_axes_limits():
                 assert(axes[z][y].get_ylim() == (c, d))
 
 
-def test_kde_plot_1d():
+@pytest.mark.parametrize('plot_1d', [kde_plot_1d, fastkde_plot_1d])
+def test_kde_plot_1d(plot_1d):
```
lukashergt, line 191: @williamjameshandley, did you know about the `pytest.mark.parametrize` functionality? Learnt about it recently. Super useful. That way there is no need for for-loops and when a test fails, it tells you which item caused the issue.

williamjameshandley: I have seen it used in the past, but never managed to find the time to implement it in anesthetic. I agree that this would make the whole test suite far neater. Do you want to raise an issue/PR that starts this off?

anesthetic/plot.py
```diff
@@ -268,13 +268,15 @@ def fastkde_plot_1d(ax, data, *args, **kwargs):
...
+    q = kwargs.pop('q', '5sigma')
+    q = quantile_plot_interval(q=q)
 
     try:
         x, p = fastkde_1d(data, xmin, xmax)
     except NameError:
         raise ImportError("You need to install fastkde to use fastkde")
     p /= p.max()
-    i = (x < quantile(x, 0.99, p)) & (x > quantile(x, 0.01, p)) | (p > 0.1)
+    i = ((x > quantile(x, q[0], p)) & (x < quantile(x, q[1], p)))
```
williamjameshandley, line 279: You removed the `| (p>0.)` code here. I think the intention of this was to make sure that if the posterior was pressed up against a boundary you didn't clip it nearby. Was there a reason you removed it?

tests/test_plot.py
```diff
@@ -400,3 +409,28 @@ def test_scatter_plot_2d():
...
+    q1, q2 = quantile_plot_interval(q=floats)
+    assert q1 == min(floats, 1 - floats)
+    assert q2 == max(floats, 1 - floats)
+
+
+@pytest.mark.parametrize('q1, q2', [(0, 1), (0.1, 0.9), (0, 0.9), (0.1, 1)])
+def test_quantile_plot_interval_tuple(q1, q2):
+    _q1, _q2 = quantile_plot_interval(q=(q1, q2))
+    assert _q1 == q1
+    assert _q2 == q2
```
williamjameshandley, line 436: These parameterised tests are very neat! It would be great to update the rest of the suite.

tests/test_samples.py
```diff
@@ -600,7 +600,7 @@ def test_contour_plot_2d_nan():
     ns = NestedSamples(root='./tests/example_data/pc')
 
     ns.loc[:9, 'x0'] = np.nan
-    with pytest.raises((np.linalg.LinAlgError, RuntimeError)):
+    with pytest.raises((np.linalg.LinAlgError, RuntimeError, ValueError)):
```
williamjameshandley, line 603: This is preferable to a more general `Exception`, since this way we will catch any changes which raise an exception for the wrong reason.

anesthetic/plot.py
lukashergt: This gave unexpected output when specifying `q`, e.g. even though specifying `q='sigma1'` it would plot up to almost `'sigma3'`.

tests/test_plot.py
lukashergt: Agreed, but that will have to be for a future PR.

anesthetic/plot.py
williamjameshandley: That makes sense.

lukashergt: Do you think the `p>0.1` is really needed?

williamjameshandley: No. Keep an eye on it as you use it, but I think it was an attempt to address the issues which the rest of this PR does

Issue #92, opened 2020-06-08
docstring entry for weight in NestedSamples


lukashergt:
When trying to pass data manually to `NestedSamples` I stumbled over a missing docstring entry for `weight`. It is specified in `MCMCSamples`. It really should be specified for both, though, just like `logL` is specified in both cases.

I wouldn't have stumbled over this had it not been for a naming inconsistency. `MCMCSamples` takes the kwarg `w` for the weights, but subsequently that column is always referred to as `weight`. I think it would be better to rename that kwarg to `weight` for consistency.

PR #103, opened 2020-06-23
Added missing default values to NestedSamples

### LINKED ISSUE #91: defaults in docstrings
If you were to provide data by hand instead of reading from a chain, it would be useful to have the default values in the docstring as is already the case for `logzero`. 

* missing default values in `NestedSamples`:
- [ ] `beta`: Default is `either inferred from root or 1`.
- [ ] `label`: Default is `base name of root` (default value is undefined if root is not specified).
- [ ] `limits`: Default is `either inferred from .ranges file or minimum and maximum of data`.
- [ ] `logL_birth`: Default is `either inferred from root or None`.

williamjameshandley:
# Description

This addresses all the points in #91, aside from `logL_birth`, which like `logL` is overriden by `root`, so there is no 'default' per-se. If root is not supplied, then you should supply a `logL_birth` of some form, although this cannot be an error if not on account of how it extends the underlying `pd.DataFrame` class.

Fixes #91

# Checklist:

### DIFF
--- a/anesthetic/weighted_pandas.py
+++ b/anesthetic/weighted_pandas.py
@@ -45,6 +45,7 @@ class WeightedSeries(_WeightedObject, pandas.Series):
 
     def __init__(self, *args, **kwargs):
         w = kwargs.pop('w', None)
+        w = kwargs.pop('weight', w)
         super(WeightedSeries, self).__init__(*args, **kwargs)
         self._construct_weights(w)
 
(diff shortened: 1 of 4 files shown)


williamjameshandley:
I've also taken this opportunity to fix #92, in a non-breaking way, so that now both `w` and `weight` can be passed as kwargs to `WeightedSeries` and `WeightedDataFrame` objects.

anesthetic/weighted_pandas.py
```diff
@@ -45,6 +45,7 @@ class WeightedSeries(_WeightedObject, pandas.Series):
 
     def __init__(self, *args, **kwargs):
         w = kwargs.pop('w', None)
+        w = kwargs.pop('weight', w)
```
lukashergt, line 48: Do we want to add a `DeprecationWarning` here, such that we can reduce this to one line in the future?

PR #104, opened 2020-06-23
Equal min and max limits fix

### LINKED ISSUE #89: equal min and max limits
Plot types that require the limits attribute such as `'hist'` run into trouble when the minimum and maximum limits are equal. This can happen e.g. for the `logL` parameter when using the `one` likelihood.

Not sure what would be the best way of handling such a special case. 
* Leave it for manual handling?
* Test for `min=max` and add/subtract a fraction in that case? That would require special handling of zero.

williamjameshandley:
# Description

This PR cleans up the xmin and xmax interface by allowing users to override xmin and xmax as user-provided kwargs (which also reduces the length of the code).

Fixes #89 by returning no limits `(None, None)` if equal limits are specified for any reason.

Initial test fails for first commit, which is fixed by the second.

# Checklist:

### DIFF
--- a/tests/test_samples.py
+++ b/tests/test_samples.py
@@ -571,3 +571,24 @@ def test_limit_assignment():
     assert ns.limits['weight'][1] == 1
     assert ns.limits['nlive'][0] == 0
     assert ns.limits['nlive'][1] == 125
+
+
+def test_xmin_xmax_1d():
+    """Test to provide a solution to #89"""
+    np.random.seed(3)
+    ns = NestedSamples(root='./tests/example_data/pc')
+    fig, ax = ns.plot_1d('x0', plot_type='hist')
+    assert ax['x0'].get_xlim() != (-1, 1)
+    fig, ax = ns.plot_1d('x0', plot_type='hist', xmin=-1, xmax=1)
+    assert ax['x0'].get_xlim() == (-1, 1)
+
+
+def test_equal_min_max():
+    """Test to provide a solution to #89"""
+    np.random.seed(3)
+    ns = NestedSamples(root='./tests/example_data/pc')
+    with pytest.raises(ValueError):
+        ns.plot_2d(['x0', 'x1', 'x2'], xmin=3, xmax=3)
+
+    ns.limits['x0'] = (3, 3)
+    ns.plot_2d(['x0', 'x1', 'x2'])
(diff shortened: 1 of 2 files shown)


tests/test_samples.py
```diff
@@ -573,6 +573,27 @@ def test_limit_assignment():
...
+
+def test_equal_min_max():
+    """Test to provide a solution to #89"""
+    np.random.seed(3)
+    ns = NestedSamples(root='./tests/example_data/pc')
+    with pytest.raises(ValueError):
+        ns.plot_2d(['x0', 'x1', 'x2'], xmin=3, xmax=3)
+
+    ns.limits['x0'] = (3, 3)
+    ns.plot_2d(['x0', 'x1', 'x2'])
```
lukashergt, line 594: The code snippet:
```
ns = NestedSamples(root="./tests/example_data/pc")
ns.limits['x0'] = (0, 5)
ns.plot_2d(['x0', 'x1', 'x2']);
```
does not produce the expected output:

equal_min_max

lukashergt: This actually already is an issue in `master`.

williamjameshandley: This is an unexpected side-effect. I don't think I've tried to set xmin and xmax less than the actual ranges before this. What's happening is that `ncompress` is applied to all the samples, and then the xmin and xmax limits are applied to these compressed samples, so there aren't enough points to do a decent contour plot.

This is not an easy fix, as at the moment it will require separate treatment for `'fastkde'`, `'scatter'` and `'kde'`, and ideally these would be treated in the same way. This also likely interacts with #99 and #95. 

Given that this affects master as well, and is likely better addressed in conjunction with the fixes to #99 and #95, could you raise this as a separate issue, or add it to one of the others (with your nice MWE copy-pasted)?

lukashergt: This crossed with a comment from me in #99, where I wonder whether we should have separate `xmin` and `xmax` for compression and plotting.

lukashergt: > Given that this affects master as well, and is likely better addressed in conjunction with the fixes to #99 and #95, could you raise this as a separate issue, or add it to one of the others (with your nice MWE copy-pasted)?

Will do.

williamjameshandley:
>With the current setup, the 1d histogram of e.g. the one-likelihood still isn't plotted. Is that the intention?

I think so, since I'm not sure if there is a correct way to plot this histogram (and at least this is consistent with what matplotlib would do if you asked it). What happens for a kde?

lukashergt:
> What happens for a kde?

kde doesn't plot anything either.

williamjameshandley:
Many thanks @lukashergt

PR #109, opened 2020-07-01
Pop quantile kwarg q in 2d

lukashergt:
The quantile kwarg `q` introduced in #99 causes warnings and errors in the 2d plots for fastkde, kde, and scatter.
This PR removes it (`kwargs.pop('q', None)`) before calling the plotting functions.

# Checklist:


anesthetic/plot.py
```diff
@@ -694,6 +696,7 @@ def scatter_plot_2d(ax, data_x, data_y, *args, **kwargs):
     xmax = kwargs.pop('xmax', None)
     ymin = kwargs.pop('ymin', None)
     ymax = kwargs.pop('ymax', None)
+    _ = kwargs.pop('q', None)
```
williamjameshandley, line 699: You don't need to assign to `_`. I think it's probably clearer to just have `kwargs.pop('q', None)` on its own, and it should probably be at the end of the block of kwarg processing where possible.

lukashergt: Done.

PR #102, opened 2020-06-23
Nan - zero weight bug fix for contour plot covariance computation

williamjameshandley:
# Description

This PR starts a fix for #96, by pushing a failing test, which will be fixed by masking out zero weights in the covariance matrix calculation.

Fixes #96

# Checklist:

- [X] I have performed a self-review of my own code
- [X] My code is PEP8 compliant (`flake8 anesthetic tests`)
- [X] My code contains compliant docstrings (`pydocstyle --convention=numpy anesthetic`)
- [x] New and existing unit tests pass locally with my changes (`python -m pytest`)
- [X] I have added tests that prove my fix is effective or that my feature works

PR #112, opened 2020-07-04
add tests for nan in data with zero weights causing issues in mean, var, std, cov

### LINKED ISSUE #111: `nan` handling of statistical functions such as `mean`, `std` etc.
Currently the statistical functions such as `mean`, `std` etc. return `nan` if there are `nan` in the data, even if the corresponding weights are zero. 

To fix this, the zero weight data should be discarded before calculating the mean etc., similarly to #102, where contour plots were fixed for `nan` in the data.

lukashergt:
Make a change similar to #102 for statistical functions (mean, var, etc.) in `WeightedSeries` and `WeightedDataFrame` to better handle `nan` in the data.

Fixes #111 

# Checklist:

### DIFF
--- a/anesthetic/weighted_pandas.py
+++ b/anesthetic/weighted_pandas.py
@@ -55,11 +55,18 @@ class WeightedSeries(_WeightedObject, pandas.Series):
 
     def mean(self):
         """Weighted mean of the sampled distribution."""
-        return np.average(self, weights=self.weight)
+        if self.weight is None:
+            return np.average(self)
+        nonzero = self.weight > 0
+        return np.average(self[nonzero], weights=self.weight[nonzero])
 
     def var(self):
         """Weighted variance of the sampled distribution."""
-        return np.average((self-self.mean())**2, weights=self.weight)
+        if self.weight is None:
+            return np.average((self-self.mean())**2)
+        nonzero = self.weight > 0
+        return np.average((self[nonzero]-self.mean())**2,
+                          weights=self.weight[nonzero])
 
     def quantile(self, q=0.5):
         """Weighted quantile of the sampled distribution."""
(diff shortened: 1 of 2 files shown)


anesthetic/weighted_pandas.py
```diff
@@ -55,11 +55,18 @@ def __init__(self, *args, **kwargs):
 
     def mean(self):
         """Weighted mean of the sampled distribution."""
-        return np.average(self, weights=self.weight)
+        if self.weight is None:
+            return np.average(self)
+        nonzero = self.weight > 0
+        return np.average(self[nonzero], weights=self.weight[nonzero])
```
lukashergt, line 61: The failing code coverage comes from the `if.self.weight is None` case.

Do we need to check for this? Or are we confident that any `weight=None` is turned into `weight=1` by the weighted objects?

lukashergt:
> The more 'pandas' way to do this would probably be
> 
> ```
> samples.dropna().mean()
> ```
> 
> or
> 
> ```
> samples.mean(skipna=True)
> ```

I am unsure whether we would want to fully ignore `nan`. This way we only discard `nan` if they have zero weight, which is the safer way, I think.

lukashergt:
> At the moment we don't have negative weights, but is there any reason we need to exclude them here? It would make more sense in terms of readability for `nonzero = self.weight !=0`

Are negative weights a thing? But true, we actually only want to discard zero weights. 

I'll change it accordingly.

anesthetic/weighted_pandas.py
lukashergt: Tests run fine upon deleting those lines...

williamjameshandley:
> I am unsure whether we would want to fully ignore `nan`. This way we only discard `nan` if they have zero weight, which is the safer way, I think.

I agree. This is why I think that your solution of dropping zero weight nans, and then leaving it up to the user if they want to drop all nans is the correct thing to do. It's important for future developers (including us) that we record this here as the reasoning for doing so.

anesthetic/weighted_pandas.py
williamjameshandley: I think that this comes from an earlier version of the code when weights were treated slightly differently. `weight` should now never be `None` (if `_weight==None`, then `weight` is returned as a set of appropriate `1.`s). We should drop it.

lukashergt:
> but `skipna` would be a good arg to add in this pull request, so that if you want you can also easily skip non-zero weight nans.

According to [pandas](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.mean.html) the signature of `skipna` is `None` but the default is `True`, is that what we want, too? 

I'm not sure whether skipping `nan` with nonzero weight is indeed what we want...

williamjameshandley:
Hmm. I agree, although it does mean that NestedSamples has different behaviour with repect to pandas. Leave the PR as it is for now.

PR #113, opened 2020-07-21
Bug fix for contour plots

williamjameshandley:
# Description

Eagle-eyed anesthetic users will have noted that sometimes there is a small shaded rectangle in the lower left corner of a contour plot canvas. This is caused by the patches that are appended to contour plots in order to generate a pleasing legend icon. The small rectangle can be removed by adjusting the patch to be zero size. This is a very minor bug fix, so should not require any additional tests.

Note that since the recent release of matplotlib 3.3, a couple of the tests now fail due to a change in the behaviour of gridspec. This PR also addresses this by checking before attempting to construct a zero sized grid of plots.

# Checklist:

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -3,7 +3,7 @@ anesthetic: nested sampling visualisation
 =========================================
 :anesthetic: nested sampling visualisation
 :Author: Will Handley
-:Version: 1.3.3
+:Version: 1.3.4
 :Homepage: https://github.com/williamjameshandley/anesthetic
 :Documentation: http://anesthetic.readthedocs.io/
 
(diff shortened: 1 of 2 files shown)


williamjameshandley:
Note that since the recent release of matplotlib 3.3, a couple of the tests now fail due to a change in the behaviour of gridspec. This PR also addresses this

williamjameshandley:
There are other, less kind, synonyms for "eagle-eyed" that more accurately describe my attention to detail. I spotted it in the big, single page figure in version 1 of this [arXiv paper](https://arxiv.org/abs/2007.08496) where it was quite isolated. I'll correct this  for version 2.

appetrosyan:
As an "eagle eyed" user of Anesthetic, I'd like to point out that the figures contain the same drawings multiple times. I spotted that when I was changing the colour scheme of a .pdf plot (rather than re-running the code that generated it). As anesthetic's performance is the main bottleneck for Paranestamol too, I would suggest having a look.

williamjameshandley:
This does sound inefficient if true -- are you sure that it's multiple drawings, and not multiple matplotlib references to the same object? Are you able to provide a MWE of this so I can take a look?

appetrosyan:
It might actually be a reference that becomes a duplication in PDF export, but to see it, 

**MWE**:
Try plotting the BAO data to PDF. Open in Inkscape and remove objects one by one. 

Looking through the artists, there are far too many to conclusively say that it's not a duplicated reference.

PR #114, opened 2020-07-27
Unfilled contours and inherit color from cmap

lukashergt:
Been wanting the option to create unfilled contours for  a while. Particularly useful when faced with slim contours such as for inflation models in an ns-r-plot, which basically only consist of contour lines and thus all appear black. 

Also, if `cmap` is provided instead of `color`, then `color` is set to the colour of the inner contours for line plots or scatter plots.

# Todo:

- [x] Adapt changes also for `fastkde_plot_2d` once happy with changes for `kde_plot_2d`.

# Checklist:

### DIFF
--- a/anesthetic/plot.py
+++ b/anesthetic/plot.py
@@ -269,6 +269,9 @@ def fastkde_plot_1d(ax, data, *args, **kwargs):
 
     xmin = kwargs.pop('xmin', None)
     xmax = kwargs.pop('xmax', None)
+    cmap = kwargs.pop('cmap', None)
+    color = kwargs.pop('color', (next(ax._get_lines.prop_cycler)['color']
+                                 if cmap is None else cmap(2/3)))
     q = kwargs.pop('q', '5sigma')
     q = quantile_plot_interval(q=q)
 
@@ -279,7 +282,7 @@ def fastkde_plot_1d(ax, data, *args, **kwargs):
     p /= p.max()
     i = ((x > quantile(x, q[0], p)) & (x < quantile(x, q[1], p)))
 
-    ans = ax.plot(x[i], p[i], *args, **kwargs)
+    ans = ax.plot(x[i], p[i], color=color, *args, **kwargs)
     ax.set_xlim(*check_bounds(x[i], xmin, xmax), auto=True)
     return ans
 
@@ -399,16 +402,20 @@ def hist_plot_1d(ax, data, *args, **kwargs):
         xmax = quantile(data, 0.99, weights)
     range = kwargs.pop('range', (xmin, xmax))
     histtype = kwargs.pop('histtype', 'bar')
+    cmap = kwargs.pop('cmap', None)
+    color = kwargs.pop('color', (next(ax._get_lines.prop_cycler)['color']
+                                 if cmap is None else cmap(2/3)))
 
     if plotter == 'astropyhist':
         try:
-            h, edges, bars = hist(data, ax=ax, range=range,
+            h, edges, bars = hist(data, ax=ax, color=color, range=range,
                                   histtype=histtype, *args, **kwargs)
         except NameError:
             raise ImportError("You need to install astropy to use astropyhist")
     else:
-        h, edges, bars = ax.hist(data, range=range, histtype=histtype,
-                                 weights=weights, *args, **kwargs)
+        h, edges, bars = ax.hist(data, color=color, range=range,
+                                 histtype=histtype, weights=weights,
+                                 *args, **kwargs)
 
     if histtype == 'bar':
         for b in bars:
--- a/anesthetic/plot.py
+++ b/anesthetic/plot.py
@@ -35,7 +35,7 @@ from anesthetic.utils import (sample_compression_1d, quantile,
                               triangular_sample_compression_2d,
                               iso_probability_contours,
                               iso_probability_contours_from_samples,
-                              scaled_triangulation)
+                              scaled_triangulation, match_contour_to_contourf)
 from anesthetic.boundary import cut_and_normalise_gaussian
 
 
@@ -504,20 +504,24 @@ def fastkde_contour_plot_2d(ax, data_x, data_y, *args, **kwargs):
                             *args, **kwargs)
         for c in contf.collections:
             c.set_cmap(cmap)
-        ax.patches += [plt.Rectangle((0, 0), 0, 0, fc=cmap(0.999),
-                                     ec=cmap(0.32), lw=2, label=label)]
+        ax.patches += [plt.Rectangle((0, 0), 0, 0, lw=2, label=label,
+                                     fc=cmap(0.999), ec=cmap(0.32))]
         cmap = None
     else:
-        contf = None
-        cmap = kwargs.pop('cmap', None)
-        edgecolor = edgecolor if cmap is None else None
         linewidths = kwargs.pop('linewidths',
                                 plt.rcParams.get('lines.linewidth'))
-        ax.patches += [plt.Rectangle((0, 0), 0, 0, fc=None, ec=edgecolor,
-                                     lw=2, label=label)]
+        cmap = kwargs.pop('cmap', None)
+        contf = None
+        ax.patches += [
+            plt.Rectangle((0, 0), 0, 0, lw=2, label=label,
+                          fc='None' if cmap is None else cmap(0.999),
+                          ec=edgecolor if cmap is None else cmap(0.32))
+        ]
+        edgecolor = edgecolor if cmap is None else None
 
+    vmin, vmax = match_contour_to_contourf(levels, vmin=0, vmax=pdf.max())
     cont = ax.contour(x[i], y[j], pdf[np.ix_(j, i)], levels, zorder=zorder,
-                      vmin=0, vmax=0.32 * pdf.max(), linewidths=linewidths,
+                      vmin=vmin, vmax=vmax, linewidths=linewidths,
                       colors=edgecolor, cmap=cmap, *args, **kwargs)
 
     ax.set_xlim(*check_bounds(x[i], xmin, xmax), auto=True)
@@ -613,19 +617,24 @@ def kde_contour_plot_2d(ax, data_x, data_y, *args, **kwargs):
                                vmin=0, vmax=p.max(), *args, **kwargs)
         for c in contf.collections:
             c.set_cmap(cmap)
-        ax.patches += [plt.Rectangle((0, 0), 0, 0, fc=cmap(0.999),
-                                     ec=cmap(0.32), lw=2, label=label)]
+        ax.patches += [plt.Rectangle((0, 0), 0, 0, lw=2, label=label,
+                                     fc=cmap(0.999), ec=cmap(0.32))]
         cmap = None
     else:
-        contf = None
+        linewidths = kwargs.pop('linewidths',
+                                plt.rcParams.get('lines.linewidth'))
         cmap = kwargs.pop('cmap', None)
+        contf = None
+        ax.patches += [
+            plt.Rectangle((0, 0), 0, 0, lw=2, label=label,
+                          fc='None' if cmap is None else cmap(0.999),
+                          ec=edgecolor if cmap is None else cmap(0.32))
+        ]
         edgecolor = edgecolor if cmap is None else None
-        linewidths = kwargs.pop('linewidths', 1.5)
-        ax.patches += [plt.Rectangle((0, 0), 0, 0, fc=None, ec=edgecolor,
-                                     lw=2, label=label)]
 
+    vmin, vmax = match_contour_to_contourf(contours, vmin=0, vmax=p.max())
     cont = ax.tricontour(tri, p, contours, zorder=zorder,
-                         vmin=0, vmax=0.32 * p.max(), linewidths=linewidths,
+                         vmin=vmin, vmax=vmax, linewidths=linewidths,
                          colors=edgecolor, cmap=cmap, *args, **kwargs)
 
     ax.set_xlim(*check_bounds(tri.x, xmin, xmax), auto=True)
--- a/anesthetic/plot.py
+++ b/anesthetic/plot.py
@@ -623,7 +623,7 @@ def kde_contour_plot_2d(ax, data_x, data_y, *args, **kwargs):
         ax.patches += [plt.Rectangle((0, 0), 0, 0, fc=None, ec=edgecolor,
                                      lw=2, label=label)]
 
-    cont = ax.tricontour(tri, p, contours, zorder=zorder, vmin=0, vmax=p.max(),
+    cont = ax.tricontour(tri, p, contours, zorder=zorder, vmin=0, vmax=0.32 * p.max(),
                          linewidths=linewidths, colors=edgecolor, cmap=cmap,
                          *args, **kwargs)
 
--- a/anesthetic/plot.py
+++ b/anesthetic/plot.py
@@ -526,6 +526,11 @@ def kde_contour_plot_2d(ax, data_x, data_y, *args, **kwargs):
         A set of contourlines or filled regions
 
     """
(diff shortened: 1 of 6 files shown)


anesthetic/plot.py
```diff
@@ -269,6 +269,9 @@ def fastkde_plot_1d(ax, data, *args, **kwargs):
 
     xmin = kwargs.pop('xmin', None)
     xmax = kwargs.pop('xmax', None)
+    cmap = kwargs.pop('cmap', None)
+    color = kwargs.pop('color', (next(ax._get_lines.prop_cycler)['color']
+                                 if cmap is None else cmap(2/3)))
```
williamjameshandley, line 274: Probably safer for those used to reading python2 to make this 2./3., or 0.66 rather than 2/3 (since in python2, this would be 0). Better to be explicit (even though it doesn't matter for python3). Should be changed throughout.

anesthetic/plot.py
```diff
@@ -526,6 +536,13 @@ def kde_contour_plot_2d(ax, data_x, data_y, *args, **kwargs):
         A set of contourlines or filled regions
 
     """
+    kwargs = cbook.normalize_kwargs(kwargs,
+                                    dict(linewidths=['linewidth', 'lw'],
+                                         linestyles=['linestyle', 'ls'],
+                                         color=['c'],
+                                         facecolor=['fc'],
+                                         edgecolor=['ec']),
+                                    forbidden=['colors'])
```
williamjameshandley, line 545: I'm getting quite a lot of the warning message:
```python
MatplotlibDeprecationWarning:             
The 'forbidden' parameter of normalize_kwargs() was deprecated in Matplotlib 3.3 and will be removed two minor releases later. If any parameter follows 'forbidden', they should be passed as keyword, not positionally.                                                                           
```

Is this something that we should be worried about?

lukashergt: Maybe, I've been swamped with matplotlib warnings lately (probably due to my stylesheet going out of date) that I didn't notice this particular one. Will look into it.

lukashergt:
> At the moment there is not much guidance to users as to how to use this new functionality. You should either add a note in the docs (which may not be a good fit), or an example to the plot gallery (`demo.py`). This should then also be converted to the corresponding (`demo.ipynb`) using [py2nb](https://github.com/williamjameshandley/py2nb)

Been meaning to add a minimal script to this, but forgot. Here is what this looks like:

```python
ns = NestedSamples(root="/home/hergtl/Documents/Projects/AnestheticPrj/anesthetic/tests/example_data/pc")
fig = plt.figure(figsize=(8, 8))
fig, axes = make_2d_axes(['x0', 'x1', 'x2', 'x3', 'x4'], fig=fig)
ns.plot_2d(axes, lower_kwargs=dict(fc=None));
```

unfilled

Should I add this to the demo?

lukashergt:
> If I understand correctly, you can now turn off the filling by providing the `facecolor=None` kwarg to the contour plots. 

Correct.

> Is this something that should also be added to the fastkde plots as well?

Yes (see my TODO in initial PR post), just wanted to put that off till we have the final version of the desired behaviour.

williamjameshandley:
I think this is a relatively consistent way to turn off the contour filling, although the concern would be that `facecolor` or `fc` are not obviously 'don't fill' to newer users. Do you think it would be worth providing a `fill=False` alternative kwarg as well?

williamjameshandley:
That is a very cool plot btw. Do you think it would look better if the 2 sigma contours were a lighter shade, or would that be confusing in practice?

anesthetic/plot.py
lukashergt: done

anesthetic/plot.py
lukashergt: I removed the `forbidden` parameter. This wasn't crucial anyhow, I just put it in as a reminder/early catch that anesthetic does not work with kwarg `colors`. This is because contour plots only allow either `cmap` or `colors` as input, not both. This is not a new thing of this PR.

lukashergt:
> That is a very cool plot btw. Do you think it would look better if the 2 sigma contours were a lighter shade, or would that be confusing in practice?

That's what the `cmap` kwarg is for.

The above plot could also be gotten by explicitly calling `ec` which otherwise inherits from `color`:

```python
ns.plot_2d(axes, lower_kwargs=dict(fc=None, ec='C0'));
```

If you'd rather have a `cmap`, do e.g.:
```python
ns.plot_2d(axes, lower_kwargs=dict(fc=None, cmap=plt.cm.Blues));
```

lukashergt:
> I think this is a relatively consistent way to turn off the contour filling, although the concern would be that `facecolor` or `fc` are not obviously 'don't fill' to newer users. Do you think it would be worth providing a `fill=False` alternative kwarg as well?

Anesthetic's philosophy is trying to go hand in hand with matplotlib, and `facecolor` seems to me the logical choice from a matplotlib mindset. Introducing `fill` would unnecessarily add to/clutter the name space, in my opinion.

anesthetic/plot.py
```diff
@@ -450,15 +464,22 @@ def fastkde_contour_plot_2d(ax, data_x, data_y, *args, **kwargs):
...
     ymin = kwargs.pop('ymin', None)
     ymax = kwargs.pop('ymax', None)
     label = kwargs.pop('label', None)
     zorder = kwargs.pop('zorder', 1)
-    linewidths = kwargs.pop('linewidths', 0.5)
     levels = kwargs.pop('levels', [0.68, 0.95])
     color = kwargs.pop('color', next(ax._get_lines.prop_cycler)['color'])
+    facecolor = kwargs.pop('facecolor', color)
+    edgecolor = kwargs.pop(
+        'edgecolor', color if facecolor in [None, 'None', 'none'] else 'k')
```
lukashergt, line 482: The matplotlib standard for markers or similar is to use `'None'` or `'none'` to make them unfilled, because `None` typically is used for automatic color assignment. This is a common gotcha for many matplotlib users, and since we don't have any use for `None` here, I've added it to the list of values requesting unfilled contours.

demo.ipynb
```diff
@@ -153,6 +154,72 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "fig, axes = make_2d_axes(['omegabh2', 'omegach2', 'H0'], tex=mcmc.tex)\n",
+    "mcmc.plot_2d(axes.iloc[0:1, :], types=dict(upper='kde', lower='kde', diagonal='kde'), fc=None);\n",
+    "mcmc.plot_2d(axes.iloc[1:2, :], types=dict(upper='kde', lower='kde', diagonal='kde'), fc=None, cmap=plt.cm.Oranges, lw=3);\n",
+    "mcmc.plot_2d(axes.iloc[2:3, :], types=dict(upper='kde', lower='kde', diagonal='kde'), fc='C2', ec='C3', c='C4', lw=2);"
```
lukashergt, line 220: This is looking a bit gaudy, hope that's ok. It does provide a quite detailed example of how to customise, though.

anesthetic/plot.py
```diff
@@ -566,22 +605,31 @@ def kde_contour_plot_2d(ax, data_x, data_y, *args, **kwargs):
...
-        c.set_cmap(cmap)
-
-    ax.tricontour(tri, p, contours, zorder=zorder,
-                  vmin=0, vmax=p.max(), linewidths=linewidths, colors='k',
-                  *args, **kwargs)
-    ax.patches += [plt.Rectangle((0, 0), 0, 0, fc=cmap(0.999), ec=cmap(0.32),
-                                 lw=2, label=label)]
+    cont = ax.tricontour(tri, p, contours, zorder=zorder,
+                         vmin=0, vmax=0.32 * p.max(), linewidths=linewidths,
+                         colors=edgecolor, cmap=cmap, *args, **kwargs)
```
lukashergt, line 628: I was surprised at very faint contour lines when using `cmap`. Turns out this had to do with `vmax=p.max()`. In the filled case, this is not noticeable, because `vmax` is ignored when `colors` is a constant. Setting `vmax=0.32 * p.max()` sets the color of the inner contour line to the extreme end of `cmap`. Happy with this behaviour?

anesthetic/plot.py
```diff
@@ -534,8 +571,10 @@ def kde_contour_plot_2d(ax, data_x, data_y, *args, **kwargs):
     ncompress = kwargs.pop('ncompress', 1000)
     label = kwargs.pop('label', None)
     zorder = kwargs.pop('zorder', 1)
-    linewidths = kwargs.pop('linewidths', 0.5)
     color = kwargs.pop('color', next(ax._get_lines.prop_cycler)['color'])
+    facecolor = kwargs.pop('facecolor', color)
+    edgecolor = kwargs.pop(
+        'edgecolor', color if facecolor in [None, 'None', 'none'] else 'k')
```
lukashergt, line 577: Is `edgecolor='k'` our favourite default? Another option would be `edgecolor=color`...

anesthetic/plot.py
lukashergt: The hard coded `0.32` are due to the hard coded `contours=[0.68, 0.95]` in `iso_probability_contours_from_samples`.

demo.ipynb
williamjameshandley: This is fine, as it it quite fine-grained control, although I agree that in anesthetic 2.0 we should probably have a cleaner api. You could probably clean it up by defining `kde=dict(upper='kde', lower='kde', diagonal='kde')`, and passing `types=kde` to each of them.

anesthetic/plot.py
williamjameshandley: >I was surprised at very faint contour lines when using cmap

Could you provide a picture to show this behaviour, I'm not sure exactly what you mean

anesthetic/plot.py
williamjameshandley: Could you provide a picture to show the difference here too?

anesthetic/plot.py
lukashergt: Both following plots were generated with:

```python
from matplotlib.colors import LinearSegmentedColormap as LSC
ns = NestedSamples(root=anesthetic/tests/example_data/pc")
fig = plt.figure(figsize=(8, 8))
fig, axes = make_2d_axes(['x0', 'x1', 'x2', 'x3', 'x4'], fig=fig)
ns.plot_2d(axes, fc=None, c='C0', cmap=LSC.from_list('', ['w', 'C0']));
```

This is with `vmax=0.32*p.max()` (new behaviour):

unfilled

This is with `vmax=p.max()` (old behaviour):

unfilled

lukashergt: Note, that this behaviour also applies to the filled contours. If you look closely, the inner contour does not have the same saturated hue as lines in the 1d plots. For the filled contours, I think this is actually desirable, but for line contours this seems a bit too washed out.

What is special in our case is that we have only 2 levels, which is not the typical usecase of these contour plots.

anesthetic/plot.py
lukashergt: Here a direct comparison of the two cases. Black contours make plots a bit crisper on their own, however it becomes cluttered more quickly when having multiple contours per axes.

Generated with:
```python
fig = plt.figure(figsize=(8, 8))
fig, axes = make_2d_axes(['x0', 'x1', 'x2', 'x3', 'x4'], fig=fig)
ns.plot_2d(axes, c='C0', ec='k' , types=dict(lower='kde', diagonal='kde'));
ns.plot_2d(axes, c='C0', ec='C0', types=dict(upper='kde'));
```

unfilled

williamjameshandley: Nice! I really don't know which I prefer, but here are arguments in favour of `'k'`:
- Preserving backward-compatible behaviour
- The other variant will be more useful for large, complicated plots, where code compactness is probably less useful.
I imagine I'll make a lot of use of this regardless, so this is an excellent addition.

anesthetic/plot.py
williamjameshandley: This makes it a lot clearer.

anesthetic/plot.py
lukashergt: Ok, I'll keep `'k'` as default, then.

anesthetic/plot.py
lukashergt: Hmm, there seems to be a more general problem with the calculation of `contours` in `kde_plot_2d`. Can you see how  the outer contour lines in `x0-x3`, `x1-x3` and `x3-x4` are darker than in the other plots? This is independent of filled/unfilled contours (see e.g. plot below comparing black to blue edges), but also happens in the current master, just less noticeable because the black edges hide the (lacking) contrast to the white background.

I've looked at what contours computes and got:
```
01, contours=[2.000, 10.996, 34.730],    level2 / level3 = 0.317,    level1 / level3 = 0.058
02, contours=[0.959,  7.355, 28.050],    level2 / level3 = 0.262,    level1 / level3 = 0.034
12, contours=[0.788,  6.859, 26.510],    level2 / level3 = 0.259,    level1 / level3 = 0.030
03, contours=[0.693,  2.406,  4.204],    level2 / level3 = 0.572,    level1 / level3 = 0.165   *
13, contours=[0.730,  2.415,  4.095],    level2 / level3 = 0.590,    level1 / level3 = 0.178   *
23, contours=[0.477,  3.865,  8.169],    level2 / level3 = 0.473,    level1 / level3 = 0.058
04, contours=[0.319,  1.178,  4.240],    level2 / level3 = 0.278,    level1 / level3 = 0.075
14, contours=[0.288,  1.176,  4.072],    level2 / level3 = 0.289,    level1 / level3 = 0.071
24, contours=[0.240,  1.974,  7.138],    level2 / level3 = 0.277,    level1 / level3 = 0.034
34, contours=[0.244,  0.563,  1.234],    level2 / level3 = 0.456,    level1 / level3 = 0.198   *
```
Note the lines marked with `*` deviate from the expected `0.32` and `0.05` values associated with 68% and 95% contours.

Is maybe something wrong in `anesthetic.utils.iso_probability_contours_from_samples`? Or are outliers such as these expected?

lukashergt: Maybe related, I've noticed that the maximum width of the posterior contours in the `*` marked axes are a bit slimmer than in the others. Wouldn't you expect the maximum width to be roughly consistent across axes?

williamjameshandley: This difference in colouration is a feature rather than a bug -- it's actually quietly mentioned on the front page of the repo:

>Many KDE plotting tools have conventions that don't play well with uniformly distributed parameters, which presents a problem if you are trying to plot priors along with your posteriors. anesthetic has a sensible mechanism, by defining the contours by the amount of iso-probability mass they contain, but colouring the fill in relation to the probability density of the contour.

This makes a big difference when plotting priors. For an example, have a look at the blue or orange contours in [Figure 4](https://arxiv.org/pdf/1908.09139v1.pdf). If these were coloured in accordance with mass, rather than density, the picture would misleadingly look like a constraint.

williamjameshandley: However, I agree that the contour lines themselves, if not coloured `'k'`, should be the same colour as the 68% region.

lukashergt: > However, I agree that the contour lines themselves, if not coloured `'k'`, should be the same colour as the 68% region.

Except they weren't. My thingy was an ugly hack. I have now improved things, such that the colours from `contour` match the colours from `contourf` when using a colour map. The problem was that `contourf` assigns colours to the region inbetween levels, whereas `contour` assigns colours to the level itself...

Sorry for yet another review request... I think I've got it all now.

lukashergt:
This is strange. Tests pass locally for me... Could you confirm whether tests fail locally for you @williamjameshandley?

lukashergt:
> This is strange. Tests pass locally for me... And I am running the latest matplotlib... Could you confirm whether tests fail locally for you @williamjameshandley?

#115 shows that tests fail even on the master branch, so this is not related to this PR.

Issue #86, opened 2020-05-31
Possible anti-pattern with current weights behaviour


williamjameshandley:
This issue arose from #84. The relevent text is copied, edited an collated from that PR for further discussion here @lukashergt, @appetrosyan

The code `samples['weight'][i] = 0` doesn't actually do anything, since `samples['weight']` is different from `samples.weight`, and it is the latter which is used for computing the weighted sums. This possible anti-pattern was introduced in #55 and #56. I'm not sure of the best way to ameliorate this clearly confusing state of affairs. One could:
- drop the weight as a column (but this then makes it less clear that it is a weighted frame)
    - @lukashergt : "I don't like this solution. It is useful to see the weight column when looking at the full table."
- make the weight column read-only (don't know if this is easy to do in pandas, and would likely cause no end of bother down the line)
   - @lukashergt : "Well, it seems like this is already the case, right? Trying to overwrite the weight column actually doesn't work..."
       - This is only partially true. It's worse than doing nothing, since users might think that they have done something when they haven't. [True read-only](https://stackoverflow.com/a/5541452) would raise an exception when a user tried to write to the column.
- synchronise the weight column with the weight attribute more effectively. This would require someone with an understanding of how the weighted data frames work in `anesthetic/weighted_pandas.py` which is currently just me at the moment (AFAIK).
    - It would be better practice if more than one person understood this portion of the codebase, otherwise the risk is increased that hidden bad behaviour is not being discovered.

------

The issue is that there is a hidden attribute in the `WeightedSeries` and `WeightedDataFrame` called `_weight`. 

The hidden attribute is necessary for `WeightedSeries`, as if you want the pandas equivalent of a `Series` you can only have one column, so the weights have to be an attribute. This is forwarded as a readable attribute via a `@property`, and the correct behaviour is also enforced for slicing.

The same was originally true for `WeightedDataFrame`, but this was slightly counterintuitive as you would not see the weights in the same way that you see `nlive` or other derived `NestedSamples` columns. The contents of `_weight` is therefore copied into a new column whenever you change `_weight` using the anesthetic functions, but if you go and change this new column manually it has no effect on the weighted calculations (like `mean`, `std`, `median` etc)
- synchronise the weight column with the weight attribute more effectively. This would require someone with an understanding of how the weighted data frames work in `anesthetic/weighted_pandas.py` which is currently just me at the moment (AFAIK).

lukashergt:
[...]

Issue #83, opened 2020-05-29
Possible issue in dlogX calculation


williamjameshandley:
As discovered  by @appetrosyan in #81 there is a possible lurking off-by-one error in the`dlogX` calculation that is hidden by a `numpy.errstate` section. This issue is designed for further discussion on this point.

williamjameshandley:
This is something that I will fix in a PR shortly alongside a couple of extra corner-case fixes to the nlive calculation that I have seen in the past week.

Issue #90, opened 2020-06-04
Weight normalisation


williamjameshandley:
Currently in anesthetic the nested sampling weights are normalised so that the maximum weight is 1. The other convention (more consistent with multinest, and possibly safer) is to normalise so that they sum to 1. I don't think this would be a breaking change for anybody, since if you were to be using the current weights directly, you would probably be normalising anyway.

What are people's views on changing this? @lukashergt @appetrosyan @Pablo-Lemos 

This is a change that could likely be implemented at the same time as fixing #83

appetrosyan:
Can we create a new function and deprecate the old one? I would argue that the normalisation convention is an internal implementation issue, so it shouldn't be directly exposed. I would add a layer of getters and setters and change the internal representation.

lukashergt:
Either way would work for me. 

Important to me is that subsamples/slicing/masking/etc. work as expected. That means that we shouldn't rely on a certain normalisation within the code. In places where we need the weights normalised by their sum we should use `w/w.sum()` and not just `w` which would only be normalised for the full sample.

williamjameshandley:
Closing this for now -- going with the status quo, rather than making a change. Feel free to resurrect this issue if you feel strongly about weights being normalised to unity.

Issue #59, opened 2019-09-11
Closer mirroring of pandas


williamjameshandley:
One thing that has been concerning me for a while is that the `MCMCSamples.plot` command does not quite mirror the same behaviour as `pandas.DataFrame.plot`, and its overriding could be somewhat a surprise to the uninitiated or a restriction to the aficionado. Changing `MCMCSamples.plot` to say `MCMCSamples._plot` would be a mild breaking change.

A larger-scale change to the code could be the introduction of `MCMCSeries` and `NestedSeries` as the series-sliced versions of `MCMCSamples` and `NestedSamples`. This would allow code like:
```python
samples = MCMCSamples(root=file_root)
samples.x0.plot_1d()
```
for plotting single 1d distributions.

It also might be good to have an option to 'plot all', with possibly a guard against too many parameters. For example:
```python
samples[['x0','x1','x2']].plot_1d()
samples[['x0','x1','x2']].plot_2d()
```
would plot the three selected parameters. If one tried to plot more than 100 things (i.e. a 10x10 2d plot) this way, there could be an exception that is raised unless you explicitly pass `plot_all=True`

williamjameshandley:
One change we should make is for `plot_2d` and `plot_1d` to just return `Axes`, rather than `Figure, Axes`, in order to be more consistent with `pandas.plot`. (see the PR which will soon be submitted from the MCMC_samples branch).

PR #115, opened 2020-07-28
Pandas 1.1 update

lukashergt:
Tests were failing in #114 but it seemed unrelated to any of my commits and tests passed locally.

This PR starts out as the pure master branch and shows that this fails, so there is an issue unrelated to #114.

Maybe something to do with updates of some dependencies? However, my local virtualenv is up to date...

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -3,7 +3,7 @@ anesthetic: nested sampling visualisation
 =========================================
 :anesthetic: nested sampling visualisation
 :Author: Will Handley
-:Version: 1.3.6
+:Version: 1.4.0
 :Homepage: https://github.com/williamjameshandley/anesthetic
 :Documentation: http://anesthetic.readthedocs.io/
 
--- a/README.rst
+++ b/README.rst
@@ -7,8 +7,8 @@ anesthetic: nested sampling visualisation
 :Homepage: https://github.com/williamjameshandley/anesthetic
 :Documentation: http://anesthetic.readthedocs.io/
 
-.. image:: https://travis-ci.org/williamjameshandley/anesthetic.svg?branch=master
-   :target: https://travis-ci.org/williamjameshandley/anesthetic
+.. image:: https://travis-ci.com/williamjameshandley/anesthetic.svg?branch=master
+   :target: https://travis-ci.com/williamjameshandley/anesthetic
    :alt: Build Status
 .. image:: https://circleci.com/gh/williamjameshandley/anesthetic.svg?style=svg
    :target: https://circleci.com/gh/williamjameshandley/anesthetic
--- a/README.rst
+++ b/README.rst
@@ -2,7 +2,7 @@
 anesthetic: nested sampling visualisation
 =========================================
 :anesthetic: nested sampling visualisation
-:Author: Will Handley
+:Author: Will Handley and Lukas Hergt
 :Version: 1.4.0
 :Homepage: https://github.com/williamjameshandley/anesthetic
 :Documentation: http://anesthetic.readthedocs.io/
--- a/README.rst
+++ b/README.rst
@@ -3,7 +3,7 @@ anesthetic: nested sampling visualisation
 =========================================
 :anesthetic: nested sampling visualisation
 :Author: Will Handley and Lukas Hergt
-:Version: 1.4.0
+:Version: 2.0.0-beta.1
 :Homepage: https://github.com/williamjameshandley/anesthetic
 :Documentation: http://anesthetic.readthedocs.io/
 
--- a/anesthetic/weighted_pandas.py
+++ b/anesthetic/weighted_pandas.py
@@ -145,9 +145,9 @@ class WeightedDataFrame(_WeightedObject, pandas.DataFrame):
         """
         i = compress_weights(self.weights, self._rand, nsamples)
         data = np.repeat(self.values, i, axis=0)
-        index = np.repeat(self.index.values, i)
+        index = self.index.repeat(i)
         df = pandas.DataFrame(data=data, index=index, columns=self.columns)
-        df.index = df.index.get_level_values(0)
+        df.index = df.index.get_level_values('#')
         return df
 
     @property
--- a/anesthetic/weighted_pandas.py
+++ b/anesthetic/weighted_pandas.py
@@ -3,7 +3,7 @@
 import numpy as np
 import pandas
 from anesthetic.utils import (compress_weights, channel_capacity, quantile,
-                              temporary_seed, array_to_seed)
+                              temporary_seed)
 
 
 class _WeightedObject(object):
--- a/anesthetic/weighted_pandas.py
+++ b/anesthetic/weighted_pandas.py
@@ -12,14 +12,13 @@ class _WeightedObject(object):
         if self.index.nlevels == 1:
             return pandas.Series(index=self.index, data=1.)
         else:
-            return pandas.Series(data=self.index.droplevel(0),
-                                 index=self.index)
+            return self.index.get_level_values('weight').to_numpy()
 
     @weight.setter
     def weight(self, weight):
         if weight is not None:
             self.index = [self.index.get_level_values(0), weight]
-            self.index.set_names([None, 'weight'], inplace=True)
+            self.index.set_names(['#', 'weight'], inplace=True)
 
     @property
     def _rand(self):
@@ -62,7 +61,7 @@ class WeightedSeries(_WeightedObject, pandas.Series):
 
     def quantile(self, q=0.5):
         """Weighted quantile of the sampled distribution."""
-        return quantile(self.values, q, self.weight.values)
+        return quantile(self.values, q, self.weight)
 
     def hist(self, *args, **kwargs):
         """Weighted histogram of the sampled distribution."""
--- a/anesthetic/weighted_pandas.py
+++ b/anesthetic/weighted_pandas.py
@@ -2,7 +2,8 @@
 
 import numpy as np
 import pandas
-from anesthetic.utils import compress_weights, channel_capacity, quantile
+from anesthetic.utils import (compress_weights, channel_capacity, quantile,
+                              temporary_seed, array_to_seed)
 
 
 class _WeightedObject(object):
@@ -23,9 +24,9 @@ class _WeightedObject(object):
     @property
     def _rand(self):
         """Random number for consistent compression."""
-        h = pandas.util.hash_array(self.index)
-        _rand = np.frexp(h)[0]*2-1
-        return pandas.Series(data=_rand, index=self.index)
+        seed = pandas.util.hash_pandas_object(self.index).sum() % 2**32
+        with temporary_seed(seed):
+            return np.random.rand(len(self))
 
     def std(self):
         """Weighted standard deviation of the sampled distribution."""
--- a/anesthetic/weighted_pandas.py
+++ b/anesthetic/weighted_pandas.py
@@ -2,7 +2,6 @@
 
 import numpy as np
 import pandas
-from warnings import warn
 from anesthetic.utils import compress_weights, channel_capacity, quantile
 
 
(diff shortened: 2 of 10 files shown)


williamjameshandley:
This has arisen following the [pandas 1.1 update](https://pandas.pydata.org/pandas-docs/dev/whatsnew/v1.1.0.html), which is not yet even on [arch](https://www.archlinux.org/packages/community/x86_64/python-pandas/) yet. I fresh installed in a new venv and managed to replicate the error (python can be a bit fickle with which versions of pytest you use if you have one installed globally, so you should make sure you run `python -m pytest` rather than just `pytest`, which may show you that `pytest` is not pip installed)

williamjameshandley:
OK. This is actually a pretty big project, which will involve updating how the weighted data frames are treated, so this PR should also aim to fix #86 #90 and #95 at the same time. In the mean time, we should freeze pandas at the previous version, which is what #116 does

williamjameshandley:
OK. Big update, although interestingly not that many lines of code (actually removed code here).

Executive summary: 

Implemented the discussion in #86 that moves the `['weights']` column and the `_weight` hidden property to a `pandas.MultiIndex`. This is better because:
- it more explicitly something that is clearly carried around between `WeightedDataFrames` and `WeightedSeries`
- it involves less code and overhead, getting around an issue that in the update to 1.1, pandas now assumes that the hack for `_constructor_expanddim` which replaced a class with a constructor function is no longer viable, as the constructor function would require the `@classmethods` associated with `WeightedSeries`
- it resolves #86 in that there is now only one set of weights that are carried around.

This is a (admittedly minor) API breaking change (as can be seen by the fact that I had to modify previously passing tests), so we should update to 1.4 at the least, but if we're being proper this should really a major version update to 2.0.0.

Some possible gotchas/things to comment on:
1. `samples.weight` at the moment returns a `pandas.Series` with the same `pandas.MultiIndex` as the original array. I would value opinion as to whether this should instead just be a straight numpy array (which would likely be cleaner, and is more obviously a copy rather than a view).
2. `samples.weight` can be assigned an array: `samples.weight = w`, which may be very useful for re-weighting purposes.
3. `samples.weight[0] = 0` does nothing (still), since this is a copy, rather than a view (and I've spent most of my time for this PR working out if the latter is possible, and I don't think it is). This possibly lends further weight to point 1.
3. `samples['weight']` is now gone, which is really the only breaking change (as `samples._weight` is underscored as a warning anyway).

I also took took this opportunity to deprecate the `'w'` keyword argument, although one of the things that has now started annoying me is the inconsistency between `'weights'` in matplotlib and `MCMCSamples` and `'weight'` in the `WeightedDataFrame`. There is a case for making this change now, particularly if we also update the major version number.

I would also like to credit @lukashergt, as the key change is entirely his idea. At 5000 lines of code contributions and what I believe is a serious improvement to the infrastructure of anesthetic he should now be added as a co-author on the front page.

williamjameshandley:
P.S, @lukashergt: as you are the originator of the PR, I can't nominate you as a reviewer (nor can you pass provide an approving review), but as I wrote the code you should probably take a look.

williamjameshandley:
One final slightly hacky thing that came with this was an update to the `_rand` functionality, which is now deterministically generated from the `MultiIndex`. I couldn't think of a neater way to do this post pandas 1.1, but it is something to watch out for.

anesthetic/weighted_pandas.py
```diff
@@ -2,31 +2,31 @@
...
         else:
-            return self._weight[self.index]
+            return pandas.Series(data=self.index.droplevel(0),
+                                 index=self.index)
+
+    @weight.setter
+    def weight(self, weight):
+        if weight is not None:
+            self.index = [self.index.get_level_values(0), weight]
+            self.index.set_names([None, 'weight'], inplace=True)
```
lukashergt, line 22: You chose not to give a name to the iteration index. I agree that it is not as necessary from a user point of view to have a name displayed in the table. However, for our future selves looking at this code a name might be clearer than `0` and `None`. Maybe something as simple as `'#'` would already be clearer for us?

tests/test_samples.py
```diff
@@ -499,27 +503,28 @@ def test_merging():
...
     weight = pc.weight
-    assert_array_equal(pc['weight'], pc.weight)
+    assert_array_equal(weight, pc.weight)
+    assert_array_equal(pc.index.get_level_values(1), pc.weight)
     assert pc.beta == 1
 
     prior = pc.set_beta(0)
     assert prior.beta == 0
-    assert_array_equal(prior['weight'], prior.weight)
+    assert_array_equal(prior.index.get_level_values(1), prior.weight)
```
lukashergt, line 512: Would using `'weight'` instead of `1` be more future proof, here and in the following lines?

anesthetic/weighted_pandas.py
```diff
@@ -2,31 +2,31 @@
...
-    def _construct_weights(self, weight):
-        if weight is not None:
-            self._weight = pandas.Series(index=self.index, data=weight)
-        else:
-            self._weight = None
-        rand = np.random.rand(len(self))
-        self._rand_ = pandas.Series(index=self.index, data=rand)
+        h = pandas.util.hash_array(self.index)
+        _rand = np.frexp(h)[0]*2-1
+        return pandas.Series(data=_rand, index=self.index)
```
lukashergt, line 29: > One final slightly hacky thing that came with this was an update to the _rand functionality, which is now deterministically generated from the MultiIndex. I couldn't think of a neater way to do this post pandas 1.1, but it is something to watch out for.

I'm unfamiliar with the `np.frexp` function, does this work like a random number generator? Could we alternatively use the pandas hash as seed for a random number generator?

williamjameshandley:
Changes in 624554e

>I agree that a numpy array would be cleaner. If we were to keep a pandas.Series, though, I feel we at least should drop the weight index, here.

Doesn't work with only a `'#'` index here, as the indices need to match if you say are indexing a pandas array with a boolean pandas array (e.g. in `samples[samples.weight!=0]`). It works fine as a numpy array. So let's go with this.

>You chose not to give a name to the iteration index. I agree that it is not as necessary from a user point of view to have a name displayed in the table. However, for our future selves looking at this code a name might be clearer than 0 and None. Maybe something as simple as '#' would already be clearer for us?

I quite like the look of `'#'`, so have gone for this.

anesthetic/weighted_pandas.py
```diff
@@ -45,13 +45,9 @@ class WeightedSeries(_WeightedObject, pandas.Series):
...
     def __init__(self, *args, **kwargs):
-        if 'w' in kwargs:
-            warn("'w' as a kwarg will be deprecated in the future. "
-                 "Please use 'weight'", FutureWarning)
-        weight = kwargs.pop('w', None)
-        weight = kwargs.pop('weight', weight)
+        weight = kwargs.pop('weight', None)
         super(WeightedSeries, self).__init__(*args, **kwargs)
-        self._construct_weights(weight)
+        self.weight = weight
```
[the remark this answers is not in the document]
williamjameshandley: Unfortunately not. Pandas idiosyncratic method of extension alongside diamond inheritance is a pretty fine line to tread. We should be glad it looks this simple (despite the code duplication).

README.rst
```diff
@@ -2,8 +2,8 @@
 anesthetic: nested sampling visualisation
 =========================================
 :anesthetic: nested sampling visualisation
-:Author: Will Handley
-:Version: 1.3.6
+:Author: Will Handley and Lukas Hergt
+:Version: 1.4.0
```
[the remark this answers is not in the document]
williamjameshandley: I think it probably is time for a 2.0.0 overhaul, so will promote the version number to 2.0.0-beta, and treat 1.3.6 after #117 is merged as the last stable version of anesthetic 1. 

We may want to consider #59 as part of this.

williamjameshandley:
OK, with the change from weight -> weights, this is getting to be a pretty gnarly PR, so we should probably save this as a 'WeightedDataFrame' overhaul, merge it as 2.0.0-beta, and then start flagging things that we also want to fix in the context of the eventual 2.0.0 release.

williamjameshandley:
As this is technically your PR @lukashergt , you should merge it, but only after you're happy with all the changes (feel free to push any adjustments that you think are needed)

williamjameshandley:
Fixes #95

williamjameshandley:
We could of course use the weight -> weights change to adjust the normalisation so that it sums to 1 by default, but I agree that as we now add re-weighting and slicing we should definitely not assume anywhere in the code that this is true.

anesthetic/weighted_pandas.py
```diff
@@ -157,24 +143,16 @@ def compress(self, nsamples=None):
             compression). If <=0, then compress so that all weights are unity.
 
         """
-        i = compress_weights(self.weight, self._rand, nsamples)
+        i = compress_weights(self.weights, self._rand, nsamples)
         data = np.repeat(self.values, i, axis=0)
-        index = np.repeat(self.index.values, i)
+        index = self.index.repeat(i)
```
lukashergt, line 148: Using `index.repeat` instead of `np.repeat` keeps the same `MultiIndex` structure. With `np.repeat` the `'#'` and `weights` index were turned into a tuple such that `compress` returned:
```python
> print(ns_compressed[:5])

                                   x0        x1      logL  logL_birth  nlive
(476, 0.00975239917398711)  -0.236910 -0.284601 -4.606267   -6.359672  125.0
(495, 0.016423622996336068) -0.004446  0.162164 -3.933665 -556.623795  125.0
(521, 0.03131786144184918)  -0.035268 -0.094386 -3.081024   -8.523751  125.0
(554, 0.054268694486099424) -0.115377 -0.128187 -2.268316   -2.823167  125.0
(568, 0.07583551199263458)   0.263529  0.117131 -1.822142   -1.935448  125.0
```

now it returns:
```python
> print(ns_compressed[:5])

           x0        x1      logL  logL_birth  nlive
#                                                   
476 -0.236910 -0.284601 -4.606267   -6.359672  125.0
495 -0.004446  0.162164 -3.933665 -556.623795  125.0
521 -0.035268 -0.094386 -3.081024   -8.523751  125.0
554 -0.115377 -0.128187 -2.268316   -2.823167  125.0
568  0.263529  0.117131 -1.822142   -1.935448  125.0
```

where  
```python
ns = ns[['x0', 'x1', 'logL', 'logL_birth', 'nlive']]
ns_compressed = ns.compress(1000)
```

lukashergt:
> We could of course use the weight -> weights change to adjust the normalisation so that it sums to 1 by default, but I agree that as we now add re-weighting and slicing we should definitely not assume anywhere in the code that this is true.

Yeah, maybe it is good the way it is, where it is instantaneously clear from the last element (with weight 1) that the `weights` column does not sum to 1, making it less likely that people make wrong assumptions.

anesthetic/weighted_pandas.py
williamjameshandley: Nice!

williamjameshandley:
I think I'll leave a consideration of normalisation to when I fix #83, as I'm still undecided since the pros and cons seem to be pretty balanced.

tests/test_weighted_pandas.py
```diff
@@ -188,6 +189,16 @@ def test_WeightedSeries_quantile():
...
 
 
+def test_WeightedSeries_hist():
+    series = test_WeightedSeries_constructor()
+    ax = series.hist(bins=20, density=True)
+    assert len(ax.patches) == 20
+    norm = 0
+    for patch in ax.patches:
+        norm += patch.get_height() * patch.get_width()
+    assert norm == pytest.approx(1)
```
lukashergt, line 199: Any idea why this is failing but the equivalent one for `WeightedDataFrame` is not?

The following runs locally without a problem for me:
```python
import pytest
from anesthetic.weighted_pandas import WeightedSeries

series = WeightedSeries(np.random.randn(1000), weights=np.random.rand(1000))

ax = df.hist(bins=20, density=True)

assert len(ax.patches) == 20

norm = 0
for patch in ax.patches:
    norm += patch.get_height() * patch.get_width()
print(norm)
assert norm == pytest.approx(1)
```

williamjameshandley: Now that was a non-trivial debug. Was able to replicate it locally, but only if running the tests in series with:
```bash
python -m pytest -k hist
```
Running it in parallel as I usually do with
```bash
./bin/run_tests
```
passed. I suspect this is therefore to do with running the suite in such a way that means that, when calling the parent function `Series.hist` or `WeightedFrame.hist` whilst other axes are still open, it tries to replot on a still open axis (as it would do if you call `plt.plot(...)` when there is  an axis that has not yet been closed).

Explicitly closing the plots seems to have fixed it.

lukashergt: Nice! I had the same issue, running in parallel worked but in series didn't. I tried around with `plt.close` but didn't know about the `"all"` option.

lukashergt:
Ok, so with [a185285](https://github.com/williamjameshandley/anesthetic/pull/115/commits/a185285a75f91f505110516f55cbeeeff40fde92) we have 100% patch coverage, so I'd be ready to merge as soon as you are happy with everything, too.

PR #116, opened 2020-07-29
Freezing pandas at 1.0.5

williamjameshandley:
# Description

Following #115  which notes that master is now broken on account of the pandas update to 1.1, this PR freezes to pandas to 1.0.5 until #115 is resolved

# Checklist:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -31,7 +31,7 @@ setup(name='anesthetic',
       url='https://github.com/williamjameshandley/anesthetic',
       packages=find_packages(),
       scripts=['scripts/anesthetic'],
-      install_requires=['numpy', 'scipy', 'pandas<=1.0.5', 'matplotlib'],
+      install_requires=['numpy', 'scipy>=1.2.0', 'pandas<=1.0.5', 'matplotlib>=3.1.2'],
       setup_requires=['pytest-runner'],
       extras_require={
           'docs': ['sphinx', 'sphinx_rtd_theme', 'numpydoc'],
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -31,7 +31,7 @@ def get_version(short=False):
       url='https://github.com/williamjameshandley/anesthetic',
       packages=find_packages(),
       scripts=['scripts/anesthetic'],
-      install_requires=['numpy', 'scipy', 'pandas', 'matplotlib'],
+      install_requires=['numpy', 'scipy', 'pandas<=1.0.5', 'matplotlib'],
```
lukashergt, line 34: You've added the `<=1.0.5` flag to both `requirements*.txt` and to `setup.py`. Should that generally be done? Should we do this for `matplotlib` and `scipy` also?

williamjameshandley: >You've added the <=1.0.5 flag to both requirements*.txt and to setup.py. Should that generally be done? 

There are [ways round the duplication](https://stackoverflow.com/questions/43658870/requirements-txt-vs-setup-py) inherent in having dependencies in both requirements.txt and setup.py, but they're always very fiddly, and just as liable to human error. For now it's simplest/least confusing for users to maintain both, irritating though that is

>Should we do this for matplotlib and scipy also?

Yes : ) Also synced requirements_extras.txt for compleness

williamjameshandley:
Cheers @lukashergt. You should update #114 accordingly so that we can merge that before more major changes occur.

Also just occurred to me how odd the title of this PR would look to somebody not familiar with the python datascience ecosystem.

lukashergt:
> Also just occurred to me how odd the title of this PR would look to somebody not familiar with the python datascience ecosystem.

So true! Hadn't thought of that. Haha, thanks for that!

PR #117, opened 2020-07-29
Added code to compute insertion indices

williamjameshandley:
# Description

This pull request implements some of the code that underpins https://arxiv.org/abs/2006.03371 

@andrewfowlie will likely have an opinion on which other pieces of code from
that paper are sufficiently general to make it into this paper

# Checklist:

### DIFF
--- a/anesthetic/samples.py
+++ b/anesthetic/samples.py
@@ -11,7 +11,7 @@ from anesthetic.plot import (make_1d_axes, make_2d_axes, fastkde_plot_1d,
                              fastkde_contour_plot_2d,
                              kde_contour_plot_2d, hist_plot_2d)
 from anesthetic.read.samplereader import SampleReader
-from anesthetic.utils import (compute_nlive, compute_insertion_indices,
+from anesthetic.utils import (compute_nlive, compute_insertion_indexes,
                               is_int, logsumexp)
 from anesthetic.gui.plot import RunPlotter
 from anesthetic.weighted_pandas import WeightedDataFrame, WeightedSeries
@@ -653,8 +653,8 @@ class NestedSamples(MCMCSamples):
         self.tex['nlive'] = r'$n_{\rm live}$'
         self.beta = self._beta
 
-    def _compute_insertion_indices(self):
-        self['insertion'] = compute_insertion_indices(self.logL.values,
+    def _compute_insertion_indexes(self):
+        self['insertion'] = compute_insertion_indexes(self.logL.values,
                                                       self.logL_birth.values)
 
     _metadata = MCMCSamples._metadata + ['_beta']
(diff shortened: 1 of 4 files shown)


andrewfowlie:
I think this is quite nice. So you give the user access to the insertion indexes, then it's up to them how they do the tests on them? Should we also add a method for a user to perform the test?

williamjameshandley:
I think it would be good to at least provide a function that computes the 'p value' and 'rolling p value' in `anesthetic.utils` as they were determined in `https://arxiv.org/abs/2006.03371`. You know that code best: would you be able to extract the relevant parts, paste it here, and I'll interface it appropriately to `NestedSamples`

andrewfowlie:
```
import numpy as np
from scipy.stats import kstwobign


def p_value(indexes, n_live, iterations=None):
    """
    @returns Kolmogorov-Smirnov test from indexes

    Don't use scipy.stats.kstest as it assumes continuous distributions
    """
    if iterations is not None:
        indexes = indexes[:iterations]

    empirical_pmf = np.histogram(indexes, bins=np.arange(-0.5, n_live + 0.5, 1.), density=True)[0]
    empirical_cmf = np.cumsum(empirical_pmf)
    uniform_cmf = np.arange(1., n_live + 1., 1.) / n_live

    D = abs(empirical_cmf - uniform_cmf).max()
    sample_size = len(indexes)
    K = D * np.sqrt(sample_size)

    r = {}
    r["D"] = D
    r["sample_size"] = sample_size
    r["p_value_asymptotic"] = kstwobign.sf(K)
    return r

def rolling_p_value(indexes, n_live, iterations=None, chunk_k=1):
    """
    @returns Rolling Kolmogorov-Smirnov test from indexes
    """
    if iterations is not None:
        indexes = indexes[:iterations]

    sample_size = len(indexes)
    chunk_size = chunk_k * n_live
    chunks = [indexes[i:i + chunk_size] for i in range(0, len(indexes), chunk_size)]
    tests = [ks_ns(c, n_live) for c in chunks]
    test = min(tests, key=lambda t: t["p_value_asymptotic"])
    index = tests.index(test)
    test["iterations"] = (index * chunk_size, (index + 1) * chunk_size)
    test["n_chunks"] = len(chunks)
    test["p_value_uncorrected"] = test["p_value_asymptotic"]
    test["p_value_asymptotic"] = 1. - (1. - test["p_value_uncorrected"])**test["n_chunks"]
    if test["p_value_asymptotic"] == 0.:
        test["p_value_asymptotic"] = test["p_value_uncorrected"] * test["n_chunks"]
    return test
```

williamjameshandley:
OK, I have added a version of that code as `anesthetic.utils.insertion_p_value`, and some tests to both validate, and show users how it works. Whilst the theory is still evolving (e.g. how to deal with a variable number of live points, how to correct properly for the initial and final live points), I will hold off from making a dedicated `NestedSamples` p-value checking function beyond the `NestedSamples._compute_insertion_indexes`, but I think it is very helpful to provide users with a p-value test that is appropriate for discrete uniform distributions.

williamjameshandley:
If you're happy with this @andrewfowlie, could you approve again, and I will merge once the tests pass.

PR #119, opened 2020-08-18
nlive calculation bug fix

### LINKED ISSUE #83: Possible issue in dlogX calculation

williamjameshandley:
# Description

There is an off-by-one nlive bug that has been lurking for a while. This is most clearly seen in likelihoods with plateau regions, but has also been occuring in the computation of decreasing nlive values for the final set of live points.

In anesthetic, nlive is a column in the table, and therefore 'belongs' to the point corresponding to a row. It is used to compute the volume compression via the probability distribution:
```
P(t_i) = n_i t_i**(n_i -1)
```
`n_i` therefore corresponds to the number of live points for which row i is the lowest likelihood live point. This means for example that the final point at the end of a run should have nlive=1, not 0.

This PR adjusts the tests so that the values are correct, introduces a failing test for a 'wedding cake' likelihood, which is then fixed by subsequent commits

- Fixes #83 as nlive is now never zero

# Checklist:


williamjameshandley:
@lukashergt, this likely has implications for your work concerning re-weighting/slicing of nested sampling chains. It would likely be worth your pushing a PR with a failing test that captures that behaviour, which we can check if merging this provides a fix.

anesthetic/utils.py
```diff
@@ -182,13 +182,13 @@ def compute_nlive(death, birth):
...
-    birth_index = death.searchsorted(birth)
+    birth_index = death.searchsorted(birth, side='right')
     births = pandas.Series(+1, index=birth_index).sort_index()
-    index = np.arange(death.size)
+    index = np.arange(death.size)+1
     deaths = pandas.Series(-1, index=index)
     nlive = pandas.concat([births, deaths]).sort_index()
     nlive = nlive.groupby(nlive.index).sum().cumsum()
-    return nlive.values
+    return nlive.values[:-1]
```
williamjameshandley, line 191: Most of the changes for this PR are adding tests and removing previous workarounds. 

The adjustments to these three lines are the real bug fix.

lukashergt:
> @lukashergt, this likely has implications for your work concerning re-weighting/slicing of nested sampling chains. It would likely be worth your pushing a PR with a failing test that captures that behaviour, which we can check if merging this provides a fix.

I'm not sure I can turn this into a failing test.

What I have done is infer the evidence in two separate ways:
1) by **masking** my `NestedSamples` data frame and recomputing live points:
   ```python
   ns_masked._compute_nlive(ns_masked.logL_birth)
   ```
   which then let's me run `ns_masked.ns_output()` again to get a distribution for my `logZ_new1`

2) by calculating **prior and posterior volumes** from the masked weights and doing 
   ```python
   logZ_new2 = logZ_old.mean() + np.log(V_post) - np.log(V_prior)
   ```

Should I expect that `logZ_new1.mean() == logZ_new2` ?

williamjameshandley:
>Should I expect that logZ_new1.mean() == logZ_new2 ?

Or at least that they're close to within the error margins of ns_output.

Basically it would be good to do this for a standard example (e.g. on one of the ones in tests/example_data).

andrewfowlie:
I'm not that up to speed with the anesthetic codebase, but I have checked that the `logZ` method now gives an acceptable answer on a MultiNest run with plateaus.

bin/wedding_cake.py
```diff
@@ -0,0 +1,62 @@
+import numpy as np
+from scipy.special import logsumexp
+from numpy.testing import assert_array_less
+from anesthetic import NestedSamples
+import matplotlib.pyplot as plt
+
+np.random.seed(0)
+D=4
```
andrewfowlie, line 8: space around equals?

bin/wedding_cake.py
```diff
@@ -0,0 +1,62 @@
...
+    return np.max(abs(x-0.5),axis=-1)
+
+def i(x):
+    return np.floor(D*np.log(2*r(x))/np.log(alpha)) 
+
+def logL(x):
+    ri = alpha**(i(x)/D)/2
+    return - ri**2/2/sigma**2
+
+
```
andrewfowlie, line 27: you may want to put some of this in a main guard, if __name__ == '__main__': ...? then the weddingcake function can potentially be imported for use elsewhere w/o running the computations etc here

also seems like repetition of the test/wedding_cake.py?

andrewfowlie (approved):
loooks good, a couple of very small comments

bin/wedding_cake.py
williamjameshandley: You're right that it's a repetition of the code in test/wedding_cake.py. As the wedding cake example can be generated efficiently on-demand in memory, I've removed the repeated bin/wedding_cake.py code, as well as the text file example. It would probably be good in future versions of anesthetic to integrate more examples of this kind of test

williamjameshandley:
@andrewfowlie if you're happy with the trimming down of the code, could you re-approve the PR so it can be merged?

@lukashergt let me know if you have any problem with/comments on this merge.

williamjameshandley:
as the coverage checks were failing with the minor reorganisation I also added a couple of tests to bring us up to 100% coverage of samples.

andrewfowlie (approved):

lukashergt (changes requested):
I think it would be good to have a version update for this PR (2.0.0b2).

Other than that I'm not sure whether [this](https://github.com/williamjameshandley/anesthetic/pull/120#issuecomment-675592855) indicates that there might be an over-correction, such that there still is an off-by-one error and whether that should be addressed here, or in #120.

williamjameshandley:
> I think it would be good to have a version update for this PR (2.0.0b2) such that we can easier reference these changes later.

Done, and I made 2.0.0-beta.1 an actual beta release with a tag.

> Other than that I'm not sure whether this indicates that there might be an over-correction, such that there still is an off-by-one error and whether that should be addressed here, or in #120.

I plan to look quite carefully now at #120, as it's not impossible that off-by-one errors remain.

As @andrewfowlie is in NZ and likely asleep by now, could @lukashergt re-approve the version bump.

lukashergt (approved):

PR #120, opened 2020-08-18
logzero masking at prior and likelihood level

lukashergt:
This introduces 2 tests:

1) Apply a logzero mask at the prior level (i.e. _discard_ samples completely). 

2) Apply a logzero mask at the likelihood level (i.e. set `logL` to `-inf`).

Both tests take two approaches each and compare those internally. 
The **first** approach applies the mask to the samples directly and 1) discards them or 2) sets the corresponding `logL` to `-inf` and then recomputes `ns_output()`. 
The **second** approach applies the mask to the weights and computes 1) prior and posterior volume change or 2) only posterior volume change to then infer the new log evidence.

This PR was suggested in #119 to see whether changes there affect the comparison here.

# Checklist:

---

# MWE

```python
ns0 = NestedSamples(root="/home/hergtl/Documents/Projects/AnestheticPrj/anesthetic/tests/example_data/pc")
pi0 = ns0.set_beta(0)
NS0 = ns0.ns_output(nsamples=2000)

mask = ((ns0.x0 > -0.2) & (ns0.x2 > 0.2) & (ns0.x4 < 3.5)).to_numpy()


#######################
# volume re-weighting #
#######################
V_prior     = pi0[mask].weights.sum() / pi0.weights.sum()
V_posterior = ns0[mask].weights.sum() / ns0.weights.sum()
logZ_V1 = NS0.logZ.mean() + np.log(V_posterior)
logZ_V2 = NS0.logZ.mean() + np.log(V_posterior) - np.log(V_prior)


####################
# assign logL=-inf #
####################
ns1 = merge_nested_samples((ns0, ))
ns1.logL = np.where(mask, ns1.logL, -np.inf)
ns1.logL_birth = np.where(mask, ns1.logL_birth, -np.inf)
ns1 = merge_nested_samples((ns1, ))
NS1 = ns1.ns_output(nsamples=2000)


###########
# discard #
###########
ns2 = merge_nested_samples((ns0[mask], ))
NS2 = ns2.ns_output(nsamples=2000)


##################
# posterior plot #
##################
fig = plt.figure(figsize=(7, 7))
fig, axes = make_2d_axes(['x0', 'x2', 'x4'], fig=fig)
ns0.plot_2d(axes, c='C0', ec='C0', alpha=0.7, label="0")
ns1.plot_2d(axes, c='C1', ec='C1', alpha=0.7, label="1, assign logL=-inf")
ns2.plot_2d(axes, c='C2', ec='C2', alpha=0.7, label="2, discard")
axes['x0']['x0'].set_xlim(ns0.x0.min(), ns0.x0.max())
axes['x2']['x2'].set_xlim(ns0.x2.min(), ns0.x2.max())
axes['x4']['x4'].set_xlim(ns0.x4.min(), ns0.x4.max())
axes.iloc[-1, 0].legend(bbox_to_anchor=(len(axes), len(axes)), loc='upper left');
fig.savefig("logzero_masking_posterior.png")


#################
# evidence plot #
#################
fig, axes = make_1d_axes(['logZ'], fig=fig)
NS0.plot_1d(axes, label="0")
NS1.plot_1d(axes, label="1, assign logL=-inf")
NS2.plot_1d(axes, label="2, discard")
axes['logZ'].axvline(logZ_V1, c='C1')
axes['logZ'].axvline(logZ_V2, c='C2')
axes['logZ'].set_xticks(np.arange(-8, -1, 1))
axes['logZ'].legend(bbox_to_anchor=(1, 1));
fig.savefig("logzero_masking_evidences.png")
```

### Posterior plot

logzero_masking_posterior

### Evidence plot

The vertical lines come from the volume re-weighting and the posteriors from the masked samples.

Note that for the orange posterior _both_ `logL` and `logL_birth` were set to `-inf`, else we'd get a ValueError. Is that ValueError expected? Or does it hint that there might be a bug?

logzero_masking_evidences

### DIFF
--- a/tests/test_samples.py
+++ b/tests/test_samples.py
@@ -638,3 +638,38 @@ def test_posterior_points():
     ns = NestedSamples(root='./tests/example_data/pc')
     assert_array_equal(ns.posterior_points(), ns.posterior_points())
     assert_array_equal(ns.posterior_points(0.5), ns.posterior_points(0.5))
+
+
+def test_logzero_mask_prior_level():
+    np.random.seed(3)
+    ns0 = NestedSamples(root='./tests/example_data/pc')
+    pi0 = ns0.set_beta(0)
+    NS0 = ns0.ns_output(nsamples=2000)
+    mask = ((ns0.x0 > -0.3) & (ns0.x2 > 0.2) & (ns0.x4 < 3.5)).to_numpy()
+
+    V_prior = pi0[mask].weights.sum() / pi0.weights.sum()
+    V_posterior = ns0[mask].weights.sum() / ns0.weights.sum()
+    logZ_V_mask = NS0.logZ.mean() + np.log(V_posterior) - np.log(V_prior)
+
+    ns1 = merge_nested_samples((ns0[mask],))
+    NS1 = ns1.ns_output(nsamples=2000)
+
+    assert abs(NS1.logZ.mean() - logZ_V_mask) < NS1.logZ.std()
+
+
+def test_logzero_mask_likelihood_level():
+    np.random.seed(3)
+    ns0 = NestedSamples(root='./tests/example_data/pc')
+    NS0 = ns0.ns_output(nsamples=2000)
+    mask = ((ns0.x0 > -0.3) & (ns0.x2 > 0.2) & (ns0.x4 < 3.5)).to_numpy()
+
+    V_posterior = ns0[mask].weights.sum() / ns0.weights.sum()
+    logZ_V_mask = NS0.logZ.mean() + np.log(V_posterior)
+
+    ns1 = merge_nested_samples((ns0,))
+    ns1.logL = np.where(mask, ns1.logL, -np.inf)
+    # ns1.logL_birth = np.where(mask, ns1.logL_birth, -np.inf)
+    ns1 = merge_nested_samples((ns1,))
+    NS1 = ns1.ns_output(nsamples=2000)
+
+    assert abs(NS1.logZ.mean() - logZ_V_mask) < NS1.logZ.std()
--- a/tests/test_samples.py
+++ b/tests/test_samples.py
@@ -669,12 +669,12 @@ def test_logzero_mask_prior_level():
 
     V_prior = pi0[mask].weights.sum() / pi0.weights.sum()
     V_posterior = ns0[mask].weights.sum() / ns0.weights.sum()
-    logZ_V_mask = NS0.logZ.mean() + np.log(V_posterior) - np.log(V_prior)
+    logZ_V = NS0.logZ.mean() + np.log(V_posterior) - np.log(V_prior)
 
     ns1 = merge_nested_samples((ns0[mask],))
     NS1 = ns1.ns_output(nsamples=2000)
 
-    assert abs(NS1.logZ.mean() - logZ_V_mask) < NS1.logZ.std()
+    assert abs(NS1.logZ.mean() - logZ_V) < 1.5 * NS1.logZ.std()
 
 
 def test_logzero_mask_likelihood_level():
@@ -684,12 +684,11 @@ def test_logzero_mask_likelihood_level():
     mask = ((ns0.x0 > -0.3) & (ns0.x2 > 0.2) & (ns0.x4 < 3.5)).to_numpy()
 
     V_posterior = ns0[mask].weights.sum() / ns0.weights.sum()
-    logZ_V_mask = NS0.logZ.mean() + np.log(V_posterior)
+    logZ_V = NS0.logZ.mean() + np.log(V_posterior)
 
-    ns1 = merge_nested_samples((ns0,))
-    ns1.logL = np.where(mask, ns1.logL, -np.inf)
-    # ns1.logL_birth = np.where(mask, ns1.logL_birth, -np.inf)
-    ns1 = merge_nested_samples((ns1,))
+    ns1 = NestedSamples(root='./tests/example_data/pc')
+    ns1.logL = np.where(mask, ns1.logL, -1e30)
+    ns1 = merge_nested_samples((ns1[ns1.logL > ns1.logL_birth],))
     NS1 = ns1.ns_output(nsamples=2000)
 
-    assert abs(NS1.logZ.mean() - logZ_V_mask) < NS1.logZ.std()
+    assert abs(NS1.logZ.mean() - logZ_V) < 1.5 * NS1.logZ.std()


tests/test_samples.py
```diff
@@ -638,3 +638,38 @@ def test_posterior_points():
...
+    mask = ((ns0.x0 > -0.3) & (ns0.x2 > 0.2) & (ns0.x4 < 3.5)).to_numpy()
+
+    V_prior = pi0[mask].weights.sum() / pi0.weights.sum()
+    V_posterior = ns0[mask].weights.sum() / ns0.weights.sum()
+    logZ_V_mask = NS0.logZ.mean() + np.log(V_posterior) - np.log(V_prior)
+
+    ns1 = merge_nested_samples((ns0[mask],))
+    NS1 = ns1.ns_output(nsamples=2000)
+
+    assert abs(NS1.logZ.mean() - logZ_V_mask) < NS1.logZ.std()
```
lukashergt, line 657: This tests completely discards the samples excluded by the mask.

This test actually passes, i.e. the `logZ` from prior and posterior volume reweighting lies within 1 sigma of the mean of the recomputed `ns_output()`.

tests/test_samples.py
```diff
@@ -638,3 +638,38 @@ def test_posterior_points():
...
+    V_posterior = ns0[mask].weights.sum() / ns0.weights.sum()
+    logZ_V_mask = NS0.logZ.mean() + np.log(V_posterior)
+
+    ns1 = merge_nested_samples((ns0,))
+    ns1.logL = np.where(mask, ns1.logL, -np.inf)
+    # ns1.logL_birth = np.where(mask, ns1.logL_birth, -np.inf)
+    ns1 = merge_nested_samples((ns1,))
+    NS1 = ns1.ns_output(nsamples=2000)
+
+    assert abs(NS1.logZ.mean() - logZ_V_mask) < NS1.logZ.std()
```
lukashergt, line 675: This test assigns `-inf` to the `logL` outside the mask. 

In its current form this doesn't even compute the new sample. It throws an off-by-one error on line 672:
```python
ValueError: Length of values (1501) does not match length of index (1500)
```
This off-by-one error is interesting in the light of #119, however, #119 does not seem to fix this, it just changes the lengths to 1500 and 1499.

If I uncomment line 671, then it can actually compute the new samples. However the resulting evidence is quite off from the posterior volume re-weighted one. Should I be setting both `logL` _and_ `logL_birth` to `-inf`?

tests/test_samples.py
lukashergt: This corresponds to the green line in the MWE plots.

tests/test_samples.py
lukashergt: This corresponds to the orange line in the MWE plots.

lukashergt:
Note the different outputs in the following print statements. Is that the expected behaviour?

```python
from anesthetic.samples import compute_nlive

ns0 = NestedSamples(root="anesthetic/tests/example_data/pc")
print(compute_nlive(ns0.logL, ns0.logL_birth).shape)
# > (1500,)

mask = ((ns0.x0 > -0.2) & (ns0.x2 > 0.2) & (ns0.x4 < 3.5)).to_numpy()
ns1 = NestedSamples(root="anesthetic/tests/example_data/pc")
ns1.logL = np.where(mask, ns1.logL, -np.inf)
print(compute_nlive(ns1.logL, ns1.logL_birth).shape)
# > (1499,)
```

williamjameshandley:
Many thanks for these examples @lukashergt , they are thorough and clear.

After having a play around, I think that everything is fixed if you replace your 'assign logL=-inf' procedure with the following
```python
#####################
# assign logL=-1e30 #
#####################
ns1 = ns0.copy()
ns1.logL = np.where(mask, ns1.logL, -1e30)
ns1 = merge_nested_samples((ns1[ns1.logL > ns1.logL_birth],))
NS1 = ns1.ns_output(nsamples=2000)
```
Making this change yields the evidence plot:

logzero_masking_evidences



There are two changes here -- the first it to assign a low, but not zero likelihood (`-1e30`). The second is an extra step where you mask out samples which would not form part of a nested sampling run, namely those where the birth contour is not less than the contour spawned by it. 

What is particularly cool, is that if you change `-1e30` in the above to `-np.inf`, you get `ns2`, i.e. the prior excluded run. Even better, this means that we have actually found the general re-weighting scheme for nested sampling. Imagine you wanted to 'post process' a run with a new likelihood, so that
```
logL(theta) -> logL(theta) + logLnew(theta)
```
This is often performed in cosmology, for example in the [Planck parameters tables](https://wiki.cosmos.esa.int/planck-legacy-archive/images/9/9c/Baseline_params_table_2018_95pc.pdf), anything labeled with 'post' indicates a reweighting of the above form, but applied to MCMC samples. I have been musing on and off for nearly four years now how to do this with nested sampling so that you also get evidences, but it took the edifice of anesthetic to make this so easy we could stumble on it almost by accident. In this case, the correct re-weighting could be, for example
```python
logL_new = -(ns0.x0-0.1)**2/2/0.05**2

#####################
# Importance sample #
#####################
ns3 = ns0.copy()
ns3.logL += logL_new
ns3 = merge_nested_samples((ns3[ns3.logL > ns3.logL_birth],))
NS3 = ns3.ns_output(nsamples=2000)

# Plot
fig, ax = ns0.plot_2d(['x0','x1'])
ns3.plot_2d(ax)
fig.tight_layout()
fig.savefig("importance.png")

# Compute evidence
print(ns0.logZ(), ns3.logZ())
```
importance

In general, reweighting doesn't solve everything, as if there is a big difference between the importance sampled likelihood and the actual one you sampled over in the first place then you won't have enough sampling coverage to get decent posteriors and evidences, but it would definitely be worth doing in a similar context to the planck parameters table, where the reweighting is more minor.

As this is not at all obvious to the average user, the next steps would be to explictly encapsulate this in a 'reweighting' function, either as a utility, or as a member function for NestedSamples (I would be inclined to with the latter). What is your bandwidth @lukashergt for making a PR for this which we can iterate on?

lukashergt:
Very nice!

> Making this change yields the evidence plot:

You've changed the mask for that plot, right? It seems like for that particular mask things line up pretty much bang on. For other masks I've found the lines to be not so perfectly aligned, but definitely not as bad as before. Now, it seems to match at least up to 2 sigma.

williamjameshandley:
>You've changed the mask for that plot, right? 

Ah yes -- forgot to change it back when experimenting. Here is the plot for the original mask
logzero_masking_evidences

We should note that this is a distinct advantage of doing it this way, rather than manually computing log volumes, since we get an error bar (note as well that the example has quite large error bars on account of a small number of live points). In theory the volumes themselves would have an error on them associated with fluctuations in the weights (which could be done by simulating them, although this wouldn't be that easy with anesthetic at the moment). The current block in #81 causing those tests to fail could provide a solution to both at once, by exposing an explict weight simulation method (rather than the rather cryptic dlogX)

>Is, .copy() really creating a deep copy? Or do we risk modifying ns0 alongside ns1 this way?

I think we're probably fine, since our arrays don't contain complicated objects that would need references deep copying, but this is the kind of thing that one should test for in any implementation of a reweighting scheme.

williamjameshandley:
Now that the checks have passed, we can either merge this now, or use this PR for the creation of a reweighting scheme.

williamjameshandley (approved):

lukashergt:
> Now that the checks have passed, we can either merge this now, or use this PR for the creation of a reweighting scheme.

I'll wrap this one up for now.

PR #121, opened 2020-08-18
skipna functionality for WeightedDataFrame means and (co)variances

williamjameshandley:
# Description

This is a slightly controversial PR for 2.0.0. Here we replace the 'skip nans that have zero weight' when computing means and variance, with 'skip all nans, unless told not to'.  We opted for the old choice in PR #112, on account of safety. However, I can now see two good reasons for making this backward incompatible change in light of the upcoming major version update.

1. It is more consistent with pandas, and therefore less surprising.
2. There is a well-defined use case in the case of nested sampling where the number of parameters `N` can vary. In this case, the correct thing to do is to record parameters that are not active as 'missing data', which is what pandas means by nans. Conditional means and variances can then be computed correctly.

Note that this will not break code that previously relied on the 'skip nans that have zero weight' functionality.

# Checklist:

### DIFF
--- a/anesthetic/weighted_pandas.py
+++ b/anesthetic/weighted_pandas.py
@@ -2,6 +2,7 @@
 
 import numpy as np
 import pandas
+from numpy.ma import masked_array
 from anesthetic.utils import (compress_weights, channel_capacity, quantile,
                               temporary_seed)
 
@@ -11,7 +12,7 @@ class _WeightedObject(object):
     def weights(self):
         """Sample weights."""
         if self.index.nlevels == 1:
-            return pandas.Series(index=self.index, data=1.)
+            return np.ones_like(self.index)
         else:
             return self.index.get_level_values('weights').to_numpy()
 
@@ -28,9 +29,9 @@ class _WeightedObject(object):
         with temporary_seed(seed):
             return np.random.rand(len(self))
 
-    def std(self):
+    def std(self, skipna=True):
         """Weighted standard deviation of the sampled distribution."""
-        return np.sqrt(self.var())
+        return np.sqrt(self.var(skipna=skipna))
 
     def median(self):
         """Weighted median of the sampled distribution."""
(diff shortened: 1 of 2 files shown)


lukashergt:
Where does the `skipna` name come from? Is that a pandas thing? I would prefer `skipnan` or `skipnans` or `ignore_nan` otherwise.

williamjameshandley:
>Where does the skipna name come from? Is that a pandas thing? I would prefer skipnan or skipnans or ignore_nan otherwise.

Sadly this is the [pandas notation](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.mean.html), so in the interests of 'least surprise', we should probably stick with that.

In the latest push I have also provided an axis kwarg to mean, var and quantile, in line with the pandas signature, which allows you to compute these stats across columns The pandas signature also includes level, and numeric_only, which I'm not sure there is a strong use case (axis was marginal in my view, but I can think of a couple of settings where it might be useful when combined with slicing of columns)

lukashergt:
> Sadly this is the [pandas notation](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.mean.html), so in the interests of 'least surprise', we should probably stick with that.

Hmm, I see, NA is their generalisation of NaN to also include NaT...

> which allows you to compute these stats across columns

Ha, now we can compute the average of all cosmo parameters.^^

williamjameshandley:
Looping in @tmcaloone who has been working on transdimensional nested sampling with me (being able to vary the number of parameters dynamically), so is a good candidate to act as a reviewer

williamjameshandley:
>Ha, now we can compute the average of all cosmo parameters.^^

: ) You're right that for cosmology there is not a strong use case, but there might be if you were doing a [free form reconstruction](https://arxiv.org/abs/1908.00906), and had several parameters of the same dimension (e.g. y-locations of knots). You then might be interested in `samples[['y0','y1','y2','y3']].mean(axis=1)`, adding that as a column and seeing how that evolves with the gui (tenuous, but at least kind of cosmological). There is likely other uses, and WeightedDataFrame is lower-level (and therefore more general) than MCMCSamples or NestedSamples

PR #122, opened 2020-08-20
add importance_sample method to NestedSamples and MCMCSamples

lukashergt:
As suggested in #120, this PR implements a method for importance re-weighting the nested samples, which allows e.g. discarding/penalising samples as done in #120, but also allows adding to `logL`, e.g. with data from a different experiment.

Some things to keep in mind:
* > We should note that this is a distinct advantage of doing it this way, rather than manually computing log volumes, since we get an error bar.
* > In general, reweighting doesn't solve everything, as if there is a big difference between the importance sampled likelihood and the actual one you sampled over in the first place then you won't have enough sampling coverage to get decent posteriors and evidences, but it would definitely be worth doing in a similar context to the planck parameters table, where the reweighting is more minor.

# Checklist:


anesthetic/samples.py
```diff
@@ -611,6 +611,38 @@ def dlogX(self, nsamples=None):
         else:
             return WeightedDataFrame(dlogX, self.index, weights=self.weights)
 
+    def importance_reweighting(self, logL_new, replace=True, add=False):
```
lukashergt, line 614: I chose to call the method `importance-reweighting`, because I fell like just `reweighting` insinuates that the `weights` are being accessed/changed, when really you are supposed to provide a new `logL`.

anesthetic/samples.py
```diff
@@ -611,6 +611,38 @@ def dlogX(self, nsamples=None):
...
+        Parameters
+        ----------
+        logL_new: np.array
+            New log-likelihood values. Should have the same shape as `logL`.
+
+        replace: bool
+            Replace the current `logL` with the new `logL_new`.
+
+        add: bool
+            Add the new `logL_new` to the current `logL`.
```
lukashergt, line 626: I've included these two bool args to allow the user to decide whether to completely replace the `logL` or whether to add to it. 
Happy with that? 
Would there be a crisper way of implementing this that is easy to understand?

anesthetic/samples.py
```diff
@@ -611,6 +611,38 @@ def dlogX(self, nsamples=None):
...
+        if not replace ^ add:
+            raise ValueError("One and only one of args 'replace' or 'add' "
+                             "should be set to True.")
+        elif replace:
+            samples.logL = logL_new
+        elif add:
+            samples.logL += logL_new
+        samples = merge_nested_samples(
+            (samples[samples.logL > samples.logL_birth], )
+        )
```
williamjameshandley, line 643: This use of `merge_nested_samples` on a single array to resort and recompute the live points is a really cool idiom, which is spoilt by the name of the function. Is it worth considering a re-name here to reflect this additional use (at the very least, the docstring of `merge_nested_samples` should be updated to 'one or more', and perhaps have a sentence saying that this should also be used to recompute a masked run. Alternatively, one could add a `recalculate_nested_sampling` (or equivalently named) method to NestedSamples, which is then used by merge_nested_samples.

anesthetic/samples.py
andrewfowlie: What happens when they are both true or both false? I'm not sure but I'd probably favor other options:

- a string might be clearer, something like `importance_reweighting(..., action='add')`` or `importance_reweighting(..., action='replace')'
- two separate different methods with very clear names `loglike_replace(loglike)` and `loglike_add(loglike)`

anesthetic/samples.py
andrewfowlie: see comment below about function signature

williamjameshandley (changes requested):

anesthetic/samples.py
lukashergt: > What happens when they are both true or both false?

In that case I am raising an exception that let's the user know that one and only one of those two should be True.

williamjameshandley:
postscript -- mine and Andrew's comments overlap (apologies, I was mid-review and did not see andrew).

anesthetic/samples.py
lukashergt: > * two separate different methods with very clear names loglike_replace(loglike) and loglike_add(loglike)

I think it would be nicer to keep this in one place.

> * just a single bool, since in the end it's a choice between two things

I went with the two args to be more descriptive and couldn't come up with a good name that would merge this into one.

> * a string might be clearer, something like importance_reweighting(..., action='add') or importance_reweighting(..., action='replace')

I like this option. This would also be easily expandable to other actions.

anesthetic/samples.py
lukashergt: Good ideas, happy to integrate this into this PR, but ambivalent as to which route to take.

williamjameshandley: I would go with the `NestedSamples.recalculate` option.

lukashergt: Hmm, on closer inspection, the reason I like using `merge_nested_samples` in these situations is it's call to `pandas.concat` which returns a new pandas dataframe which subsequently gets turned into a `NestedSamples` object, thus making a deep copy.

A `NestedSamples.recalculate` method would essentially be an exact copy of `merge_nested_samples` and would not be worth calling from `merge_nested_samples`, so I'm going with a docstring update after all.

anesthetic/samples.py
lukashergt: Changed accordingly in [bb7a013](https://github.com/williamjameshandley/anesthetic/pull/122/commits/bb7a013b6ecf386f9e40e96849f30ef9f2e6ad34).

Any further suggestions which other actions might be useful?

anesthetic/samples.py
lukashergt: > Would `importance_sample` resolve this?

Renamed accordingly in [bfd5183](https://github.com/williamjameshandley/anesthetic/pull/122/commits/bfd51830f904a35ac81984698231f2c620f68c92).

anesthetic/samples.py
lukashergt: done in [fc52995](https://github.com/williamjameshandley/anesthetic/pull/122/commits/fc529951dc7550cc5bf2fdc3de54d9eeb908ce03).

anesthetic/samples.py
lukashergt: added `action='mask'` in [45562a0](https://github.com/williamjameshandley/anesthetic/pull/122/commits/45562a086fd397d029c87e7bc467d5f06ffd8d7d).

tests/test_samples.py
```diff
@@ -648,6 +648,45 @@ def test_posterior_points():
...
+    ns1 = ns0.importance_sample(np.zeros_like(ns0.logL), action='add')
+    assert_array_equal(ns0.logL, ns1.logL)
+    assert_array_equal(ns0.logL_birth, ns1.logL_birth)
+    assert_array_equal(ns0.weights, ns1.weights)
+
+    mask = ((ns0.x0 > -0.1) & (ns0.x2 > 0.3) & (ns0.x4 < 3)).to_numpy()
+    ns1 = merge_nested_samples((ns0[mask], ))
+    logL_new = np.where(mask, ns0.logL, -np.inf)
+    ns2 = ns0.importance_sample(logL_new=logL_new)
+    assert_array_equal(ns1.logL, ns2.logL)
```
williamjameshandley, line 672: If I've understood this test correctly, aren't ns1 and ns2 identical? If so, all of the tests below can be reduced to a single
`assert_array_equal(ns1, ns2)`. This would also make the checks against logZ and D trivially true.

It would be good to instead compare `NS1` to an `NS0` which then checks against the volume factors that you created in #120.

anesthetic/samples.py
```diff
@@ -611,6 +611,43 @@ def dlogX(self, nsamples=None):
         else:
             return WeightedDataFrame(dlogX, self.index, weights=self.weights)
 
+    def importance_sample(self, logL_new, action='replace'):
```
williamjameshandley, line 614: This is much better now with the `action` kwarg.

I'm willing to be argued down on this, but I think that it would be less surprising to have `action='add'` as the default, since this is what [GetDist](https://getdist.readthedocs.io/en/latest/mcsamples.html#getdist.mcsamples.MCSamples.reweightAddingLogLikes) does.

I know that your original use case was a replacement, so it might be helpful to note that replacing with `-1e30` will be the same in most cases to double precision as adding it (and will regardless have an identical effect).

williamjameshandley (changes requested):

williamjameshandley:
@pstoecker, given your work on gambit related importance weighting, you will likely be interested in this PR when it is merged later today, as well as the [conversation](https://github.com/williamjameshandley/anesthetic/pull/120#issuecomment-676537096) in #120.

anesthetic/samples.py

williamjameshandley (changes requested):
Sorry to be a pain in requesting another change, but something I just realised when adjusting the title is that a key thing that this PR is missing is an equivalent method for `MCMCSamples`, which would just reweight the logL column. `NestedSamples` should then `super()` call the reweighting portion, and then recompute the nlive with `merge_nested_samples`

tests/test_samples.py
lukashergt: > If I've understood this test correctly, aren't ns1 and ns2 identical?

I thought `importance_sample` with `-inf` would leave a different shape than direct masking, but no, they indeed are identical. Consequently, the structure of the rest of the test in relation to the volume factor tests, does not work the way I envisioned.

I fixed the tests accordingly.

anesthetic/samples.py
lukashergt: > OK, we'll go with recommending the 'merge a single nested sampling run' idiom for now. You should still update the docstring so that it reads "Merge one or more nested sampling runs."

I did that (see preceding post with commit link)... Or do you mean somewhere else?

williamjameshandley: Top of the docstring still reads "Merge two or more nested sampling runs." on both my clone and on github.

tests/test_samples.py
williamjameshandley: This was mentioned [here](https://github.com/williamjameshandley/anesthetic/pull/120#issuecomment-676537096) in #120, but I agree that this is unexpected. I'll raise a separate discussion issue now for how to resolve this.

anesthetic/samples.py
lukashergt: Apologies, I was looking at the sentence below the `runs` parameter docstring, not the title docstring.

Fixed accordingly.

tests/test_samples.py
lukashergt: Yeah, I read that at first just as "produces the same result" and did not appreciate that it's actually identical.

williamjameshandley (approved):
Excellent work.

One further optional addition is to add an inplace=False optional kwarg to `importance_sample`, in the same vein as `NestedSamples.set_beta`, and generally quite common in pandas for this kind of large-scale operation.. This would be implemented simply as
```python
def importance_sample(self, logL_new, action='add', inplace=False):
￼....
     if inplace:
        self = samples
    else:
        return samples
```

lukashergt:
> One further optional addition is to add an inplace=False optional kwarg to `importance_sample`, in the same vein as `NestedSamples.set_beta`, and generally quite common in pandas for this kind of large-scale operation.. This would be implemented simply as
> 
> ```python
> def importance_sample(self, logL_new, action='add', inplace=False):
> ￼....
>      if inplace:
>         self = samples
>     else:
>         return samples
> ```

The situation here is a tiny bit more complicated than for `set_beta`, as `self` might (or probably will) change shape here. I gave it a try and it didn't work as easily as anticipated and now I'm wondering whether this is really needed? 

Typical situations where `inplace` is used is where the structure of the object stays intact and only some elements are modified, e.g. to get a perfomance gain. This is not the case, here, so I'd vote for leaving this option out.

lukashergt:
> Sorry to be a pain in requesting another change, but something I just realised when adjusting the title is that a key thing that this PR is missing is an equivalent method for `MCMCSamples`, which would just reweight the logL column. `NestedSamples` should then `super()` call the reweighting portion, and then recompute the nlive with `merge_nested_samples`

This thing made me wish for docstring inheritance (#22, #24) again... In the current version the docstrings for the `MCMCSamples` and the `NestedSamples` method are almost identical...

lukashergt:
Many thanks to @andrewfowlie and @williamjameshandley for your helpful input!

PR #126, opened 2020-08-20
importance_sample with inplace=False

williamjameshandley:
# Description

Quick follow-up to #122 implementing an inplace argument for importance_sample. The key was to use the self.__init__ function, rather than assigning to a local copy self.

(NB messed up the PR #124 with a stale local branch -- ignore.

# Checklist:

### DIFF
--- a/anesthetic/samples.py
+++ b/anesthetic/samples.py
@@ -341,7 +341,7 @@ class MCMCSamples(WeightedDataFrame):
 
         return fig, axes
 
-    def importance_sample(self, logL_new, action='add'):
+    def importance_sample(self, logL_new, action='add', inplace=False):
         """Perform importance re-weighting on the log-likelihood.
 
         Parameters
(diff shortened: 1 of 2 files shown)


anesthetic/samples.py
```diff
@@ -645,35 +653,43 @@ def dlogX(self, nsamples=None):
...
         samples = super(NestedSamples, samples).importance_sample(
-            logL_new=logL_new, action=action
+            logL_new, action=action
         )
         samples = merge_nested_samples(
             (samples[samples.logL > samples.logL_birth], )
         )
-        return samples
+        if inplace:
+            self.__init__(samples)
```
lukashergt, line 690: This line is going far back its dynasty to its (great?)grandparents, and I believe it loses some of its `NestedSamples` attributes in the process.

This is partly due to `merge_nested_samples` not taking care of all attributes, and as such already applies to non-inplace samples. But with inplace, we loose a few more (at least in `.tex`).

How do we want `.tex`, `.limits`, `.label`, `.root` to behave under importance re-weighting?

williamjameshandley: Nice spot! This was not expected, and I believe a property of how pandas interprets __init__. I have a replaced with a more verbose method, and with a test that checks all of these are passed on to the object, and that we have a copies of the relevant mutable metadata.

PR #130, opened 2020-10-15
Added prior and prior_points shortcuts

### LINKED ISSUE #118: prior_samples shortcut
In a similar vein to `NestedSamples.posterior_samples()`, there should be a shortcut for `NestedSamples.prior_samples()` (basically posterior_samples, but with beta=0). This would be a good PR for a newcomer to implement to understand how to contribute to anesthetic.

EthanCarragher:
- Added `NestedSamples.prior_points()` as alias for `NestedSamples.posterior_points(0)`.
- Added `NestedSamples.convert_to_prior()` as alias for `NestedSamples.set_beta(0)`. (Not sure if `NestedSamples.prior()` would be a better name).

Fixes #118

### DIFF
--- a/anesthetic/samples.py
+++ b/anesthetic/samples.py
@@ -523,6 +523,10 @@ class NestedSamples(MCMCSamples):
             data.beta = beta
             return data
 
+    def prior(self, inplace=False):
+        """Re-weight samples at infinite temperature to get prior samples."""
+        return self.set_beta(beta=0, inplace=inplace)
+
     def ns_output(self, nsamples=200):
         """Compute Bayesian global quantities.
 
@@ -629,6 +633,10 @@ class NestedSamples(MCMCSamples):
     def posterior_points(self, beta=1):
         """Get equally weighted posterior points at temperature beta."""
         return self.set_beta(beta).compress(-1)
+
+    def prior_points(self, params=None):
+        """Get equally weighted prior points."""
+        return self.posterior_points(beta=0)
 
     def gui(self, params=None):
         """Construct a graphical user interface for viewing samples."""
--- a/tests/test_samples.py
+++ b/tests/test_samples.py
@@ -10,6 +10,7 @@ from anesthetic import MCMCSamples, NestedSamples, make_1d_axes, make_2d_axes
 from anesthetic.samples import merge_nested_samples
 from numpy.testing import (assert_array_equal, assert_array_almost_equal,
                            assert_array_less)
+from pandas.testing import assert_frame_equal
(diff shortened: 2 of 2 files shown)


anesthetic/samples.py
```diff
@@ -630,6 +634,10 @@ def posterior_points(self, beta=1):
         """Get equally weighted posterior points at temperature beta."""
         return self.set_beta(beta).compress(-1)
 
+    def prior_points(self, params=None):
+        """Get prior points, i.e. posterior points at infinite temperature."""
```
williamjameshandley, line 638: I think the docstring here should probably also mention equally weighted (and avoid i.e. as this is an implementation detail that will probably not be of use to users who only interact with the API), e.g. `"""Get equally weighted prior points."""`

anesthetic/samples.py
```diff
@@ -523,6 +523,10 @@ def set_beta(self, beta, inplace=False):
             data.beta = beta
             return data
 
+    def convert_to_prior(self, inplace=False):
```
williamjameshandley, line 526: The name of this function is a bit of a mouthful. Would there be anything wrong with renaming it to just `prior`, so that e.g. `samples.prior()` returned a prior-weighted set of samples?

williamjameshandley (changes requested):
Many thanks @EthanCarragher . These are a very nice addition.

At the moment the continuous integration is flagging a failure because the code you have added is untested. Could you add a test or two of the new functions in `tests/test_samples.py`?

You will find the helper script `bin/run_tests` helpful for checking that your new tests are working correctly.

anesthetic/samples.py
EthanCarragher: I avoided this name because of the discussion in #118, but there is nothing wrong with it. I have renamed it now.

EthanCarragher:
Okay, I have added some tests. Are these sufficient?

williamjameshandley (approved):
Excellent. Many thanks for this @EthanCarragher. Please squash and merge.

EthanCarragher:
Hmm... I didn't see any option to squash and merge on this page, so I squashed the commits from the command line and now the build isn't successful. I'm not sure what the problem is here.

williamjameshandley (approved):

williamjameshandley:
Travis weirdness. I restarted the failing job and that seems to have fixed it. Is there a big green button below the checks that says 'merge', which you could possibly choose an option to say 'squash and merge'?

williamjameshandley:
Many thanks @EthanCarragher

PR #136, opened 2021-01-21
Function to merge data sets with weights

### LINKED ISSUE #135: Merging models with evidence weighting
Does anesthetic have a function to merge datasets (`NestedSamples`) _with weights per dataset_?

When considering multiple models for an inference, such as the [FlexKnot reionization model by Millea&Bouchet](https://arxiv.org/pdf/1804.08476.pdf), one want's to marginalize over different models, i.e. average the constraints weighted with the model's evidence. [fgivenx](https://fgivenx.readthedocs.io/en/latest/fgivenx.html#fgivenx.drivers.plot_contours) seems to support such options, does `anesthetic` have a similar function?

A function that merges multiple sets of `NestedSamples`, using the evidence Z (or provided weighting), or a way to directly plot posteriors using multiple sets of samples with weights.

**Describe alternatives you've considered**
One could change the weights manually and use the `merge` function to merge the samples. This is straightforward after converting to MCMCSamples, but I'm not sure about the implications for NestedSamples.

**Additional Context**
This is not urgent or necessary for my work now (I can convert to MCMCSamples and do it manually) but since fgivenx does have the functionality it might be easy (?) to implement and nice to have here :)

Stefan-Heimersheim:
# Description

Adds a function `merge_samples_weighted` that takes two runs and merges them such that
P(x|combined) = weight1 P(x|run1) + weight2 P(x|run2).
The weights can be given or the Bayesian evidences of `NestedSamples`.

Fixes #135 

# Checklist:

# Remaining

* Should the sample weights be normalized in some way?

* ~I'm currently checking if the tests work (some seem to fail with pytest) and I'll try out the functionality in my own work to see if I notice and errors or inconvenient things (e.g. better name or call signature).~


williamjameshandley:
Many thanks @Stefan-Heimersheim, this is already a very polished PR. Unfortunately it looks like the travis CI has been broken since November, since under their new [pricing plan](https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing) we've run out of credits. This may mean that the failure is due to e.g. an update in pandas rather than your code changes specifically. I'm going to have to migrate from travis to GitHub actions over the weekend, and then I'll review the changes here. My apologies for the delay.

Stefan-Heimersheim:
> This may mean that the failure is due to e.g. an update in pandas rather than your code changes specifically. I'm going to have to migrate from travis to GitHub actions over the weekend, and then I'll review the changes here. My apologies for the delay.

I'm running pytest locally now, there seem to be actual test failures so I'm going to fix that later. Surprisingly `test_convert` also fails, I'll look into that later.

Stefan-Heimersheim:
The test failed due to a typo in the test function, that's fixed with the last commit. The results of `python -m pytest` now look like those of the master branch. I think I caused some warnings in my test code due to using the example files from other tests, but it doesn't look like a problem.
```
WARNING  root:mcsamples.py:552 outlier fraction 0.00014285714285714287
```
Summary:
```
============================== short test summary info ===============================
FAILED tests/test_convert.py::test_to_getdist - AssertionError: 
=========== 1 failed, 93 passed, 1 xfailed, 8 warnings in 97.19s (0:01:37) ===========
```
I also have been the function in my code and didn't notice any obvious problems.

Stefan-Heimersheim:
I noticed 3 issues when playing around with the functionality this evening:

1. When using it multiple times like this, the weights are set to `NaN` after the 2nd use.
```python
A = NestedSamples(root='./tests/example_data/pc')
B = merge_samples_weighted([A])
B = merge_samples_weighted([A])
B
```
I suspect there are reference-shenanigans going on, this fixes it:
```python
mcmc_samples = copy.deepcopy([MCMCSamples(s) for s in samples])
```
I'll make a comment to add `deepcopy` to this line and the `logZ()` line.

2. When calling the merge function with a (single) sample instead of a list of samples the error message `ValueError: DataFrame constructor not properly called!` is not helpful.
I'll add an assertion to check this (somewith with `isinstance` `collections.Sequence`).

3. Calling the function with a single array `B = merge_samples_weighted([A])` changes the weights (B.weights != A.weights) if `A.weights.sum != 1`. I'm not sure if this is un-intuitive or good.
In general, the current code always causes `B.weights.sum()==1` (approx), which is different from the `max(B.weights)==1` convention.

williamjameshandley (changes requested):
Hi @Stefan-Heimersheim,

This is a great looking PR, in particular that you already have new tests which add the relevant coverage so that all checks pass.

In response to your comments

1. I agree that there should be a deep copy where you suggest to fix the (sadly usual) pandas-related gotcha
2. I think that in this case it probably is worth checking that the set of samples is indeed iterable
3. I would vote for staying with the anesthetic convention of leaving the weights normalised so that the maximum is 1. Since this will be a deep copy, I don't think this is unintuitive. If we do vote for changing the convention so that the sum is to 1 (which has its own merits), as is discussed in #90, then we should make sure to update this portion of the code accordingly. (this will now be flagged in that issue as it is mentioned here).

I think the main thing which will confuse users is to if they pass a list which includes an `MCMCSamples` object, which then fails when it tries to compute the logZ. You should put a `try...except` block to catch the `AttributeError` and convert it into a value error indicating that if `weights is None` then you can't pass `MCMCSamples`

You should update the PR as you suggest by pushing to your forked branch `weighted_merge`.

williamjameshandley:
Could you also bump the version number in README.rst to 2.0.0-beta.5?

Stefan-Heimersheim:
1. Done

2. I checked for `Sequence` because the dataframe is already `Iterable`, it now raises `TypeError: samples must be a Sequence (list of samples).`

3. I added `new_samples.weights /= new_samples.weights.max()` at the end.

4. Passing `MCMCSamples` without `weights` raises `ValueError: If samples includes MCMCSamples then weights must be given.`

5. Bumped README

williamjameshandley (changes requested):
Excellent. Could you add some tests to check that:
a) the appropriate Exception is raised when a sequence is not passed
b) the appropriate Exception is raised when an MCMC samples is passed without eror
c) The function fails appropriately when len(weights) != len(samples)

It would be more consistent with the rest of the function if the assertion associated with (c) were made into an explicit exception.

After that I think you'll be good to merge

Stefan-Heimersheim:
Thanks for the helpful feedback, I implemented those checks and replaced the assertion by a ValueError

williamjameshandley (approved):

williamjameshandley:
Excellent -- many thanks @Stefan-Heimersheim. Please press squash and merge

Issue #134, opened 2021-01-13
Update project requirements


yallup:
**Describe the bug**
Calls to `anesthetic.plot.scatter_plot_2d` error with messages along the lines of:

`
FAILED tests/test_samples.py::test_plot_2d_types - AttributeError: type object 'Line2D' has no attribute 'items'
`

When running tests with mpl 3.1.3 (meeting current requirement of mpl>=3.1.2)

**To Reproduce**
Run test suite with mpl=3.1.3

**Expected behavior**
Update requirements file. It looks like the cbook.normalize_kwargs function changed since matplotlib>=3.3.0, so this should be sufficient but I only checked with 3.3.3 (latest)

**Additional context**

lukashergt:
Thanks for flagging this up, @yallup, this will be updated in #138.

PR #127, opened 2020-08-22
Missing weighted functions added to WeightedPandas

williamjameshandley:
# Description

There are a few statistical functions in WeightedSeries and WeightedDataFrame which still use the unweighted form of Series and DataFrame:
- [skew](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.skew.html)
- [kurtosis](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.kurtosis.html)
- [sem](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.sem.html)
- [mad](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.mad.html)
- [sample](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.sample.html)
- [corr](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.corr.html)
- [corrwith](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.corr.html)

I think with this PR, which adds weighted versions of the above, we have now overridden all of the statistical functions which need weights, but would appreciate a second set of eyes looking through [the list of methods](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.html) in case I have missed any.

# Checklist:

- [X] I have performed a self-review of my own code
- [X] My code is PEP8 compliant (`flake8 anesthetic tests`)
- [X] My code contains compliant docstrings (`pydocstyle --convention=numpy anesthetic`)
- [X] New and existing unit tests pass locally with my changes (`python -m pytest`)
- [X] I have added tests that prove my fix is effective or that my feature works

PR #138, opened 2021-01-23
Create CI.yml

williamjameshandley:
Trialling github actions for continuous integration

As per the discussion in #136, Travis CI is no longer building on account of an upper limit on the number of credits this account can use.

This PR migrates the test suite to GitHub actions.

These seem fast and more complete than the existing suite (allowing for windows tests as well as osx and linux).


.github/workflows/python-package.yml
```diff
@@ -0,0 +1,39 @@
+# This workflow will install Python dependencies, run tests and lint with a variety of Python versions
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
+
+name: Python package
```
lukashergt, line 4: What is the subtle difference between `CI.yml` and `python-package.yml`? Why do we need both? Is `python-package.yml` even executed?

.github/workflows/python-package.yml
```diff
@@ -0,0 +1,39 @@
...
+  pull_request:
+    branches: [ $default-branch ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.7, 3.8, 3.9]
```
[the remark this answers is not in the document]
williamjameshandley: see below

.github/workflows/python-package.yml
williamjameshandley: No, this was added in error!

williamjameshandley:
The main things to discuss are:
a) CI.yml doesn't have montepython checks, but since these have been (allowably) failing for many months, I'm not sure whether this instability is something we should be including (and it would add extra complexity)
b) The build matrix is quite large, although the impressive speed of github actions means that this isn't as much of a problem as I was thinking it would be. Currently we do 4x3x3 = 36 runs of pytest (3.6,3.7,3.8,3.9)x(linux, osx, windows)x(simple, extras, conda). To reduce it, we could only run one python version without extras on osx and windows), which would make 18 runs, so more than twice as fast (as there are fewer osx builds)
@lukashergt, as a regular PR submitter, how valuable would a less complete but faster test be?

lukashergt:
Shall we update the matplotlib requirements in this PR as remarked in #134?

lukashergt:
> The main things to discuss are:
> a) CI.yml doesn't have montepython checks, but since these have been (allowably) failing for many months, I'm not sure whether this instability is something we should be including (and it would add extra complexity)

Yes, it might be easiest to drop the MontePython checks. Those were only for MontePython MCMC chains anyhow, so doesn't affect MontePython+PolyChord runs.

> b) The build matrix is quite large, although the impressive speed of github actions means that this isn't as much of a problem as I was thinking it would be. Currently we do 4x3x3 = 36 runs of pytest (3.6,3.7,3.8,3.9)x(linux, osx, windows)x(simple, extras, conda). To reduce it, we could only run one python version without extras on osx and windows), which would make 18 runs, so more than twice as fast (as there are fewer osx builds)
> @lukashergt, as a regular PR submitter, how valuable would a less complete but faster test be?

Hmm, not too fussed either way. Reducing might also be worth considering in terms of a limited free github computing time. But I'd say that extras should be checked on all systems, but maybe no need to check all python versions on all systems.

williamjameshandley:
>Shall we update the matplotlib requirements in this PR as remarked in #134?

Done, so this now resolves #134. Also updated setup.py so that it reads the requirements from the requirements.txt to avoid duplication.

setup.py
```diff
@@ -31,7 +31,7 @@ def get_version(short=False):
       url='https://github.com/williamjameshandley/anesthetic',
       packages=find_packages(),
       scripts=['scripts/anesthetic'],
-      install_requires=['numpy', 'scipy>=1.2.0', 'pandas', 'matplotlib>=3.1.2'],
+      install_requires=open('requirements.txt').read().splitlines(),
```
lukashergt, line 34: Do we need a similar `install_requires` for `requirements_extras.txt`?

williamjameshandley: It depends whether or not we want this as a dict (as it is at the moment), so that you can install extras individually as necessary, or whether we want it more automated and just have it as a (filtered) list. I would vote for keeping it as-is at the moment.

williamjameshandley:
>Yes, it might be easiest to drop the MontePython checks. Those were only for MontePython MCMC chains anyhow, so doesn't affect MontePython+PolyChord runs.

OK, let's drop MontePython testing for now then.

>Hmm, not too fussed either way. Reducing might also be worth considering in terms of a limited free github computing time. But I'd say that extras should be checked on all systems, but maybe no need to check all python versions on all systems.

I believe that as a public repository, we have unlimited github computing time (this was part of my motivation for the non-trivial amount of work to migrate from Travis+circle to just github actions). See [here](https://github.community/t/for-public-repositories-is-there-a-monthly-limit-on-minutes/129017) and [here](https://docs.github.com/en/github/setting-up-and-managing-billing-and-payments-on-github/about-billing-for-github-actions).

lukashergt:
> I believe that as a public repository, we have unlimited github computing time (this was part of my motivation for the non-trivial amount of work to migrate from Travis+circle to just github actions). See [here](https://github.community/t/for-public-repositories-is-there-a-monthly-limit-on-minutes/129017) and [here](https://docs.github.com/en/github/setting-up-and-managing-billing-and-payments-on-github/about-billing-for-github-actions).

Ah, that's nice. In that case might as well go for the full grid and make sure everything is covered. With ~2min per check and ~10min total that doesn't seem too bad.

williamjameshandley:
Agreed. It's pretty impressive how much faster GitHub actions are than Travis in this instance. Unless you have any other suggestions, could you re-approve and we'll start testing it on some 'real' PRs like #127 and #136

lukashergt (approved):

PR #139, opened 2021-01-24
fix `MCMCSamples.importance_sample`

### LINKED ISSUE #133: MCMCSamples.importance_sample is missing weight adjustment
**Describe the bug**
`MCMCSamples.importance_sample` does not adjust the weights at all (just the logL) which leads to importance sampling doing nothing.

**To Reproduce**
```
chain = MCMCSamples(orig_chain)
chain.tex = orig_chain.tex

fig, axes = chain.plot_1d("Omega_m", label='Original', color='blue')

logL_new = scipy.stats.norm.logpdf(chain.Omega_m, loc=0.3, scale=0.01)
reweighted = chain.importance_sample(logL_new, action='add')

reweighted.plot_1d(axes, label='Reweighted', color='red', linestyle='dashed')

if np.all(reweighted.weights == chain.weights):
    print("Weights equal")


plt.legend()
plt.show()
```
See #132 [here](https://github.com/williamjameshandley/anesthetic/issues/132#issuecomment-757992829) for plots.

**Expected behavior**
The reweighted chain should have different weights, the posterior should look different

**Screenshots**
Figure_1

**Additional context**
`anesthetic/samples.py`, line 345. There is no code to adjust the weights:
```
    def importance_sample(self, logL_new, action='add', inplace=False):
        """Perform importance re-weighting on the log-likelihood.

        Parameters
        ----------
        logL_new: np.array
            New log-likelihood values. Should have the same shape as `logL`.

        action: str, optional
            Can be any of {'add', 'replace', 'mask'}.
                * add: Add the new `logL_new` to the current `logL`.
                * replace: Replace the current `logL` with the new `logL_new`.
                * mask: treat `logL_new` as a boolean mask and only keep the
                        corresponding (True) samples.
            default: 'add'

        inplace: bool, optional
            Indicates whether to modify the existing array, or return a new
            frame with importance sampling applied.
            default: False

        Returns
        -------
        samples: MCMCSamples
            Importance re-weighted samples.
        """
        samples = self.copy()
        if action == 'add':
            samples.logL += logL_new
        elif action == 'replace':
            samples.logL = logL_new
        elif action == 'mask':
            samples = samples[logL_new]
        else:
            raise NotImplementedError("`action` needs to be one of "
                                      "{'add', 'replace', 'mask'}, but '%s' "
                                      "was requested." % action)

        return modify_inplace(self, samples, inplace)

```
PS: `modify_inplace` does also not affect weights
```
def modify_inplace(orig, new, inplace):
    """Conditionally modify object inplace or return."""
    if inplace:
        orig.__dict__ = copy.deepcopy(new.__dict__)
    else:
        return new 
```

lukashergt:
As mentioned in #133, `MCMCSamples.importance_sample` is not adjusting the weights (which it should).

Fixes #133 

# Checklist:

### DIFF
--- a/anesthetic/samples.py
+++ b/anesthetic/samples.py
@@ -370,8 +370,11 @@ class MCMCSamples(WeightedDataFrame):
         """
         samples = self.copy()
         if action == 'add':
+            samples.weights *= np.exp(logL_new - logL_new.max())
             samples.logL += logL_new
         elif action == 'replace':
+            logL_new2 = logL_new - samples.logL
+            samples.weights *= np.exp(logL_new2 - logL_new2.max())
             samples.logL = logL_new
         elif action == 'mask':
             samples = samples[logL_new]
--- a/tests/test_samples.py
+++ b/tests/test_samples.py
@@ -12,7 +12,7 @@ from numpy.testing import (assert_array_equal, assert_array_almost_equal,
                            assert_array_less)
 from pandas.testing import assert_frame_equal
 from matplotlib.colors import to_hex
-from scipy.stats import ks_2samp, kstest
+from scipy.stats import ks_2samp, kstest, norm
 from wedding_cake import WeddingCake
 try:
     import montepython  # noqa: F401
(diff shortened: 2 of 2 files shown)


williamjameshandley (commented):
Clearly fixes #133, with a good set of tests to check. 

One thing we should probably consider is whether or not we should re-normalise the weights at this stage so that their maximum is 1. anesthetic doesn't (and in general shouldn't, in case users modify the weights non-trivially) assume any particular normalisation, but is it worth keeping the convention by putting an extra `samples.weights/=samples.weights.max()` after the `if` block?

This would also be another point to consider #90

lukashergt:
> One thing we should probably consider is whether or not we should re-normalise the weights at this stage so that their maximum is 1. anesthetic doesn't (and in general shouldn't, in case users modify the weights non-trivially) assume any particular normalisation, but is it worth keeping the convention by putting an extra `samples.weights/=samples.weights.max()` after the `if` block?

If we were to normalise the weights here, then I think we would need to normalise them in general for MCMC samples (for whatever normalisation we choose). Otherwise a user might get confused finding their MCMC samples to sometimes be normalised, but other times to have just the raw MCMC weights.

Might be best to adapt the same normalisation for MCMCSamples as we have for NestedSamples, what do you think?

williamjameshandley:
This is a tricky one. MCMCSamples generally have useful information in unnormalised form, namely that they are integers referring to the number of rejections at that point in parameter space. This means that converting to our (self-proclaimed) arbitrary convention of normalising to weight 1 actually loses information. In my view this argues for why in the specific case of MCMCSamples we should not follow that convention. Once you've importance weighted however, I don't think you can claim that the raw weights have any additional information, so we're safe to go back to the convention of max weight being 1.

lukashergt:
> MCMCSamples generally have useful information in unnormalised form, namely that they are integers referring to the number of rejections at that point in parameter space. This means that converting to our (self-proclaimed) arbitrary convention of normalising to weight 1 actually loses information. In my view this argues for why in the specific case of MCMCSamples we should not follow that convention.

Agreed, better to keep the original MCMC weights. 

But in that case I'd also be in favour of not renormalising the weights upon importance sampling. If one were to compare an original `mc0` to an importance sampled `mc1` run, then one might try to double check the change in weights and logL. Normalisation probably will cause more confusion than help in that case.

williamjameshandley (approved):

williamjameshandley:
OK. This issue is flagged in #90 for us to consider if we make a more wholesale adjustment to weight normalisation. Please squash and merge. Many thanks @lukashergt for implementing this and @Stefan-Heimersheim for spotting the bug.

Issue #140, opened 2021-01-24
contour edgecolor matching to facecolor


lukashergt:
I propose changing the default behaviour of the color handling of contours for `anesthetic 2.0.0`. So far the default was to have colored contour faces and black edges. I suggest using the same color for faces and edges (the edges will be a bit darker than the faces anyhow, since the face colors come from a linear colormap ranging from white to input color).

This is mostly a question of aesthetical preference. When there is only one contourplot it doesn't matter all that much. However, when there are multiple contours, then I find that the different edgecolors help discern the contours.

What are other peoples opinions?

lukashergt:
### Two quick examples to show the difference:

* example with **black edges**

```python
mc0 = MCMCSamples(root="/Users/lukashergt/Documents/Projects/AnestheticPrj/anesthetic/tests/example_data/gd")
mc1 = mc0.importance_sample(scipy.stats.norm.logpdf(mc0.x3, loc=0.4, scale=0.1))
mc2 = mc0.importance_sample(scipy.stats.norm.logpdf(mc0.x3, loc=0.6, scale=0.1))

fig, axes = make_2d_axes(['x1', 'x2', 'x3'])
mc0.plot_2d(axes, alpha=0.8);
mc1.plot_2d(axes, alpha=0.8);
mc2.plot_2d(axes, alpha=0.8);
```

![with black edges](https://user-images.githubusercontent.com/18258042/105646432-d0239e80-5e97-11eb-8580-9cc41bd230bd.png)



* example with **edgecolor equal to facecolor**

```python
fig, axes = make_2d_axes(['x1', 'x2', 'x3'])
mc0.plot_2d(axes, alpha=0.8, ec='C0');
mc1.plot_2d(axes, alpha=0.8, ec='C1');
mc2.plot_2d(axes, alpha=0.8, ec='C2');
```
![with edgecolor equal to facecolor](https://user-images.githubusercontent.com/18258042/105646447-e16cab00-5e97-11eb-9628-ec334d35c29f.png)

williamjameshandley:
Both look nice in my view, and I'm happy with either to be the default. The first version emphasises the location of the contours, but at the expense of some information since you are right that it becomes harder to distinguish which contour is which. I think given that the precise location of the contours is not necessarily something we want to emphasise, I lean slightly toward the default being the second, but would be interested to hear others opinions

lukashergt:
[...]

PR #145, opened 2021-02-08
1d Confidence Regions

### LINKED ISSUE #142: Confidence regions for 1D posteriors

I would like to be able to plot shaded regions under a 1D posterior corresponding to 68% and 95% confidence intervals using anesthetic. See example graph attached below where I have just used np.random.normal to generate some gaussian distributed "samples".

I have written a code that can do this and could submit this as a pull request. The code is in the form of a function that takes in an axis to plot onto and the samples to produce the 1D posterior. I think it would fit in to the "plots.py" code in anesthetic as it is similar to the function "fastkde_contour_plot_2d". It will need documenting, flake 8 testing ect. if this sounds good?

**Additional context**
example_1d

htjb:
# Description

Fixes #142

I have added a function, 'iso_probability_1d' to utils to find the sample values where a 1D posterior has a given probability (e.g. [0.68, 0.95]). I have also edited 'kde_plot_1d' and 'fastkde_plot_1d' so that the user can optionally shade between these regions. The function is flexible enough to take in any number of probability values/ levels (e.g. [0.68, 0.95, 0.99] ect.) and plot the related shaded regions. The user can change the colour of the shaded regions, the gradient of shading (matplotlib alpha values) and invert the shading if they want to emphases disfavoured regions rather than favoured.

Docs, tests and flake8 have all been checked locally.

The basic example use case below will give something like the attached plot.

      import numpy as np
      from anesthetic.plot import kde_plot_1d
      
      x = np.hstack([np.random.normal(0, 3, 1000), np.random.normal(10, 2, 2000)])
      
      fig, axes = plt.subplots(1, 1)
      # additional kwargs for formatting have been added to documentation
      kde_plot_1d(axes, x, fill=True)
      plt.show()

# Checklist:

example

### DIFF
--- a/anesthetic/plot.py
+++ b/anesthetic/plot.py
@@ -268,7 +268,6 @@ def fastkde_plot_1d(ax, data, *args, **kwargs):
         matplotlib matplotlib.axes.Axes.plot command)
 
     """
-
     kwargs = normalize_kwargs(
         kwargs,
         dict(linewidth=['lw'], linestyle=['ls'], color=['c'],
@@ -366,7 +365,6 @@ def kde_plot_1d(ax, data, *args, **kwargs):
         matplotlib matplotlib.axes.Axes.plot command)
 
     """
-
     if len(data) == 0:
         return np.zeros(0), np.zeros(0)
 
--- a/anesthetic/plot.py
+++ b/anesthetic/plot.py
@@ -295,7 +295,7 @@ def fastkde_plot_1d(ax, data, *args, **kwargs):
     ax.set_xlim(*check_bounds(x[i], xmin, xmax), auto=True)
 
     if filled1d is True:
-        c = iso_probability_contours_from_samples(p[i], contours=levels)
+        c = iso_probability_contours(p[i], contours=levels)
         cmap = kwargs.pop('cmap', basic_cmap(facecolor))
         for j in range(len(c)-1):
             ax.fill_between(x[i], p[i], where=p[i] >= c[j],
--- a/anesthetic/plot.py
+++ b/anesthetic/plot.py
@@ -25,7 +25,7 @@ except ImportError:
 import matplotlib.cbook as cbook
 import matplotlib.lines as mlines
 from matplotlib.ticker import MaxNLocator
-from matplotlib.colors import LinearSegmentedColormap
+from matplotlib.colors import LinearSegmentedColormap, from_levels_and_colors
 from matplotlib.transforms import Affine2D
 from anesthetic.utils import check_bounds, nest_level
 from anesthetic.utils import (sample_compression_1d, quantile,
@@ -296,7 +296,8 @@ def fastkde_plot_1d(ax, data, *args, **kwargs):
 
     if filled1d is True:
         c = iso_probability_contours(p[i], contours=levels)
-        cmap = kwargs.pop('cmap', basic_cmap(facecolor))
+        if cmap is None:
+            cmap = basic_cmap(facecolor)
         for j in range(len(c)-1):
             ax.fill_between(x[i], p[i], where=p[i] >= c[j],
                             color=cmap(c[j]))
@@ -392,7 +393,8 @@ def kde_plot_1d(ax, data, *args, **kwargs):
     if filled1d is True:
         c = iso_probability_contours_from_samples(pp, contours=levels,
                                                   weights=w)
-        cmap = kwargs.pop('cmap', basic_cmap(facecolor))
+        if cmap is None:
+            cmap = basic_cmap(facecolor)
         for j in range(len(c)-1):
             ax.fill_between(x[i], pp, where=pp >= c[j],
                             color=cmap(c[j]))
--- a/anesthetic/plot.py
+++ b/anesthetic/plot.py
@@ -268,6 +268,12 @@ def fastkde_plot_1d(ax, data, *args, **kwargs):
         matplotlib matplotlib.axes.Axes.plot command)
 
     """
+
+    kwargs = normalize_kwargs(
+        kwargs,
+        dict(linewidth=['lw'], linestyle=['ls'], color=['c'],
+        facecolor=['fc'], edgecolor=['ec']))
+
     if len(data) == 0:
         return np.zeros(0), np.zeros(0)
 
@@ -282,6 +288,13 @@ def fastkde_plot_1d(ax, data, *args, **kwargs):
     color = kwargs.pop('color', (next(ax._get_lines.prop_cycler)['color']
                                  if cmap is None else cmap(0.68)))
     facecolor = kwargs.pop('facecolor', False)
+
+    if 'edgecolor' in kwargs:
+        color = kwargs['edgecolor']
+        edgecolor = kwargs.pop('edgecolor')
+    else:
+        edgecolor = color
+
     q = kwargs.pop('q', '5sigma')
     q = quantile_plot_interval(q=q)
 
@@ -302,7 +315,7 @@ def fastkde_plot_1d(ax, data, *args, **kwargs):
         cmap = basic_cmap(facecolor)
         for j in range(len(c)-1):
             ax.fill_between(x[i], p[i], where=p[i] >= c[j],
-                            color=cmap(c[j]), edgecolor=color)
+                            color=cmap(c[j]), edgecolor=edgecolor)
 
     return ans
 
@@ -349,6 +362,7 @@ def kde_plot_1d(ax, data, *args, **kwargs):
         matplotlib matplotlib.axes.Axes.plot command)
 
     """
+
     if len(data) == 0:
         return np.zeros(0), np.zeros(0)
 
@@ -357,8 +371,8 @@ def kde_plot_1d(ax, data, *args, **kwargs):
 
     kwargs = normalize_kwargs(
         kwargs,
-        dict(linewidth=['lw'], linestyle=['ls'], color=['c']),
-        drop=['fc', 'ec'])
+        dict(linewidth=['lw'], linestyle=['ls'], color=['c'],
+        facecolor=['fc'], edgecolor=['ec']))
 
     levels = kwargs.pop('levels', [0.68, 0.95])
 
@@ -370,6 +384,12 @@ def kde_plot_1d(ax, data, *args, **kwargs):
     color = kwargs.pop('color', (next(ax._get_lines.prop_cycler)['color']
                                  if cmap is None else cmap(0.68)))
     facecolor = kwargs.pop('facecolor', False)
+
+    if 'edgecolor' in kwargs:
+        color = kwargs['edgecolor']
+        edgecolor = kwargs.pop('edgecolor')
+    else:
+        edgecolor = color
 
     q = kwargs.pop('q', '5sigma')
     q = quantile_plot_interval(q=q)
@@ -401,7 +421,7 @@ def kde_plot_1d(ax, data, *args, **kwargs):
         cmap = basic_cmap(facecolor)
         for j in range(len(c)-1):
             ax.fill_between(x[i], pp, where=pp >= c[j],
-                            color=cmap(c[j]), edgecolor=color)
+                            color=cmap(c[j]), edgecolor=edgecolor)
 
     return ans
 
--- a/anesthetic/plot.py
+++ b/anesthetic/plot.py
@@ -32,7 +32,8 @@ from anesthetic.utils import (sample_compression_1d, quantile,
                               triangular_sample_compression_2d,
                               iso_probability_contours,
                               iso_probability_contours_from_samples,
-                              scaled_triangulation, match_contour_to_contourf)
+                              scaled_triangulation, match_contour_to_contourf,
+                              iso_probability_1d)
 from anesthetic.boundary import cut_and_normalise_gaussian
 
 
@@ -309,6 +310,10 @@ def kde_plot_1d(ax, data, *args, **kwargs):
         lower/upper prior bound.
         optional, default None
 
+    fill: Bool
+        fills between iso probability confidence regions (68%, 95%) if True.
+        Default False.
+
     Returns
     -------
     lines: matplotlib.lines.Line2D
@@ -326,6 +331,7 @@ def kde_plot_1d(ax, data, *args, **kwargs):
         kwargs,
         dict(linewidth=['lw'], linestyle=['ls'], color=['c']),
         drop=['fc', 'ec'])
+    fill = kwargs.pop('fill', False)
     xmin = kwargs.pop('xmin', None)
     xmax = kwargs.pop('xmax', None)
     weights = kwargs.pop('weights', None)
@@ -354,6 +360,21 @@ def kde_plot_1d(ax, data, *args, **kwargs):
     pp /= pp.max()
     ans = ax.plot(x[i], pp, color=color, *args, **kwargs)
     ax.set_xlim(*check_bounds(x[i], xmin, xmax), auto=True)
+
+    if fill is True:
+        lims68, lims95 = iso_probability_1d(x[i], pp)
+        for j in range(0, len(lims68), 2):
+            lim0Lower = lims68[j]
+            lim0Upper = lims68[j+1]
+            x_per = np.linspace(lim0Lower, lim0Upper, 1000)
+            y_per = np.interp(x_per, x[i], pp)
+            ax.fill_between(x_per, y_per, color='gray', label='68%')
+        for j in range(0, len(lims95), 2):
+            lim0Lower = lims95[j]
+            lim0Upper = lims95[j+1]
+            x_per = np.linspace(lim0Lower, lim0Upper, 1000)
+            y_per = np.interp(x_per, x[i], pp)
+            ax.fill_between(x_per, y_per, color='gray', label='95%', alpha=0.3)
     return ans
 
 
--- a/anesthetic/plot.py
+++ b/anesthetic/plot.py
@@ -32,8 +32,7 @@ from anesthetic.utils import (sample_compression_1d, quantile,
                               triangular_sample_compression_2d,
                               iso_probability_contours,
                               iso_probability_contours_from_samples,
-                              scaled_triangulation, match_contour_to_contourf,
-                              iso_probability_1d)
+                              scaled_triangulation, match_contour_to_contourf)
 from anesthetic.boundary import cut_and_normalise_gaussian
 
 
@@ -233,26 +232,6 @@ def make_2d_axes(params, **kwargs):
     return fig, axes
 
 
-def fill_plot_1d(ax, x, p, levels, cmap):
-    """Plot filled regions in 1D posteriors."""
-    lims = iso_probability_1d(x, p, levels)
-    c = [round(1-level, 2) for level in levels]
-    for j in range(len(lims)):
-        if j < len(lims) - 1:
-            if lims[j] != lims[j+1]:
-                x_per = np.linspace(lims[j], lims[j+1], 1000)
-                y_per = np.interp(x_per, x, p)
-                y_max = round(y_per.max(), 2)
-                if y_max > max(c):
-                    fill_color = cmap(max(c))
-                else:
-                    for h in range(1, len(c), 1):
-                        if y_max > c[h] and \
-                                y_max <= c[h-1]:
-                            fill_color = cmap(c[h])
-                ax.fill_between(x_per, y_per, color=fill_color)
-
-
 def fastkde_plot_1d(ax, data, *args, **kwargs):
     """Plot a 1d marginalised distribution.
 
@@ -315,7 +294,10 @@ def fastkde_plot_1d(ax, data, *args, **kwargs):
     ax.set_xlim(*check_bounds(x[i], xmin, xmax), auto=True)
 
     if fill is True:
-        fill_plot_1d(ax, x[i], p[i], levels, cmap)
+        c = iso_probability_contours_from_samples(p[i], contours=levels)
+        for j in range(len(c)-1):
+            ax.fill_between(
+                x[i], p[i], where=(p[i] >= c[j]), color=cmap(c[j]))
 
     return ans
 
@@ -404,7 +386,10 @@ def kde_plot_1d(ax, data, *args, **kwargs):
     ax.set_xlim(*check_bounds(x[i], xmin, xmax), auto=True)
 
     if fill is True:
-        fill_plot_1d(ax, x[i], pp, levels, cmap)
+        c = iso_probability_contours_from_samples(pp, contours=levels)
+        for j in range(len(c)-1):
+            ax.fill_between(
+                x[i], pp, where=(pp >= c[j]), color=cmap(c[j]))
 
     return ans
 
--- a/anesthetic/plot.py
+++ b/anesthetic/plot.py
@@ -234,7 +234,7 @@ def make_2d_axes(params, **kwargs):
 
 
 def fill_plot_1d(ax, x, p, levels, cmap):
-    """Function for plotting filled regions in 1D posteriors."""
+    """Plot filled regions in 1D posteriors."""
     lims = iso_probability_1d(x, p, levels)
     c = [round(1-level, 2) for level in levels]
     for j in range(len(lims)):
--- a/anesthetic/plot.py
+++ b/anesthetic/plot.py
@@ -255,6 +255,12 @@ def fastkde_plot_1d(ax, data, *args, **kwargs):
         values at which to draw iso-probability lines.
         optional, default [0.68, 0.95]
 
+    facecolor: bool or string
+        If set to True then the 1d plot will be shaded with the value of the
+        ``color`` kwarg. Set to a string such as 'blue', 'k', 'r', 'C1' ect.
+        to define the color of the shading directly.
+        optional, default False
+
     Returns
     -------
     lines: matplotlib.lines.Line2D
@@ -289,14 +295,14 @@ def fastkde_plot_1d(ax, data, *args, **kwargs):
     ans = ax.plot(x[i], p[i], color=color, *args, **kwargs)
     ax.set_xlim(*check_bounds(x[i], xmin, xmax), auto=True)
 
-    if facecolor is True:
-        facecolor = color
+    if facecolor:
+        if facecolor is True:
+            facecolor = color
         c = iso_probability_contours(p[i], contours=levels)
-        if cmap is None:
-            cmap = basic_cmap(facecolor)
+        cmap = basic_cmap(facecolor)
         for j in range(len(c)-1):
             ax.fill_between(x[i], p[i], where=p[i] >= c[j],
-                            color=cmap(c[j]))
+                            color=cmap(c[j]), edgecolor=color)
 
     return ans
 
@@ -329,6 +335,12 @@ def kde_plot_1d(ax, data, *args, **kwargs):
     levels: list
         values at which to draw iso-probability lines.
         optional, default [0.68, 0.95]
+
+    facecolor: bool or string
+        If set to True then the 1d plot will be shaded with the value of the
+        ``color`` kwarg. Set to a string such as 'blue', 'k', 'r', 'C1' ect.
+        to define the color of the shading directly.
+        optional, default False
 
     Returns
     -------
@@ -381,15 +393,15 @@ def kde_plot_1d(ax, data, *args, **kwargs):
     ans = ax.plot(x[i], pp, color=color, *args, **kwargs)
     ax.set_xlim(*check_bounds(x[i], xmin, xmax), auto=True)
 
-    if facecolor is True:
-        facecolor = color
+    if facecolor:
+        if facecolor is True:
+            facecolor = color
         c = iso_probability_contours_from_samples(pp, contours=levels,
                                                   weights=w)
-        if cmap is None:
-            cmap = basic_cmap(facecolor)
+        cmap = basic_cmap(facecolor)
         for j in range(len(c)-1):
             ax.fill_between(x[i], pp, where=pp >= c[j],
-                            color=cmap(c[j]))
+                            color=cmap(c[j]), edgecolor=color)
 
     return ans
 
--- a/anesthetic/plot.py
+++ b/anesthetic/plot.py
@@ -251,10 +251,6 @@ def fastkde_plot_1d(ax, data, *args, **kwargs):
         lower/upper prior bound
         optional, default None
 
-    filled1d: Bool
-        fills between iso probability confidence regions if True.
-        Default False.
-
     levels: list
         values at which to draw iso-probability lines.
         optional, default [0.68, 0.95]
@@ -272,7 +268,6 @@ def fastkde_plot_1d(ax, data, *args, **kwargs):
     if data.max()-data.min() <= 0:
         return
 
-    filled1d = kwargs.pop('filled1d', False)
     levels = kwargs.pop('levels', [0.68, 0.95])
 
     xmin = kwargs.pop('xmin', None)
@@ -280,7 +275,7 @@ def fastkde_plot_1d(ax, data, *args, **kwargs):
     cmap = kwargs.pop('cmap', None)
     color = kwargs.pop('color', (next(ax._get_lines.prop_cycler)['color']
                                  if cmap is None else cmap(0.68)))
-    facecolor = kwargs.pop('facecolor', color)
+    facecolor = kwargs.pop('facecolor', False)
     q = kwargs.pop('q', '5sigma')
     q = quantile_plot_interval(q=q)
 
@@ -294,7 +289,8 @@ def fastkde_plot_1d(ax, data, *args, **kwargs):
     ans = ax.plot(x[i], p[i], color=color, *args, **kwargs)
     ax.set_xlim(*check_bounds(x[i], xmin, xmax), auto=True)
 
-    if filled1d is True:
+    if facecolor is True:
+        facecolor = color
         c = iso_probability_contours(p[i], contours=levels)
         if cmap is None:
             cmap = basic_cmap(facecolor)
@@ -330,10 +326,6 @@ def kde_plot_1d(ax, data, *args, **kwargs):
         lower/upper prior bound.
         optional, default None
 
-    filled1d: Bool
-        fills between iso probability confidence regions if True.
-        Default False.
-
     levels: list
         values at which to draw iso-probability lines.
         optional, default [0.68, 0.95]
@@ -356,7 +348,6 @@ def kde_plot_1d(ax, data, *args, **kwargs):
         dict(linewidth=['lw'], linestyle=['ls'], color=['c']),
         drop=['fc', 'ec'])
 
-    filled1d = kwargs.pop('filled1d', False)
     levels = kwargs.pop('levels', [0.68, 0.95])
 
     xmin = kwargs.pop('xmin', None)
@@ -366,7 +357,7 @@ def kde_plot_1d(ax, data, *args, **kwargs):
     cmap = kwargs.pop('cmap', None)
     color = kwargs.pop('color', (next(ax._get_lines.prop_cycler)['color']
                                  if cmap is None else cmap(0.68)))
-    facecolor = kwargs.pop('facecolor', color)
+    facecolor = kwargs.pop('facecolor', False)
 
     q = kwargs.pop('q', '5sigma')
     q = quantile_plot_interval(q=q)
@@ -390,7 +381,8 @@ def kde_plot_1d(ax, data, *args, **kwargs):
     ans = ax.plot(x[i], pp, color=color, *args, **kwargs)
     ax.set_xlim(*check_bounds(x[i], xmin, xmax), auto=True)
 
-    if filled1d is True:
+    if facecolor is True:
+        facecolor = color
         c = iso_probability_contours_from_samples(pp, contours=levels,
                                                   weights=w)
         if cmap is None:
--- a/anesthetic/plot.py
+++ b/anesthetic/plot.py
@@ -252,6 +252,26 @@ def fastkde_plot_1d(ax, data, *args, **kwargs):
         lower/upper prior bound
         optional, default None
 
+    fill: Bool
+        fills between iso probability confidence regions (68%, 95%) if True.
+        Default False.
+
+    levels: list
+        amount of mass within each iso-probability lines.
+        optional, default [0.68, 0.95]
+
+    fill_color: string
+        string defining the color to plot filled regions with
+        optional, default 'gray'.
+
+    fill_alpha: list
+        variation in intensity of color to shade each region with
+        optional, default [1, 0.5]
+
+    invert_alpha: Bool
+        flips shading on filled regions
+        default False
+
     Returns
     -------
     lines: matplotlib.lines.Line2D
@@ -264,6 +284,12 @@ def fastkde_plot_1d(ax, data, *args, **kwargs):
 
     if data.max()-data.min() <= 0:
         return
+
+    fill = kwargs.pop('fill', False)
+    levels = kwargs.pop('levels', [0.68, 0.95])
+    fill_color = kwargs.pop('fill_color', 'gray')
+    fill_alpha = kwargs.pop('fill_alpha', [1, 0.5])
+    invert_alpha = kwargs.pop('invert_alpha', False)
 
     xmin = kwargs.pop('xmin', None)
     xmax = kwargs.pop('xmax', None)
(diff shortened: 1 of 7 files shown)


williamjameshandley:
Many thanks for this @htjb. Let me know when the code is ready for review

htjb:
Hi @williamjameshandley, no problem! It looks like all the checks are passing now and the code should be ready for review. Thanks!

htjb:
Hi @williamjameshandley, no problem! It looks like all the checks are passing now and the code should be ready for review. Thanks!

anesthetic/plot.py
```diff
@@ -251,6 +252,26 @@ def fastkde_plot_1d(ax, data, *args, **kwargs):
         lower/upper prior bound
         optional, default None
 
+    fill: Bool
+        fills between iso probability confidence regions (68%, 95%) if True.
+        Default False.
+
+    levels: list
+        amount of mass within each iso-probability lines.
```
williamjameshandley, line 260: Grammar is a bit off here -- perhaps drop 'each'.

The full stops are also quite inconsistent, although I think this is endemic to a lot of the code, but it's nice for additions to be consistent (perhaps raise an issue for this, which could form a general 'docstring improvement' suggestion place).

anesthetic/utils.py
```diff
@@ -264,6 +264,46 @@ def iso_probability_contours_from_samples(pdf, contours=[0.68, 0.95],
     return c
 
 
+def iso_probability_1d(x, pdf, contours=[0.68, 0.95]):
```
williamjameshandley, line 267: I'm not entirely satisfied with this implementation involving turning points, as I think it relies on certain assumed properties of what a given PDF will look like.

 Is there any reason why the same scheme (or indeed the same functions `iso_probability_contours` or `iso_probability_contours_from_samples`)  couldn't be used here?

anesthetic/plot.py
```diff
@@ -281,6 +308,35 @@ def fastkde_plot_1d(ax, data, *args, **kwargs):
...
+                        ax.fill_between(
+                            x_per, y_per,
+                            color=fill_color,
+                            alpha=1-alpha)
+                    else:
+                        ax.fill_between(
+                            x_per, y_per,
+                            color=fill_color,
+                            alpha=alpha)
+
```
williamjameshandley, line 339: This code is repeated below, so if it does remain, you should probably move it into a helper function to avoid code repetition.

anesthetic/plot.py
```diff
@@ -281,6 +308,35 @@ def fastkde_plot_1d(ax, data, *args, **kwargs):
...
     ax.set_xlim(*check_bounds(x[i], xmin, xmax), auto=True)
+
+    if fill is True:
+        lims = iso_probability_1d(x[i], p[i], levels)
+        c = [round(1-level, 2) for level in levels]
+        for j in range(len(lims)):
+            if j < len(lims) - 1:
+                if lims[j] != lims[j+1]:
+                    x_per = np.linspace(lims[j], lims[j+1], 1000)
+                    y_per = np.interp(x_per, x[i], p[i])
```
williamjameshandley, line 319: Why do we need an interpolation? Would x[i] and p[i] not be sufficient to provide the points (i.e. if you changed the above two lines for `x_per = x[i]`, `y_per=y[i]`, would anything change?

williamjameshandley (changes requested):
Many thanks @htjb. The final result looks very slick. I think there are a couple of algorithmic/coding improvements we could make which are summarised below.

In terms of the final result, do we think it's better for gray to be the default fill colour, or for the fill colour to be the same as the line? This would make it more consistent with the contour colouring after #140. With that in mind, it might also be more consistent instead of using an alpha-based mode of fill colouring, to use the same cmap scheme as the plot. (this would potentially allow overlays of several lines).

htjb:
Hi @williamjameshandley, thanks for the comments. 

In terms of the colour and cmap issue, it sounds like a good idea to use the line colour and the existing cmap kwarg. This should simplify the changes being made. I will have a look at doing this shortly as it may effect some of the more specific comments you have.

anesthetic/utils.py
htjb: I am not sure I follow the method in iso_probability_contours but from the looks of it it uses a cdf to find the contours? 

This is slightly different to what I am proposing for the 1D posteriors I think. Interpolation of a cdf will give you contours around the mean value of the posterior or have I got that wrong? Rather than contours corresponding to where the pdf=0.68 ect.

The interpolation of the pdf divided at the turning points should work regardless of the number of turning points in the pdf (i.e. there is no shape assumption here).

anesthetic/plot.py
htjb: y_per and x_per get used in the ax.fill_between function, might be able to simplify this but I will need to have a think.

htjb:
@williamjameshandley I have now edited the code so that it uses the cmap kwarg and removed the alpha/fill_color kwargs. I also moved the repeated code for plotting the filled regions into it's own function 'fill_plot_1d' to avoid repetition and hopefully made the remaining doc strings more consistent.

There are some replies to the specific method/code comments still left above. Thanks!

williamjameshandley:
Could you update the figure above to see what it now looks like with the colour updates?

anesthetic/plot.py
williamjameshandley: Have you tried just dropping x_per and y_per and replacing them with x[i] and y[i]?

anesthetic/utils.py
williamjameshandley: Can you explain why the method used in 2d contours would not work for 1d 'contours' as well?

htjb:
> Could you update the figure above to see what it now looks like with the colour updates?

See above, with cmap = plt.get_cmap('Blues')

Also removed the invert alpha function as well since you can pass things like 'Blues_r' now.

anesthetic/plot.py
htjb: Yes this wont work unfortunately, you just get one filled region with the darkest shade of the colour map. x_per and y_per denote the sections of the pdf divided up at the limits. 

The figure below shows the different regions for the example distribution I am using here (each coloured line is a y_per-x_per). The limits of x_per denote the boundaries of each shaded region and the max value of y_per relative to the 'contour' levels gives the shade.

sections

anesthetic/utils.py
htjb: I'm not entirely sure I follow what is happening in the method to calculate 2d contours.

But I think that the method calculates a cdf from the pdf. Then you interpolate this at say 0.32 and 0.68 to get the corresponding x value which, for a 1d posterior, gives you *two* limits to fill between for the 68% confidence interval. Whereas the method I have given interpolates the pdf at 0.32 giving in the example here 4 limits. Slightly different definitions of a confidence region right?

Using a cdf based method for the same example problem I'm showing will give a graph like this.
cdf_limits

htjb: ...or something similar at least.

williamjameshandley: It calculates a CDF as a function of p (probability) not x (position), which I believe is what you have computed above.

To compute the crossing points with the CDF based approach, you would have to compute the probability at which e.g. the 68% contour occurs (using the CDF as a function of p), and then look for crossing points in the PDF with this value.

>Can you explain why the method used in 2d contours would not work for 1d 'contours' as well?

I think this is the key thing you need to think about. The CDF approach works for 2D, which arguably is substantially more challenging than 1D.

anesthetic/plot.py
williamjameshandley: I think if we can replace the turning point based strategy with a CDF based one, then we should get the crossing points without the need for this interpolation, so you should focus on the CDF based issues

williamjameshandley:
This looks awesome in blue! We should probably get @lukashergt 's opinion on the API. To be more consistent, it might be worth considering using something like the `facecolor` keyword argument (defaulting to None), although this would not quite be the same as for 2D contours, since these are by default filled (unless we changed the default to be filled for 1D).

htjb:
@williamjameshandley I did wonder about setting default to filled for 1D because at the moment you have to set fill=True and provide a cmap which is a bit messy. Perhaps just providing a cmap is sufficient?

anesthetic/utils.py
htjb: Ahh okay, I am interpolating my cdf as a function of position if this is what you mean? I will have a think and see if I can make this work for 1D, should be a lot simpler than 2D I agree.

williamjameshandley:
I'm minded purely on consistency grounds to consider filling in 1D by default if we can get this working satisfactorily. 

To give us an idea of this in action, are you able to update your gaussian mixture model example to be 3D, with more than one posterior, and loaded as MCMCSamples so that we can see how these look as overlayed triangle plots (this could also then be added as an example for plotting).

htjb:
HI @williamjameshandley, I can definitely look at expanding the example to higher dimensions perhaps with the data that is already being used in the example notebook. Just to clarify you mean plot 3 1D posteriors on the same graph right? Or a triangle plot? or both? 

I will look at changing the method to use a cdf first, shouldn't take to long (famous last words!).

williamjameshandley:
Basically I had in mind the ability to see what the overall effect of filled 1d plots down the diagonal in [figures like these](https://github.com/williamjameshandley/anesthetic/issues/140#issuecomment-766455729), but ideally with a gaussian mixture model example like you have in your first example. Something like:
```python
A = MCMCSamples(np.vstack([np.random.normal(mu1, cov1, 1000), np.random.normal(mu2, cov2, 2000)]))
B = MCMCSamples(np.vstack([np.random.normal(mu3, cov3, 1000), np.random.normal(mu4, cov4, 2000)]))
fig, ax = A.plot_2d([0,1,2])
B.plot_2d(ax)
```

anesthetic/utils.py
htjb: @williamjameshandley I have the workings of a code to calculate these credibility intervals with a cdf. I haven't yet fitted it into `plot.py` but it relies on this function `iso_probability_contours` in `utils`. My only concern is that the results look very different to those in my example figure and the reasoning behind the method doesn't make much sense to me. I would of thought that the 68% probability interval would correspond to p=0.32, which is what my original approach gives, but this method gives something like p=0.70.

cdf_method

The code (I will tidy this later):
          
          import numpy as np
          from anesthetic.utils import sample_compression_1d, iso_probability_contours
          from scipy.stats import gaussian_kde
          
          data = np.hstack([np.random.normal(0, 3, 1000), np.random.normal(10, 2, 2000)])
          
          x, w = sample_compression_1d(data, None, 1000)
          kde = gaussian_kde(x, weights=None)
          p = kde(x)
          p /= p.max()
          
          levels = iso_probability_contours(p, contours=[0.68, 0.95])
          print(levels)
          idx = np.where(p >= levels[0])[0]
          idx2 = np.where(p >= levels[1])[0]
          
          # Bit messy but I fill fix this when I add it into utils/plots
          split = []
          for i in range(1, len(idx)):
              if not np.isclose(idx[i] - idx[i-1], 1, rtol=1e-1):
                  split = idx[i-1]
                  split_far = idx[i]
          
          print(idx[0], split)
          plt.plot(x, p)
          plt.fill_between(x[idx[0]:split], p[idx[0]:split],
              color='blue', alpha=0.3, label='68%')
          plt.fill_between(
              x[split_far:idx[-1]], p[split_far:idx[-1]],
              color='blue', alpha=0.3)
          plt.fill_between(x[idx2], p[idx2], color='blue', alpha=0.9, label='95%')
          plt.savefig('cdf_method.png')
          plt.show()

htjb: Ahh we need to be using iso_probabiliy_contours_from_samples to get this working properly by the looks.

About to push some changes to the repo which should fix this.

htjb:
@williamjameshandley Here is an example plot using the code you have suggested.

I think there may be an ordering issue with the filled regions and lines.

multi_dim

williamjameshandley:
Very nice -- could you post the code that produces this plot?

anesthetic/plot.py
```diff
@@ -281,6 +292,13 @@ def fastkde_plot_1d(ax, data, *args, **kwargs):
 
     ans = ax.plot(x[i], p[i], color=color, *args, **kwargs)
     ax.set_xlim(*check_bounds(x[i], xmin, xmax), auto=True)
+
+    if fill is True:
+        c = iso_probability_contours_from_samples(p[i], contours=levels)
+        for j in range(len(c)-1):
+            ax.fill_between(
+                x[i], p[i], where=(p[i] >= c[j]), color=cmap(levels[j]))
```
williamjameshandley, line 300: Following the rest of the anesthetic conventions, this would probably be better formatted as
```python
            ax.fill_between(x[i], p[i], where=p[i] >= c[j], 
                            color=cmap(levels[j]))


```

williamjameshandley (changes requested):
Wow -- this is considerably more compact!

I think the only problems I see now are:
1. The colours don't quite seem right, as they are too deep. I think this issue arises from the fact that the 2d contours effectively use `cmap(c[j+1])`, but you may need to experiment here
2. the consideration whether we should move from `fill` to `facecolor` and change the default to a filled 1d plot. I'd value @lukashergt 's opinion on this, although it is likely we should fix the first of these before making a decision.

htjb:
> Very nice -- could you post the code that produces this plot?

Sure thing:

```python
from anesthetic.samples import MCMCSamples
import numpy as np

mu1 = [2, 10, -1]
mu2 = [0, 8, 0]
mu3 = [0, 1, 0]
mu4 = [4, 0, 2]
cov1 = 1*np.eye(3)
cov2 =  1*np.eye(3)
cov4 =  1*np.eye(3)
cov3 =  [1, 0.7, 0.2]*np.eye(3)

A = MCMCSamples(np.vstack([
    np.random.multivariate_normal(mu1, cov1, 2000), 
    np.random.multivariate_normal(mu2, cov2, 2000)]))
B = MCMCSamples(np.vstack([
    np.random.multivariate_normal(mu3, cov3, 2000), 
    np.random.multivariate_normal(mu4, cov4, 2000)]))

fig, ax = A.plot_2d(
    [0, 1], 
    types={'diagonal': 'fastkde', 'lower': 'fastkde', 'upper': 'scatter'},
    diagonal_kwargs={'fill': True, 'cmap': plt.get_cmap('Blues')})
B.plot_2d(
    ax, diagonal_kwargs={'fill': True, 'cmap': plt.get_cmap('Oranges')})
plt.savefig('multi_dim.png')
plt.show()
```

htjb:
> Wow -- this is considerably more compact!
> 
> I think the only problems I see now are:
> 
>     1. The colours don't quite seem right, as they are too deep. I think this issue arises from the fact that the 2d contours effectively use `cmap(c[j+1])`, but you may need to experiment here
> 
>     2. the consideration whether we should move from `fill` to `facecolor` and change the default to a filled 1d plot. I'd value @lukashergt 's opinion on this, although it is likely we should fix the first of these before making a decision.

Yes the addition looks a bit tidier now! I will have a look at the colour map and the intensity of plots, sounds like it should be a simple fix.

williamjameshandley:
I think that `c[j]` is what one needs to get the colours right.

htjb:
> I think that `c[j]` is what one needs to get the colours right.

Yeah it doesn't seem to work with the current scheme using a cmap with c[j] or c[j+1] but if I change to using the kwargs 'color', 'facecolor' and the `basic_colormap` function I can get the following (using default settings and changing to c[j]).

I will push the changes for you to have a look at.

multi_dim

htjb:
@williamjameshandley Could even give the filled regions in the above black edges as is done with the 2D posteriors if we are looking for consistency? The fill is still optional at the moment by the way, easy change to make this the default behaviour.

williamjameshandley:
It depends on whether we opt for the recommendations of #140. Could you produce a plot of both versions (black contours for all, coloured contours for all)

htjb:
> It depends on whether we opt for the recommendations of #140. Could you produce a plot of both versions (black contours for all, coloured contours for all)

@williamjameshandley see attached here. I had to make some edits to plot.py to get the black edge colour on the 1D posteriors but I will hold off pushing these until there is a consensus on the formatting.

multi_dim_black_contours

multi_dim_coloured_contours

williamjameshandley:
Interesting -- I had originally thought of only making the PDF curve black/coloured, but now realise that there is an option associated with the fill boundaries (black, not emphasised or coloured). Can we see a plot with the PDF curve black, but the fill boundaries unemphasised (as in the previous case)?

htjb:
Ahh okay do you mean like this?

multi_dim_black_lines

appetrosyan:
As an observation, the black concours seem to be twice the width of the border of the triangle plots. The doubled width is fine for multicoloured plots, but makes the black contours stick out like a sore thumb. I would suggest reducing the thickness to at least the thickness of the line of the bounding box (and further, if possible). 

Also, the draw ordering is an underlying issue of anaesthetic. When working on `paranestamol` I noticed that the number of objects was double and PDFs produced using matplotlib contain double the number of curves and areas, thus increasing the size of the figures. I don’t think that it’s feasible to solve within this PR, and probably opened as a separate issue.

lukashergt:
> This looks awesome in blue! We should probably get @lukashergt 's opinion on the API. To be more consistent, it might be worth considering using something like the `facecolor` keyword argument (defaulting to None), although this would not quite be the same as for 2D contours, since these are by default filled (unless we changed the default to be filled for 1D).

Hi all. Thanks for your work, @htjb.

1) Question about defaults: 
While I do like this option for single (maybe two) contours, I'm not a big fan of making this a default, since I frequently have multiple contours in one plot. Keep in mind that changing defaults is much simpler (i.e. less involved) for single contours than for multiple contours. With that in mind I'd be for optimising default behaviour towards multiple contours in one plot.

2) Is the `fill` kwarg really what we want?
For one thing, adding this kwarg goes beyond typical matplotlib kwargs. Further, I fear that the typical user will confuse this to be a kwarg for the 2d contours and will get very confused when `fill=False` does not lead to unfilled 2d contours. I see two options here:
    * If we really want the introduction of a new _kwarg_, then I think we need to have kwargs for both 1d and 2d, e.g. `filled1d` and `filled2d`.
    * As an alternative, I could imagine introducing this behaviour as a new plot _type_, i.e. another category in the list of `['kde', 'fastkde', 'hist', ...]`.

What do you guys think?

williamjameshandley:
Many thanks @lukashergt for these valuable comments.

Of these suggestions, I would lean to adding `'fillkde'` and `'fillfastkde'` as new types exposed to `NestedSamples.plot_1d`. I don't think the code in `anesthetic.plot` would need to change, as this would just involve adjusting the interface in `anesthetic.samples`

htjb:
@lukashergt Thanks! 

@williamjameshandley I also like the sound of this. It's a relatively straightforward change to samples.py by the looks. I will push it now but it does require, I think, having some kwarg to pass to `kde_plot_1d' such a `fill' or `fill1d' which I will start using.

lukashergt:
> but it does require, I think, having some kwarg to pass to `kde_plot_1d` such a `fill` or `filled1d` which I will start using.

Does it really? The way I was thinking for this to work was that `type='fillkde'` implies that the 1d plots are filled, so no need for another bool kwarg `fill=True`. And the `type=kde` automatically implies `fill=False`. So the `fill` kwarg is actually redundant and can be removed...?

lukashergt:
* > As an observation, the black concours seem to be twice the width of the border of the triangle plots. The doubled width is fine for multicoloured plots, but makes the black contours stick out like a sore thumb. I would suggest reducing the thickness to at least the thickness of the line of the bounding box (and further, if possible).

  I agree with @appetrosyan on that the black 1d lines seem a bit too thick. For the filled version we could envision to change the 1d linewidth to match that of the 2d contour lines.

* I do think that the 1d filled versions look clearer with lines delineating the filled regions. What do you guys think?

* I prefer the coloured version over the black lines, which would be another point in favour of changing the default behaviour for version 2, as discussed in #140.

htjb:
> Edit: Nevermind, should have taken a look at the code first. Yes, I guess the function in `plot.py` needs that kwarg, but the user doesn't need to worry about it in `samples.py`.

@lukashergt Yes the user doesn't need to worry about unless they are directly calling `kde_plot_1d` or the `fastkde` equivalent and want filled plots. 

The alternative is to define the function as `def kde_plot_1d(ax, data, filled1d=False, *args, **kwargs):` so that `filled1d` isn't a documented kwarg but I think this is messy and spoils the consistency in `plot.py`.

appetrosyan:
@htjb wouldn’t it be better to leave the option to add the confidence intervals in the return type? The plotting functions are overloaded as is, and I can definitely sympathise with new users not knowing what each option does. 

As a compromise, the default could be to use the old empty lines, and add the option to custom choose between the three provided options: coloured borders, black borders and a custom fill for the confidence intervals. Honestly, all three look fine, and sadly, the confidence intervals are not something that everyone may necessarily want for all the parameters. It’s a very good thing to have, and I would very much like to use it in my plots, but I don’t see why this should be made the new default. Some staging before release can be a good middle ground, allowing you to use it currently, without altering everyone else’s workflow. 

How does that sound? @lukashergt

htjb:
Hi @appetrosyan, not to sure what you mean by "to leave the option to add the confidence intervals _in the return type_?" here? 

Just to reiterate, the most recent push is set up so that `kde_plot_1d` has a kwarg `filled1d` which is False by default and I have added new `types` to the samples code for plotting `filledkde` and `filledfastkde` 1D plots. Therefore, the current default is that the intervals are not filled and a user can fill them with `plot_2d()` ect by specifying a type and also with `kde_plot_1d` by setting `filled1d=True`.  <- This is I think the best option.

If I had to pick a color/border scheme I would go with the one copied below again.

image

appetrosyan:
> Hi @appetrosyan, not to sure what you mean by "to leave the option to add the confidence intervals in the return type?" here?

Hi, this is just a suggestion, but for the record, the return type of every plot function including `kde_plot_1d()` and `plot2d()` is an `fig, axes` object. It behaves as a pandas dataframe, so in case you e.g. needed to add legends and some such, you'd use the methods of the return type object, rather than pass those in as keyword parameters. This has the advantage, that if you needed the confidence intervals added to just *one* of the 1d parameter plots, you could do `ax[paramIDX][paramIDX].filled(True)` and be allowed some other keyword arguments to control the colour, hatching and which percentiles are shown. Adding those as keywords to `plot2d()` and `kde_plot_1d()` would make the documentation overly verbose, while not adding them in, makes the API more rigid. 

I understand that this involves a bit more work, so I don't necessarily think that is *the best option*, however it has certain advantages over the way it's currently implemented. Again, I'm not insisting that it be done this way, just suggesting it as a possibility for you to consider.

htjb:
Ahh okay this kind of makes sense! I think I would have to try and implement it to fully understand it but thanks for clarifying. I would just point out that, although it might be a bit fiddly, with the current set up a user can have some filled 1d  and some not filled 1d plots. For example,

```python
from anesthetic.plot import kde_plot_1d, fastkde_contour_plot_2d
from anesthetic.samples import MCMCSamples
import numpy as np

mu1 = [2, 10, -1]
mu2 = [0, 8, 0]
cov1 = 1*np.eye(3)
cov2 =  1*np.eye(3)

A = MCMCSamples(np.vstack([
    np.random.multivariate_normal(mu1, cov1, 2000),
    np.random.multivariate_normal(mu2, cov2, 2000)]))
N = 2

fig, axes = plt.subplots(N, N, sharex='col')
for i in range(N):
    for j in range(N):
        if i < j:
            axes[i, j].axis('off')

fill = [True, False]
for i in range(N):
    kde_plot_1d(axes[i, i], A[i], filled1d=fill[i])

fastkde_contour_plot_2d(axes[1, 0], A[0], A[1])

axes[1, 1].set_yticks([])
plt.subplots_adjust(hspace=0, wspace=0)
plt.savefig('gaussian_mix_filled_notFilled.png')
plt.show()
```
gaussian_mix_filled_notFilled

appetrosyan:
@htjb This looks good! 

After a bit of thought moving to the axes object warrants a separate PR. 

@williamjameshandley As a suggestion, IMHO the best option would be to set the 1d KDE line style to the style of the boundary of the contour plot via a variable, and set both to the `cmap` value and the same thickness. Then adjust the confidence interval colours to be `cmap[i]` with `alpha = confidence*alpha`, so a 95% confidence region of a plot with alpha of 0.75 has the alpha 0.75*0.95 = 0.7125 in both the contour plot and if one chooses the 1d KDEs. This should resolve #140 .

anesthetic/plot.py
```diff
@@ -281,6 +293,14 @@ def fastkde_plot_1d(ax, data, *args, **kwargs):
 
     ans = ax.plot(x[i], p[i], color=color, *args, **kwargs)
     ax.set_xlim(*check_bounds(x[i], xmin, xmax), auto=True)
+
+    if filled1d is True:
+        c = iso_probability_contours_from_samples(p[i], contours=levels)
```
williamjameshandley, line 298: I think that fastkde requires `iso_probability_contours` rather than `iso_probability_contours_from_samples`, since fastkde uses a grid, rather than samples (you can see this difference in between the 2d contour versions.

williamjameshandley (changes requested):
My main concern is that if I run the examples in the tests, the plots it produces are not quite as expected
```python
params_x = ['x0', 'x1', 'x2']
fig, axes = ns.plot_2d(params, types={'lower': 'kde','diagonal': 'filledkde'})
```
Figure_1

```python
fig, axes = ns.plot_2d(params, types={'lower': 'kde','diagonal': 'filledfastkde'})
```

Figure_2

Neither of these quite look as expected. The problem with the difference between the two likely resides in the fastkde adjustment of `iso_probability_contours`, however the one-sided contour in `kdefill` looks buggy.

I definitely approve of this way of doing things -- the number of lines of code added is admirably compact, and @lukashergt 's suggestion of adding this as a new plot type seems to work very well. I think the defaults as stated looks good, as it is consistent with 'adding a fill' to a 1d plot, and would be consistent with #140 if/when we choose to switch to that new default.

williamjameshandley:
>Hi, this is just a suggestion, but for the record, the return type of every plot function including kde_plot_1d() and plot2d() is an fig, axes object. It behaves as a pandas dataframe, so in case you e.g. needed to add legends and some such, you'd use the methods of the return type object, rather than pass those in as keyword parameters. This has the advantage, that if you needed the confidence intervals added to just one of the 1d parameter plots, you could do ax[paramIDX][paramIDX].filled(True) and be allowed some other keyword arguments to control the colour, hatching and which percentiles are shown. Adding those as keywords to plot2d() and kde_plot_1d() would make the documentation overly verbose, while not adding them in, makes the API more rigid.

I agree that as a paradigm this would be a neat way of doing it, but it goes against the anesthetic philosophy of trying to follow matplotlib/pandas/numpy defaults/strategy as far as possible. One of the big jobs for anesthetic 2.0.0 release is a more wholesale updating of the plotting system #59, so that it extends the pandas strategy directly.

anesthetic/plot.py
htjb: Ahh sorry I missed this. It's an easy fix and I have double checked it works locally, I will push it shortly.

htjb:
> My main concern is that if I run the examples in the tests, the plots it produces are not quite as expected
> 
> ```python
> ns = NestedSamples(root='./tests/example_data/pc')
> params_x = ['x0', 'x1', 'x2']
> fig, axes = ns.plot_2d(params, types={'lower': 'kde','diagonal': 'filledkde'})
> ```
> 
> Figure_1

^This is a problem with the way in which the contours are coloured. Currently they are coloured with the value of cmap(c[j]) where c[j] is returned from `iso_probability_contours{_from_samples}`. The value of c[j] varies dependent on the shape of the distribution and so the colors are inconsistent across different 1d plots. Equally if the 95% limit has a value of c[j] = 0.01 say, as I imagine is the case in the figure above then for this blue color map the shaded region is essentially going to be the same as the background.

The way around this is to use cmap(levels[j]) where levels are always consistent across plots (eg [0.68, 0.95]). However, it starts to look inconsistent with the 2d plots. See below.

tests_kde_examples_with_cmap_levels

lukashergt:
> This is a problem with the way in which the contours are coloured. Currently they are coloured with the value of cmap(c[j]) where c[j] is returned from `iso_probability_contours{_from_samples}`. The value of c[j] varies dependent on the shape of the distribution and so the colors are inconsistent across different 1d plots. Equally if the 95% limit has a value of c[j] = 0.01 say, as I imagine is the case in the figure above then for this blue color map the shaded region is essentially going to be the same as the background.
> 
> The way around this is to use cmap(levels[j]) where levels are always consistent across plots (eg [0.68, 0.95]). However, it starts to look inconsistent with the 2d plots. See below.
> 
> tests_kde_examples_with_cmap_levels

This reminds me of [this discussion](https://github.com/williamjameshandley/anesthetic/pull/114#discussion_r461703277) from PR #114 and might be helpful for questions of cmap colouring.

htjb:
Thanks @lukashergt, I actually think it might be a bug in the code. I was calling `iso_probability_contours_from_samples` without the weights kwarg as in the case of the 2d plots. If I add this in I get...

tests_kde_examples_with_cmap_levels

Looking at the cmap values they are much more consistent now. The above might also provide an argument for adding in borders for the 1d region because the lighter region is very faint.

anesthetic/plot.py
```diff
@@ -264,11 +272,15 @@ def fastkde_plot_1d(ax, data, *args, **kwargs):
...
 
+    filled1d = kwargs.pop('filled1d', False)
+    levels = kwargs.pop('levels', [0.68, 0.95])
+
     xmin = kwargs.pop('xmin', None)
     xmax = kwargs.pop('xmax', None)
     cmap = kwargs.pop('cmap', None)
     color = kwargs.pop('color', (next(ax._get_lines.prop_cycler)['color']
                                  if cmap is None else cmap(0.68)))
+    facecolor = kwargs.pop('facecolor', color)
```
williamjameshandley, line 283: Could facecolor act as the same as filled1d, defaulting to None, but if it is equal to True, then taking color as its default? I don't think facecolor is used otherwise. This would mean we wouldn't need 'filledkde' as an option to plot_1d.

tests/test_plot.py
```diff
@@ -230,6 +230,12 @@ def test_kde_plot_1d(plot_1d):
         plot_1d(ax, data, q=0.9)
         plot_1d(ax, data, q=(0.1, 0.9))
 
+        # Check iso-probability code
+        plot_1d(
+            ax, data, filled1d=True)
+        plot_1d(
+            ax, data, filled1d=True, levels=[0.2, 0.6, 0.8])
```
williamjameshandley, line 237: Do we need the linebreaks here?

williamjameshandley (changes requested):
OK, minor changes and I think this is good to be added

anesthetic/plot.py
htjb: @williamjameshandley ``facecolor`` is used if ``cmap`` is ``None`` and ``filled1d`` is ``True``  to determine the color map. We could remove ``filled1d`` and replace,

```python
if filled1d is True:
        c = iso_probability_contours(p[i], contours=levels)
        if cmap is None:
            cmap = basic_cmap(facecolor)
        for j in range(len(c)-1):
            ax.fill_between(x[i], p[i], where=p[i] >= c[j],
                            color=cmap(c[j]))
```
with
```python
if any([facecolor, cmap]) is not None:
        c = iso_probability_contours(p[i], contours=levels)
        if cmap is None:
            cmap = basic_cmap(facecolor)
        for j in range(len(c)-1):
            ax.fill_between(x[i], p[i], where=p[i] >= c[j],
                            color=cmap(c[j]))
```
and set ``facecolor`` to None as a default. The user could then pass a color map or a facecolor from which to build a cmap to the function to trigger a filled plot?

tests/test_plot.py
htjb: No, this is not needed. I will update it in the next commit.

anesthetic/plot.py
williamjameshandley: I would probably implement this by defaulting it to False in the kwargs.pop, and then using an `if facecolor` check in place of `if filled1d`, and then within that checking `if facecolor is True: facecolor=color`. I think that covers all bases.

htjb: @williamjameshandley No problem that sounds good to me. See latest commit.

anesthetic/plot.py
```diff
@@ -281,6 +288,16 @@ def fastkde_plot_1d(ax, data, *args, **kwargs):
 
     ans = ax.plot(x[i], p[i], color=color, *args, **kwargs)
     ax.set_xlim(*check_bounds(x[i], xmin, xmax), auto=True)
+
+    if facecolor is True:
+        facecolor = color
+        c = iso_probability_contours(p[i], contours=levels)
```
williamjameshandley, line 294: This logic isn't quite right. It should be
```python
if facecolor:
    if facecolor is True:
        facecolor = color
    c = ...
```
The difference being that if you pass `facecolor='blue'` as an argument, then this would give you control over the contour colouring. It's unlikely you'd actually want this except for when you want dark edges by passing `facecolor='blue', color='k'` as kwargs.

The same should be done in the second function

tests/test_plot.py
```diff
@@ -230,6 +230,10 @@ def test_kde_plot_1d(plot_1d):
         plot_1d(ax, data, q=0.9)
         plot_1d(ax, data, q=(0.1, 0.9))
 
+        # Check iso-probability code
+        plot_1d(ax, data, facecolor=True)
+        plot_1d(ax, data, facecolor=True, levels=[0.2, 0.6, 0.8])
```
williamjameshandley, line 235: Could we add a test that checks the `facecolor` and `color` kwargs?

anesthetic/samples.py
```diff
@@ -175,9 +176,21 @@ def plot(self, ax, paramname_x, paramname_y=None, *args, **kwargs):
...
                                        ncompress=ncompress,
                                        *args, **kwargs)
+                elif plot_type == 'filledkde':
+                    if ncompress is None:
+                        ncompress = 1000
+                    return kde_plot_1d(ax, self[paramname_x],
+                                       weights=self.weights,
+                                       ncompress=ncompress,
+                                       facecolor=True,
+                                       *args, **kwargs)
```
williamjameshandley, line 186: Do we need this `plot_type` any more, given that the only difference between `'filledkde'` and `'kde'` is that the argument facecolor=True is passed?

williamjameshandley (changes requested):

anesthetic/plot.py
```diff
@@ -251,6 +251,10 @@ def fastkde_plot_1d(ax, data, *args, **kwargs):
         lower/upper prior bound
         optional, default None
 
+    levels: list
+        values at which to draw iso-probability lines.
+        optional, default [0.68, 0.95]
+
```
williamjameshandley, line 257: Should we add facecolor as a document argument?

anesthetic/plot.py
htjb: If the logic is as you suggest then when facecolor is none and cmap is also none the following

```python
if cmap is None:
        cmap = basic_cmap(facecolor)
```

will throw an error.

For it to work as you suggest you would need to remove the cmap kwarg probably and add in an edgecolor=color to fill_between. I think this would work well...

```python
if facecolor:
        if facecolor is True:
            facecolor = color
        c = iso_probability_contours(p[i], contours=levels)
        cmap = basic_cmap(facecolor)
        for j in range(len(c)-1):
            ax.fill_between(x[i], p[i], where=p[i] >= c[j],
                            color=cmap(c[j]), edgecolor=color)
```

anesthetic/samples.py
htjb: No I don't think so. It would be better to remove it I think and then users can just pass facecolor=True.

anesthetic/plot.py
htjb: @williamjameshandley This set up actually gives the user quite a bit of freedom with the colour scheme e.g.

```python
from anesthetic.samples import MCMCSamples
import numpy as np

mu1 = [2, 10, -1]
mu2 = [0, 8, 0]
cov1 = 1*np.eye(3)
cov2 =  1*np.eye(3)

A = MCMCSamples(np.vstack([
    np.random.multivariate_normal(mu1, cov1, 2000),
    np.random.multivariate_normal(mu2, cov2, 2000)]))

fig, ax = A.plot_2d(
    [0, 1],
    types={'diagonal': 'kde', 'lower': 'kde', 'upper': 'scatter'},
    diagonal_kwargs={'facecolor': 'C0', 'color': 'k'})
plt.show()
```
Looks pretty nice to me...
plot_example

williamjameshandley (approved):
Nice work, and pleasingly compact. Please squash and merge

anesthetic/plot.py
williamjameshandley: Nice!

williamjameshandley:
The only other thing is that if you want this to included in a new minor release you could bump the version number to 2.0.0-beta.6 in the README.rst

lukashergt (changes requested):
I am currently getting an error `AttributeError: 'Line2D' object has no property 'facecolor'` when trying to run the most recent example by you, @htjb. Is that last code example that you've posted a MWE that should work? What did we decide to use for `types` in the end?

htjb:
> The only other thing is that if you want this to included in a new minor release you could bump the version number to 2.0.0-beta.6 in the README.rst

This would be a good thing to do I think. I notice the version is already at 2.0.0-beta.6 I am assuming you meant 2.0.0-beta.7?

htjb:
@lukashergt in response to your first comment

```python
from anesthetic.samples import NestedSamples, make_2d_axes
import numpy as np

ns = NestedSamples(root="tests/example_data/pc_250")
fig, axes = make_2d_axes(['x0', 'x1', 'x2', 'x3', 'x4'])
ns.plot_2d(axes, diagonal_kwargs={'facecolor': 'C0', 'color': 'k'})
plt.show()
```
gives 
lukas_test_plot

With the current set up if you want facecolor to be 'C0' ie the same as the default for the 2d plots you need to set it as such otherwise it will default to None and the plots will not be filled.

I have played around with ``diagonal_kwargs={'facecolor': None}`` which does give the default behaviour and ``ec='C1', fc='C2', ms=10, diagonal_kwargs={'facecolor': 'C3'}`` which works to some extent although I don't think ``kde_1d_plot`` ect is inheriting the ``ec`` and ``fc`` parameters at all. I am guessing this is what the ``normalize_kwargs`` does? I need to look at that part of the code in more detail.

Happy to look at the testing in more detail and equally happy for it to be rolled over into a different PR.

htjb:
@lukashergt so in the latest commit the 1d plotting functions can now take ``fc`` or ``facecolor``, ``c`` or ``color`` and ``ec`` or ``edgecolor``.

I've also set it up so that the 1d line and the shaded region edges are always the same color with the ``ec`` kwarg taking presedence over the ``color`` kwarg. The default ``edgecolor`` is ``color``. Also if shading is not requested then the line color is by default given by the value of ``ec``.

Globally defined parameters also take precedence over parameters passed directly to diagonal_kwargs e.g. ``ec='C1', 
fc='C2', ms=10, diagonal_kwargs={'fc': 'C3'}`` gives for the test data

lukas_test_plot

I think that has dealt with your first three requested changes more or less. Let me know if not. I will try and have a look at the testing later today.

lukashergt:
> I have played around with `diagonal_kwargs={'facecolor': None}` which does give the default behaviour

* [ ] Matplotlib differentiates between `None` and `'None'`. `None` is normally treated as the default behaviour, i.e. pick next colour from colour cycler. `'None'` is treated as not a colour. Hence I think our check for `if facecolor` needs to be expanded to `if facecolor and facecolor not in [None, 'None', 'none']`.

&nbsp;

> Globally defined parameters also take precedence over parameters passed directly to diagonal_kwargs e.g. `ec='C1', fc='C2', ms=10, diagonal_kwargs={'fc': 'C3'}` gives for the test data

Good to see the behaviour. I don't think that is what we want, but this is not introduced by this PR, so let's leave this off for another PR.

&nbsp;

> lukas_test_plot

Thanks for plotting this. 
* Are the shaded levels for `'x2'` and `'x4'` one-sided? Or are they two-sided but one side squished?
* Is the behaviour of `'x3'` as expected? Could this be improved by having _all_ levels coloured to avoid the gap? @williamjameshandley what are your thoughts?

htjb:
@lukashergt 

I am about to push another commit with a few more detailed tests in and I have added the ``facecolor not in [None, 'none', 'None']`` statement.

>Are the shaded levels for 'x2' and 'x4' one-sided? Or are they two-sided but one side squished?

Not sure what you mean here sorry?

lukashergt:
> > Are the shaded levels for 'x2' and 'x4' one-sided? Or are they two-sided but one side squished?
> 
> Not sure what you mean here sorry?

For a Gaussian this finds the 0.025, 0.16, 0.84, and 0.975 quantiles and puts a vertical line and fills inbetween with colour, right? What is the behaviour for the `'x2'` and `'x4'` parameters in the example data (where we don't have a Gaussian, but a posterior that is bounded only from one side)?. Does it still put all 4 of those vertical quantile lines? Or does it put only 2 quantile lines? If only 2, which levels?

lukashergt (dismissed):

htjb:
> For a Gaussian this finds the 0.025, 0.16, 0.84, and 0.975 quantiles and puts a vertical line and fills inbetween with colour, right? What is the behaviour for the `'x2'` and `'x4'` parameters in the example data (where we don't have a Gaussian, but a posterior that is bounded only from one side)?. Does it still put all 4 of those vertical quantile lines? Or does it put only 2 quantile lines? If only 2, which levels?

@lukashergt We are using the same function as is used for the 2D plots here. This interpolates the probability as a function of the cdf to find the value for which 68% of the probability mass is contained for example. The relevant function in utils returns two values, ``c``, one for 68% and one for 95% and then the ``fill_between`` fills ``where= pp >= c[j]`` for j in range(2). For a one sided probability like 'x4' everything to the right of x4=3.2 is above the 68% limit so gets shaded darker right up to the edge of the pdf. Does that clarify what is happening a bit? As I said it is the same utils function as is used for the 2D plots (``iso_probability_contors{_from_samples}``).

Thanks for signing of on the changes! @williamjameshandley are you happy with the most recent edits as well?

I will have a look at the failed test now and bump the readme version to 2.0.0-beta.7.

htjb:
@williamjameshandley @lukashergt Thanks for helping with this! Hopefully people will find it a useful feature :smile:

PR #150, opened 2021-03-24
bugfix for xlim for fastkde plots

### LINKED ISSUE #149: fastkde plots change parameter limits
**Describe the bug**
Using `fastkde` in `plot_2d` sometimes increases the plot limits, e.g. adding unwanted white borders around histograms. (I think this behavior might also influence 68% bounds on `b` from `facecolor=True` but I did not confirm this yet.)
bug
Maybe this happens because FastKDE usually returns a few bounds beyond the given range and those might somehow get taken into account.

**To Reproduce**
```python
import numpy as np
import anesthetic
a = np.random.uniform(size=(1000,2))
b = anesthetic.samples.MCMCSamples(a, columns=['a','b'], limits={'a':[0,1], 'b':[0,1]})
fig, ax = b.plot_2d(['a','b'], types={'lower':'hist', 'diagonal':'fastkde'})
plt.show()
```

**Expected behavior**
Limits [0,1] as given:
expected

**Additional Context**
The author of #148 might have had the same problem, I only had to adjust prior limits due to this fastkde problem. Now however I also plot 68% limits in my plots and noticed they do not align with the ones I derived with scipy's `gaussian_kde` (`kde` in anesthetic), so while investigating that I found this.

lukashergt:
As raised in #149 there is a slight difference in how `'kde'` and `'fastkde'` treat the plot limits (`xlim`).

I believe this is down to a bug in `fastkde_plot_1d` (and `2d`?) which checks _twice_ whether the data reach up to the prior boundary (corresponding function is called `check_bounds`). This leads to (unwanted) extra 

**Subtle point to note regarding intended default behaviour for `'kde'` plots _(different from `'hist'` default)_:** 
* When the data spans the entire prior volume (e.g. uniform likelihood to uniform prior), the limits should be set to the prior boundaries. 
* When the data is contained in a much smaller volume than the prior volume (e.g. Gaussian likelihood to uniform prior), then the plotting limits should be reduced accordingly.  

  (Feel free to comment/critique.)

&nbsp;

Fixes #149 

# Checklist:

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -3,7 +3,7 @@ anesthetic: nested sampling visualisation
 =========================================
 :anesthetic: nested sampling visualisation
 :Author: Will Handley and Lukas Hergt
-:Version: 2.0.0-beta.7
+:Version: 2.0.0-beta.8
 :Homepage: https://github.com/williamjameshandley/anesthetic
 :Documentation: http://anesthetic.readthedocs.io/
 
--- a/anesthetic/kde.py
+++ b/anesthetic/kde.py
@@ -48,7 +48,7 @@ def fastkde_1d(d, xmin=None, xmax=None):
         p = p[x <= xmax]
         x = x[x <= xmax]
 
-    return x, p
+    return x, p, xmin, xmax
 
 
 def fastkde_2d(d_x, d_y, xmin=None, xmax=None, ymin=None, ymax=None):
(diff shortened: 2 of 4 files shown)


Stefan-Heimersheim:
> #150 should fix the issue from your first post. Would be great if you could test the PR and confirm that this solves the issue.

Thank you  @lukashergt! This indeed solves the (initial) issue in #149.

williamjameshandley (approved):
Many thanks @lukashergt. Please squash and merge

Stefan-Heimersheim (approved):

PR #152, opened 2021-04-08
change default edgecolor: match facecolor

### LINKED ISSUE #140: contour edgecolor matching to facecolor

lukashergt:
This implements the suggestion from #140 to match the edgecolor from contourplots to the facecolor by default.

Fixes #140 

# Checklist:


williamjameshandley (approved):
Happy to press forward with this change. Nice that it cleans up the code a bit as well.

PR #153, opened 2021-04-08
Inner ticks

lukashergt:
This implements the option of having ticks in the subplots of triangle plots or other 2d plots.

**Without** inner ticks: `inner_ticks=False`
```python
fig, axes = make_2d_axes(params, fig=fig, upper=False, tex=ns.tex, inner_ticks=False)
```
image

**With** inner ticks: `inner_ticks=True`
```python
fig, axes = make_2d_axes(params, fig=fig, upper=False, tex=ns.tex, inner_ticks=True)
```
image

Full example code snippet:
```python
import numpy as np

params = ["x0", "x1", "x2", "x3", "x4"]
ns = NestedSamples(root="/home/lukas/Documents/Projects/AnestheticPrj/anesthetic/tests/example_data/pc_250")

fig, axes = make_2d_axes(params, fig=fig, upper=False, tex=ns.tex, inner_ticks=True)
ns.plot_2d(axes);
```

# Checklist:


anesthetic/plot.py
```diff
@@ -120,6 +120,11 @@ def make_2d_axes(params, **kwargs):
             Figure to plot on.
             Default: matplotlib.pyplot.figure()
 
+        inner_ticks: bool
+            If False, plot ticks only on the very left and very bottom.
+            If True, plot ticks also in inner subplots.
+            Default: False
```
lukashergt, line 126: I first had this set up on my local branch to permanently change the plot behaviour, because I quite generally preferred having ticks, especially for parameters which might have some meaningful values (Omega_K comes to mind). However, I found that it was actually quite easy to make it optional (note, it is generally _not_ that easy to make these ticks work in this kind of lower/diagonal/upper setup), so that's what I did for now. This does mean having an additional `inner_ticks` kwarg, though. Better name suggestion?

Do we like this optional kwarg? Or would we prefer fewer kwargs and picking one appearance? If we prefer no ticks then I'll close this and keep it in my local branch. If we prefer always ticks then I can remove the extra kwarg.

williamjameshandley (commented):
This is a very nice (and careful addition). I like the attention to detail  in that there are no vertical ticks on the diagonal plots, since this subtly directs the viewer that that axis should not be considered in the same way as the others.

I think this should definitely be added as a kwarg. You could possibly just call it 'ticks', and give it a variety of options like 'inner', 'outer', None, True, False. There may be a case where people aren't actually interested in cluttering with any ticks at all (if one is just checking consistency).

As for the default, I would be happy to have all the ticks, or to leave it as 'outer'. The latter is less cluttered (and what we currently do), but the former is also quite informative and I think more similar to getdist.

Another addition I have been meaning to get into anesthetic for a while which is somewhat related is the ability to put markers onto a plot, using code similar to

```python
def plot_crosses(axes, params, *args, **kwargs):
    for param_x in axes.columns:
        for param_y in axes.index:
            ax = axes.loc[param_y,param_x]
            if ax is not None:
                ax.axvline(params[param_x], *args, **kwargs)
                ax.axhline(params[param_y], *args, **kwargs)
```
where params is a dictionary mapping paramnames to float values. This is probably better put in as an alternative PR, but it could also be added now as we are updating the api, since the natural interface to apply the above function would be through the make_axes kwarg.

williamjameshandley (dismissed):

lukashergt:
> This is a very nice (and careful addition). I like the attention to detail in that there are no vertical ticks on the diagonal plots, since this subtly directs the viewer that that axis should not be considered in the same way as the others.
> 
> I think this should definitely be added as a kwarg. You could possibly just call it 'ticks', and give it a variety of options like 'inner', 'outer', None, True, False. There may be a case where people aren't actually interested in cluttering with any ticks at all (if one is just checking consistency).

Done, options for `ticks` are now any of `['outer', 'inner', None]` with `None` removing all ticks.


> As for the default, I would be happy to have all the ticks, or to leave it as 'outer'. The latter is less cluttered (and what we currently do), but the former is also quite informative and I think more similar to getdist.

Ok, keeping things as is for now (just one kwarg away now anyhow).


> Another addition I have been meaning to get into anesthetic for a while which is somewhat related is the ability to put markers onto a plot, using code similar to
> 
> ```python
> def plot_crosses(axes, params, *args, **kwargs):
>     for param_x in axes.columns:
>         for param_y in axes.index:
>             ax = axes.loc[param_y,param_x]
>             if ax is not None:
>                 ax.axvline(params[param_x], *args, **kwargs)
>                 ax.axhline(params[param_y], *args, **kwargs)
> ```
> 
> where params is a dictionary mapping paramnames to float values. This is probably better put in as an alternative PR, but it could also be added now as we are updating the api, since the natural interface to apply the above function would be through the make_axes kwarg.

Yes, that is something I've been thinking about, too, but I think this is best to leave for a separate PR. I'll open a feature request.

williamjameshandley (changes requested):
At the moment the [coverage tests](https://app.codecov.io/gh/williamjameshandley/anesthetic/compare/153/diff#diff-YW5lc3RoZXRpYy9wbG90LnB5) are failing because the added `NotImplementedError` lines aren't being hit. Could you add a line to the test that checks that passing `ticks='foo'` yields this error?

lukashergt:
> At the moment the [coverage tests](https://app.codecov.io/gh/williamjameshandley/anesthetic/compare/153/diff#diff-YW5lc3RoZXRpYy9wbG90LnB5) are failing because the added `NotImplementedError` lines aren't being hit. Could you add a line to the test that checks that passing `ticks='foo'` yields this error?

Should be fixed, now.

williamjameshandley (approved):

PR #154, opened 2021-04-08
GUI tests quick fix due to matplotlib 3.4 API change

lukashergt:
The tests in #152 and #153 were failing, but this was actually down to an API change in matplotlib 3.4 that affected the GUI. This provides a quick fix to the GUI tests to make them pass. However with some new deprecation warnings from matplotlib it looks like we will soon have to adapt the GUI section accordingly. 

# Checklist:

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
 scipy>=1.2.0
 numpy
 pandas>=1.1.0
-matplotlib>=3.4.0
+matplotlib>=3.3.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
 scipy>=1.2.0
 numpy
 pandas>=1.1.0
-matplotlib>=3.3.0
+matplotlib>=3.4.0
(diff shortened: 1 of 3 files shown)


lukashergt:
Unsurprisingly the matplotlib API change in 3.4 is backwards incompatible and our current CI setup uses a mix of matplotlib 3.3 (tests fail with this fix) and matplotlib 3.4 (tests fail without this fix).

@williamjameshandley you are more familiar with the GUI. Do you want to make the GUI tests robust against this API change? Or should we update the requirements with `matplotlib>=3.4`?

Note, I am not even sure whether this issue might at all affect the GUI or whether it only affects the tests...

lukashergt:
> @williamjameshandley you are more familiar with the GUI. Do you want to make the GUI tests robust against this API change? Or should we update the requirements with `matplotlib>=3.4`?

Note that matplotlib version 3.4 has not made its way to conda, yet.

Note further that requiring `matplotlib>=3.4` also means dropping python 3.6 support.

lukashergt:
Ok, with the latest fix tests pass for both matplotlib versions, 3.3 and 3.4.

Merging this into #152, #153, #155, #157 should make the tests for those PRs pass as well.

williamjameshandley (approved):
Many thanks for looking into this @lukashergt, and sincere apologies for taking so long to get to the PRs. Please squash and merge, and I'll be able to get to the others once you've merged in the updates.

PR #155, opened 2021-04-08
optional normalisation by area instead of by maximum

lukashergt:
This implements the `density` kwarg to 1d plots to allow for normalisation by area instead of setting the posterior maximum to one. Do we like this? Then I would go ahead with this.

I chose the kwarg `density` in line with matplotlib's `density` kwarg for `plt.hist` which already implements exactly that.

Example behaviour:

```python
import numpy as np
from anesthetic import NestedSamples, MCMCSamples, make_1d_axes, make_2d_axes
from anesthetic.plot import kde_plot_1d, hist_plot_1d

params = ["x0", "x1", "x2"]
x0 = np.random.normal(loc=0, scale=1.0, size=2000)
x1 = np.random.normal(loc=1, scale=2.0, size=2000)
x2 = np.random.normal(loc=2, scale=0.5, size=2000)

fig, axes = plt.subplots(2, 2)
kde_plot_1d(axes[0, 0], x0)
kde_plot_1d(axes[0, 0], x1)
kde_plot_1d(axes[0, 0], x2)
hist_plot_1d(axes[0, 1], x0, bins=20, alpha=0.5)
hist_plot_1d(axes[0, 1], x1, bins=20, alpha=0.5)
hist_plot_1d(axes[0, 1], x2, bins=20, alpha=0.5)
kde_plot_1d(axes[1, 0], x0, density=True)
kde_plot_1d(axes[1, 0], x1, density=True)
kde_plot_1d(axes[1, 0], x2, density=True)
hist_plot_1d(axes[1, 1], x0, density=True, bins=20, alpha=0.5)
hist_plot_1d(axes[1, 1], x1, density=True, bins=20, alpha=0.5)
hist_plot_1d(axes[1, 1], x2, density=True, bins=20, alpha=0.5)
for ax in axes.flatten():
    ax.set_xlim(-7, 7)
```
image

Top row shows `density=False` and bottom row shows `density=True`.

# Checklist:

### DIFF
--- a/anesthetic/plot.py
+++ b/anesthetic/plot.py
@@ -382,6 +382,7 @@ def kde_plot_1d(ax, data, *args, **kwargs):
     xmax = kwargs.pop('xmax', None)
     weights = kwargs.pop('weights', None)
     ncompress = kwargs.pop('ncompress', 1000)
+    density = kwargs.pop('density', False)
     cmap = kwargs.pop('cmap', None)
     color = kwargs.pop('color', (next(ax._get_lines.prop_cycler)['color']
                                  if cmap is None else cmap(0.68)))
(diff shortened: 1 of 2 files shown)


williamjameshandley (approved):
This is a great addition. I'm not sure how easy this would be to 'test' in a robust way, so if you can't think of any to add then just squash and merge.

lukashergt:
> This is a great addition. 

Cool, added equivalent changes for `fastkde`.

> I'm not sure how easy this would be to 'test' in a robust way, so if you can't think of any to add then just squash and merge.

I've added a test that checks the height of a Gaussian and compares it to the theoretical value with a generous tolarance.  I think we don't need high precision here, just something that can clearly differentiate between "maximum set to 1" and "maximum matching normalisation". 
`np.random.seed(0)` should make this robust.

williamjameshandley (approved):
Great work! Many thanks -- please squash and merge.

PR #157, opened 2021-04-10
vmin=0 in 2d histograms

### LINKED ISSUE #156: Suggestion to make vmin=0 the default for histograms
_TL,DR:_ Histograms should look like this by default:
vmin0
instead of this
current

When plotting relatively flat posteriors (say from p.max=1.2 to p.min=0.8) the colormap will span this space, assigning 0.8 to the darkest color. Then you get plots like the 2nd one. However, I think the colour should rather represent posterior probability, which can be done by making the colormap span to 0 (`vmin=0` kwarg). `anesthetic` follows a similar idea for kde plot colours:

> 3. Many KDE plotting tools have conventions that don't play well with uniformly distributed parameters, which presents a problem if you are trying to plot priors along with your posteriors. anesthetic has a sensible mechanism, by defining the contours by the amount of iso-probability mass they contain, but colouring the fill in relation to the probability density of the contour.

For consistency with the 2d kde colouring we could making the 1st example plot the default. To do that we can introduce `vmin` as an explicit kwarg of histogram type plots, listed in the docs. The default value should be 0, though passing `None` or `'None'` or so should allow going back to the current behaviour.
The kwarg is currently passed to `ax.hist2d` in line 772 of `plot.py` when calling `hist_plot_2d()`.

**Describe alternatives you've considered**
Alternatively I would just suggest adding this kwarg in the [docs](https://anesthetic.readthedocs.io/en/latest/anesthetic.html#anesthetic.plot.hist_plot_2d) and explaining this. Currently it is not mentioned and one needs to look up the source, find `ax.hist2d` and check the [matplotlib docs](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.hist2d.html).

**Additional context**
Here's an MWE for the plots above:
```python
import anesthetic
import numpy as np
import scipy.stats as sst
import matplotlib as mpl

paramNames = ['a1','a2','a3']
limits = {'a1':[0,1], 'a2':[0,1], 'a3':[0,1]}

#Generate data

np.random.seed(42)
a = np.random.uniform(size=(100000,3))
likelihood = lambda a1,a2: sst.norm.pdf(a1, scale=0.3)+sst.norm.pdf(a2, scale=0.3)
weights = likelihood(a[:,0], a[:,1])

dynesty_results = anesthetic.samples.MCMCSamples(a, columns=paramNames, limits=limits, weights=weights)

# Plot KDE contours

fig, ax = dynesty_results.plot_2d(paramNames, types={'lower':'hist', 'diagonal':'kde'}, lower_kwargs={"cmap":mpl.cm.bwr})
plt.show()

fig, ax = dynesty_results.plot_2d(paramNames, types={'lower':'hist', 'diagonal':'kde'}, lower_kwargs={"cmap":mpl.cm.bwr, "vmin":0})
plt.show()
```

lukashergt:
This addresses #156 which pointed out that `vmin` in 2d histograms needs to be set to zero in order for uniform distributions to appear uniform. 

Fixes #156 

# Checklist:

### DIFF
--- a/anesthetic/plot.py
+++ b/anesthetic/plot.py
@@ -788,8 +788,7 @@ def hist_plot_2d(ax, data_x, data_y, *args, **kwargs):
             pdf[pdf < cmin] = np.ma.masked
         if cmax is not None:
             pdf[pdf > cmax] = np.ma.masked
-        image = ax.pcolormesh(x, y, pdf.T, cmap=cmap,
-                              vmin=vmin, vmax=pdf.max(),
+        image = ax.pcolormesh(x, y, pdf.T, cmap=cmap, vmin=vmin,
                               *args, **kwargs)
 
     ax.patches += [plt.Rectangle((0, 0), 0, 0, fc=cmap(0.999), ec=cmap(0.32),
--- a/anesthetic/plot.py
+++ b/anesthetic/plot.py
@@ -747,6 +747,7 @@ def hist_plot_2d(ax, data_x, data_y, *args, **kwargs):
     xmax = kwargs.pop('xmax', None)
     ymin = kwargs.pop('ymin', None)
     ymax = kwargs.pop('ymax', None)
+    vmin = kwargs.pop('vmin', 0)
     label = kwargs.pop('label', None)
     levels = kwargs.pop('levels', None)
     color = kwargs.pop('color', next(ax._get_lines.prop_cycler)['color'])
@@ -770,7 +771,7 @@ def hist_plot_2d(ax, data_x, data_y, *args, **kwargs):
 
     if levels is None:
         pdf, x, y, image = ax.hist2d(data_x, data_y, weights=weights,
-                                     cmap=cmap, range=rge,
+                                     cmap=cmap, range=rge, vmin=vmin,
                                      *args, **kwargs)
     else:
         bins = kwargs.pop('bins', 10)
@@ -787,7 +788,8 @@ def hist_plot_2d(ax, data_x, data_y, *args, **kwargs):
             pdf[pdf < cmin] = np.ma.masked
         if cmax is not None:
             pdf[pdf > cmax] = np.ma.masked
-        image = ax.pcolormesh(x, y, pdf.T, cmap=cmap, vmin=0, vmax=pdf.max(),
+        image = ax.pcolormesh(x, y, pdf.T, cmap=cmap,
+                              vmin=vmin, vmax=pdf.max(),
                               *args, **kwargs)
 
     ax.patches += [plt.Rectangle((0, 0), 0, 0, fc=cmap(0.999), ec=cmap(0.32),


anesthetic/plot.py
```diff
@@ -787,7 +788,8 @@ def hist_plot_2d(ax, data_x, data_y, *args, **kwargs):
             pdf[pdf < cmin] = np.ma.masked
         if cmax is not None:
             pdf[pdf > cmax] = np.ma.masked
-        image = ax.pcolormesh(x, y, pdf.T, cmap=cmap, vmin=0, vmax=pdf.max(),
+        image = ax.pcolormesh(x, y, pdf.T, cmap=cmap,
+                              vmin=vmin, vmax=pdf.max(),
```
lukashergt, line 792: Do we actually need the `vmax=pdf.max()` here? Could we remove it such that if someone wanted to fix it for some reason this would be possible?

If we do need it, do we also need it for `ax.hist2d` further up?

lukashergt: Would it make sense to normalize the histogram like in the 1d case and set `vmax=1` by default? That would make the maximum consistent across a triangle plot (keeping in mind that in most cases there won't be a colorbar to go alongside the plot).

@Stefan-Heimersheim, @williamjameshandley: Thoughts?

Stefan-Heimersheim: > `vmax=pdf.max()`

This seems to be the [default in `pcolormesh`](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.pcolormesh.html)
> **vmin, vmax**: float, default: None
> The colorbar range. If None, suitable min/max values are automatically chosen by the Normalize instance (defaults to the respective min/max values of C in case of the default linear scaling). It is deprecated to use vmin/vmax when norm is given.

So I think we can just leave the default if it doesn't do anything.


> Would it make sense to normalize the histogram like in the 1d case and set vmax=1 by default?

Doesn't this interfere with the y-axis in double-triangle plots where the y axis is actually relevant for the `upper` triangle and not necessarily 0 to 1?

Stefan-Heimersheim: > the maximum consistent across a triangle plot (keeping in mind that in most cases there won't be a colorbar to go alongside the plot).

Oh I misunderstood the `vmax=1` idea. I see the idea, but I think it might be difficult if some PDFs are much narrower than others. Say you have N(sigma=0.1) and N(sigma=1) [not too uncommon], then the 2nd PDF would peak at most at 10% of the height of the plot and be very hard to read.

Thanks Lukas for making the PR!

lukashergt: Thinking about it, I don't know anymore why I suggested that. Normalizing to maximum and setting `vmax=1` is no different to setting `vmax=pdf.max()` in all plots, so no need for that. 

But I do think that `vmax=pdf.max()` is the default anyhow, so we don't need to code it in there and by not putting it in explicitly we allow for `vmax` to be overwritten if desired, right? I've removed it in [babcf52](https://github.com/williamjameshandley/anesthetic/pull/157/commits/babcf5247c628e1c5103144941bb60fe22504b94).

williamjameshandley: @Stefan-Heimersheim, could you confirm this achieves what you want, and if so approve the review so @lukashergt can squash and merge

Stefan-Heimersheim (approved):
Thanks, this works as intended in all my applications!

PR #159, opened 2021-05-25
Change in behaviour from RuntimeError to RuntimeWarning for invalid nested sampling runs

### LINKED ISSUE #141: NestedSamples.recompute fails when logL==logL_birth==-inf
In line 739 of `anesthetic/samples.py` in `recompute(self, logL_birth, inplace)` we check that `logL <= logL_birth`:

```python
            if (samples.logL <= samples.logL_birth).any():
                raise RuntimeError("Not a valid nested sampling run. "
                                   "Require logL > logL_birth.")
```

This causes an issue in the case that `logL == logL_birth == -inf` (with weights=0). Is there a reason that the above check should be `<=` instead of `<`? If not, then I suggest changing that accordingly, if yes, then I suggest adding a check that also `weights > 0`.

Context: Stumbled on this upon trying to rerun the MWE in #94.

williamjameshandley:
# Description

This bug fixes adjusts the default behaviour. In general I have found that this error tends to occur due to floating point rounding errors at the end of the run, which can be safely discarded before recomputing the number of live points. This PR changes the behaviour to raising a RuntimeWarning with a message that indicates how many rows are discarded, and an encouragement for the user to inspect the chains files directly for anything that might be awry.

@sebhoof & @catherinewatkinson have also noted that this affects some of their pipelines -- any feedback for this potential fix would be much appreciated.

Fixes #141

# Checklist:

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -3,7 +3,7 @@ anesthetic: nested sampling visualisation
 =========================================
 :anesthetic: nested sampling visualisation
 :Author: Will Handley and Lukas Hergt
-:Version: 2.0.0-beta.8
+:Version: 2.0.0-beta.9
 :Homepage: https://github.com/williamjameshandley/anesthetic
 :Documentation: http://anesthetic.readthedocs.io/
 
--- a/anesthetic/samples.py
+++ b/anesthetic/samples.py
@@ -7,6 +7,7 @@ import os
 import numpy as np
 import pandas
 import copy
+import warnings
 from collections.abc import Sequence
 from anesthetic.plot import (make_1d_axes, make_2d_axes, fastkde_plot_1d,
                              kde_plot_1d, hist_plot_1d, scatter_plot_2d,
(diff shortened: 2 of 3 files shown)


sebhoof:
Thanks @williamjameshandley, this fixes the problem and demo.ipynb for me. I also get the RuntimeWarning as advertised.

sebhoof (commented):
Would it be useful to also report the number of exactly equal samples (`samples.logL == samples.logL_birth`)? These should usually be safe to ignore. Cases with `samples.logL < samples.logL_birth` might be more suspicious; one could report the biggest relative deviation in this case to let the user easily compare to the precision of the original data.

williamjameshandley:
OK, I've added some additional feedback which reports the number of equal samples, as I agree that this is additional useful information (equality will likely indicate rounding/truncation errors, while the wrong way round is definitely more concerning).

sebhoof (approved):

lukashergt (approved):
Looks good, MWE in #94 (which sparked #141) works again (with throwing the expected warning).

PR #162, opened 2021-05-27
Equally weighted live points

### LINKED ISSUE #160: Behavior of live_points
# Context:
I was trying to make a "static" version of anesthetic.gui, i.e make a gif of the live point evolution by taking live point sample snapshots when progressing uniformly through the logL (this may or may not be the best way of doing this). In doing this I found the fact that the live_points method returns a weighted dataframe to be maybe a bit unintuitive to me.

## MWE:
```python
import numpy as np
df=NestedSamples(root='./example_data/pc_250')
n_frames=10
for i in np.arange(0,n_frames+1,1):
    cset=df.logL_birth[~(df.logL_birth == -np.inf)]
    points = df.live_points(cset.min() + (i) / n_frames * (cset.max() - cset.min()))
    f,a=points.plot_2d(["x0","x1"],types={'lower': 'scatter'})
    f.show()
```

 
image

So my understanding is that I'm basically calling ```points.compress()``` and plotting a scatter of what's left.

In some more extreme examples I was trying to run (which I can supply input ns chains for, although they are large), the early frames can be entirely composed of 0 weight points so can't be drawn. If I can ```df.live_points(cset.min())``` where ```cset.min()``` is the lowest finite logL_birth contour I get a frame of 0.0 weight points

# Proposal
My intuition may be wrong here but I would naively expect live points to not be weighted by the posterior, if I include something hacky like:
```python
points = df.live_points(cset.min() + (i) / n_frames * (cset.max() - cset.min()))
points.weights = np.ones_like(points.weights)
f,a=points.plot_2d(["x0","x1"],types={'lower': 'scatter'})
```
then the same frame as before becomes
image

and the evolution through the frames looks qualitatively at least like what I would expect. My NS fundamentals may not be great so comments if what I am doing don't make sense are welcome!

EDIT: the actual solution I'd imagine is at least having an option to return a dataframe over a weighteddataframe rather than just squashing the weights like this

williamjameshandley:
# Description

This PR adjusts so that `NestedSamples.live_points` 
- Returns equally weighted `MCMCSamples` (rather than posterior weighted `NestedSamples`
- changes the convention so that an integer argument now refers to the live point iteration before contour i is deleted, rather than after

Fixes #160

# Checklist:

### DIFF
--- a/anesthetic/samples.py
+++ b/anesthetic/samples.py
@@ -619,7 +619,7 @@ class NestedSamples(MCMCSamples):
 
         Returns
         -------
-        live_points: NestedSamples
+        live_points: MCMCSamples
             Live points at either:
                 - contour logL (if input is float)
                 - ith iteration (if input is integer)
--- a/anesthetic/samples.py
+++ b/anesthetic/samples.py
@@ -631,7 +631,8 @@ class NestedSamples(MCMCSamples):
                 logL = float(self.logL[logL])
             except KeyError:
                 pass
-        return self[(self.logL > logL) & (self.logL_birth <= logL)]
+        i = (self.logL >= logL) & (self.logL_birth < logL)
+        return MCMCSamples(self[i], weights=np.ones(i.sum()))
 
     def posterior_points(self, beta=1):
         """Get equally weighted posterior points at temperature beta."""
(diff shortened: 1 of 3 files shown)


yallup (approved):
Tested and looks good, made a small change to the docstying for the live_points method to line up

PR #163, opened 2021-05-27
Fix for plotting with integer labels

### LINKED ISSUE #161: Failure when attempting to plot on same axes when paramnames are integers
When attempting to plot two sets of samples on the same axes when parameter names are labelled as integers rather than strings. 

**To reproduce behaviour:**
```python
from anesthetic import MCMCSamples
import numpy as np

samples_1 = MCMCSamples(data=np.random.rand(1000,3))
samples_2 = MCMCSamples(data=np.random.rand(1000,3))
fig, ax = samples_1.plot_2d([0,1,2])
samples_2.plot_2d(ax)
```
**Fails with the following error:**
```bash
/usr/local/anaconda3/lib/python3.8/site-packages/anesthetic/samples.py in plot_2d(self, axes, *args, **kwargs)
    331                                      diagonal=('diagonal' in types))
    332         else:
--> 333             fig = axes.values[~axes.isna()][0].figure
    334 
    335         for y, row in axes.iterrows():
IndexError: too many indices for array: array is 2-dimensional, but 3 were indexed
```

williamjameshandley:
# Description

This PR provides a fix for #161 which occurs whenever columns are named as integers.

As this code also previously used `pd.DataFrame.values` functionality, whereas the [recommended](https://stackoverflow.com/questions/17241004/how-do-i-convert-a-pandas-series-or-index-to-a-numpy-array/54324513#54324513) strategy is to use `pd.DataFrame.to_numpy()`, since [0.24.0](https://pandas-docs.github.io/pandas-docs-travis/whatsnew/v0.24.0.html#accessing-the-values-in-a-series-or-index) this PR also changes `.values` to `.to_numpy()` throughout the code, which should be the norm from now on.

Fixes #161

# Checklist:


tests/test_samples.py
```diff
@@ -913,3 +913,14 @@ def test_copy():
...
+
+def test_plotting_with_integer_names():
+    np.random.seed(3)
+    samples_1 = MCMCSamples(data=np.random.rand(1000, 3))
+    samples_2 = MCMCSamples(data=np.random.rand(1000, 3))
+    fig, ax = samples_1.plot_2d([0, 1, 2])
+    samples_2.plot_2d(ax)
+
+    fig, ax = samples_1.plot_1d([0, 1, 2])
+    samples_2.plot_1d(ax)
```
lukashergt, line 926: Might be good to expand this test beyond plotting to also test accessing integer labels. Would we have to write `samples_1['0']`? Would `samples_1.0` be possible?

williamjameshandley: `samples_1.0` is not possible -- a standard pandas restriction sadly. Have added some tests to show how accessing works i.e. `samples_1[0]` is the right syntax, not `samples_1['0']`, although this is really also 'standard' pandas (by which I mean, a gotcha that seems to get me every time).

lukashergt (approved):
Looks good to me, so approving this now. If there was other stuff that you meant to do here, feel free to request a review later again.

PR #166, opened 2021-06-21
auto update limits of derived parameters

### LINKED ISSUE #165: 2d kde plots can be much wider than actual distribution / samples (over-smoothing?)
**Describe the bug**
The contours from `plot_2d` are very large / over-smoothed if the PDF is very peaked/asymmetric (see example below). Not sure if this is a bug, I guess it would rather be a `scipy` bug, but should we do something about it?

**To Reproduce**
I uploaded some [example data](http://www.ast.cam.ac.uk/~sh2061/data.zip) where the problem occurs, the data files are currently available at `http://www.ast.cam.ac.uk/~sh2061/xdata.npy` etc.
```python
x = np.load("xdata.npy")
y = np.load("ydata.npy")
samples = anesthetic.MCMCSamples(np.array([x,y]).T, columns=['x','y'])
samples.limits={}
fig, axes = samples.plot_2d(['x','y'], types={"lower": "kde"})
samples.plot_2d(axes, types={"lower": "scatter"}, ncompress=2000)
```
These samples are sampled from a flat prior in model parameter, but the points show some derived parameters.
For these derived parameters (usually `samples['x']=function(samples)` the `limits` are not set by default. In many "less rectangular" cases `limits` don't help though, like this one:
```python
    x = np.load("xdata2.npy")
    y = np.load("ydata2.npy")
    samples = anesthetic.MCMCSamples(np.array([x,y]).T, columns=['x','y'])
    #samples.limits={}
    fig, axes = samples.plot_2d(['x','y'], types={"lower": "kde"})
    samples.plot_2d(axes, types={"lower": "scatter"}, ncompress=2000)
```

**Expected behavior**
A cutoff where the data is cut off, or close to it. Especially at the bottom 95 close to 68% contour lines. This can only be partially fixed with setting the `limits` dictionary.

I see that those samples have a rather difficult distribution. If this cannot be fixed, do you think it would be possible to have a warning in some way to make sure the user notices that the kde contours look very misleading? I guess always plotting the scatter plot (as anesthetic does by default) helps with noticing such problems.

**Screenshots**
This is what happens using the `kde` plot (first example)
xydemo_kde
2nd data set. Adding `ncompress` up to `10000` does not help.
xydemo2_kde
FastKDE can help sometimes:
xydemo2fastkde
But not always:
xydemo_fastkde
This is an example plotting the same data (first set) with `GetDist`:
getdist

**Additional context**
I checked
* Higher values of `ncompress` up to `10000` without visible change
* `fastkde` sometimes works, but produces rather jagged contours (sometimes quite unreadable)
* Setting the scalar `bw_method` [scipy](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.gaussian_kde.html) for `kde` to values between 1e-5 and 1e5 but that seems to have no effect

**Edit:** I noticed I didn't try high enough values of `ncompress` (I thought it was getting super slow but with FastKDE it actually is okay), and setting `ncompress=3200000` (which is the size of my sample set) produces plots somewhat similar to GetDist
Figure_1

lukashergt:
As addressed in #165, when one uses an `MCMCSamples` or `NestedSamples` instance to create new derived parameters, the corresponding limits are not automatically updated. This PR calls `self._set_automatic_limits()` at the start of the generic `plot`, thus ensuring that the limits of any new derived parameters are updated before plotting them.

Fixes #165 

# Checklist:

### DIFF
--- a/anesthetic/samples.py
+++ b/anesthetic/samples.py
@@ -159,6 +159,7 @@ class MCMCSamples(WeightedDataFrame):
             Pandas array of axes objects
 
         """
+        self._set_automatic_limits()
         plot_type = kwargs.pop('plot_type', 'kde')
         do_1d_plot = paramname_y is None or paramname_x == paramname_y
         kwargs['label'] = kwargs.get('label', self.label)
--- a/tests/test_samples.py
+++ b/tests/test_samples.py
@@ -632,6 +632,15 @@ def test_limit_assignment():
     assert ns.limits['logL'][1] == 5.748335384373301
     assert ns.limits['nlive'][0] == 1
     assert ns.limits['nlive'][1] == 125
+    # limits for derived parameters:
+    ns['x5'] = ns.x0 + ns.x1
+    assert 'x5' in ns.columns and 'x5' not in ns.limits
+    ns.plot_1d(['x5'])
+    assert 'x5' in ns.columns and 'x5' in ns.limits
+    ns['x6'] = ns.x2 + ns.x3
+    assert 'x6' in ns.columns and 'x6' not in ns.limits
+    ns.plot_2d(['x5', 'x6'])
+    assert 'x6' in ns.columns and 'x6' in ns.limits
 
 
 def test_xmin_xmax_1d():


Stefan-Heimersheim (approved):
Perfect, looks good; sorry for the late reply. And works on #165.

lukashergt:
Many thanks @Stefan-Heimersheim for input and review!

PR #158, opened 2021-04-27
[Draft] Support for converting from dynesty and emcee samples

Stefan-Heimersheim:
# Description

[This is a draft, mostly for feedback of how to best implement this. I didn't check for typos/functionality, but I have been using these methods myself.]

Import dynesty and emcee sample objects (in python, not from files) easily. Short version:
```python
dynesty_to_anesthetic = lambda s: anesthetic.samples.NestedSamples(data=s['samples'], weights=np.exp(s['logwt']), logL=s['logl'], columns=columns, tex=tex, limits=prior_dict)
emcee_to_anesthetic = lambda s: anesthetic.samples.MCMCSamples(data=s.flatchain, columns=columns, tex=tex)
```

# Todo

I will probably have time to finalize this in ~2 weeks
* Edit: I just noticed the nested samples currently don't work correctly, e.g. `logZ()` fails
* Check if any of the unused arguments are relevant (e.g. `dynesty_results['logz']`, or not passing `logL_birth`)
* Figure out how/if to do checks like `isinstance(dynesty_sampler, dynesty.dynesty.NestedSampler):` without adding dependency on `dynesty`? (maybe something like `Iterable`)
* There should be a way to obtain the `limits` from the samplers (`DynamicNestedSampler` even requires as input a function that returns the limits), but for now it's manual

# Checklist:

- [ ] I have performed a self-review of my own code
- [ ] My code is PEP8 compliant (`flake8 anesthetic tests`)
- [ ] My code contains compliant docstrings (`pydocstyle --convention=numpy anesthetic`)
- [ ] New and existing unit tests pass locally with my changes (`python -m pytest`)
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #167, opened 2021-06-22
add reader for Cobaya's MCMC chains (requires GetDist)

### LINKED ISSUE #129: Cobaya chain reading
I don't think as it stands that `MCMCSamples` can read [cobaya](https://cobaya.readthedocs.io) formatted chains. @lukashergt -- as a cobaya user have you found a workaround for this? Am I wrong in thinking that a `anesthetic.read.cobaya` submodule is needed in order to do this?

lukashergt:
This introduces a chain reader for Cobaya's MCMC chains. Cobaya (different from CosmoMC) no longer provides `.paramnames` files for its MCMC chains, but instead works with `.yaml` files which are simultaneously Cobaya's input files. These contain all information on the sampling parameters and derived parameters, including the tex string. The `CobayaMCMCReader` that is implemented here uses GetDist to infer the parameter labels used for the pandas dataframe as well as the tex strings for the `.tex` dictionary. Since GetDist is a requirement for Cobaya anyhow, having it as a requirement for anesthetic's `CobayaMCMCReader` shouldn't be too bad.

Fixes #129

# Checklist:


williamjameshandley (approved):
This looks like a very nice addition. I assume that since you're a Cobaya aficionado you've tested this 'in the wild' in addition to the tests you've put in here?

Once @Stefan-Heimersheim  has finished #158 for dynesty, I think the only one we're missing is cosmosis. @Pablo-Lemos does cosmosis have it's own chain formatting, or does it just rely on polychord/multinest?

Pablo-Lemos:
I believe CosmoSIS has its own formatting, and in fact annoyingly does not generate a paramnames file

williamjameshandley (approved):

Issue #132, opened 2021-01-11
Importance sampling unexpected behaviour [and plot gallery broken]


Stefan-Heimersheim:
**Describe the bug**
[NestedSamples.importance_sample](https://anesthetic.readthedocs.io/en/latest/anesthetic.html#anesthetic.samples.NestedSamples.importance_sample) changes weights in a weird way. E.g. importance sampling by a constant offset of `logL_new=0.1` changes the weights which it should not (I think) `a=chain.importance_sample(0.1, action="add")`. This change of weights influences posteriors, mean, std etc. I think the easiest is to look at the weights.

Resampling with a negative offset also appears to shorten the chain, I would not expect that either.

I might just be misunderstanding the importance sampling procedure, but this is not at all what I thought.

**To Reproduce**
```
a=chain.importance_sample(0.1, action="add")
b=chain.importance_sample(-0.1, action="add")
plt.plot(chain.weights, label='original weights')
plt.plot(a.weights, label='+0.1 weights')
plt.plot(b.weights, label='-0.1 weights')
plt.legend()
plt.show()
```

**Expected behavior**
I would expect the weights to stay the same, except for re-normalization. Instead they change.

**Screenshots**
Here I just simply plot the weights.
![Figure_1](https://user-images.githubusercontent.com/40799217/104187636-041ab080-5410-11eb-919d-cceab705d259.png)

**Additional context**
My samples have been generated with cobaya running pypolychord, then loading the "_polychord_raw" directory with anesthetic. I tried both the current git master branch and version python-anesthetic-git r319.2351380-2 from the AUR).

I tried to reproduce the issue with the sample file from the [plot gallery](http://htmlpreview.github.io/?https://github.com/williamjameshandley/cosmo_example/blob/master/demos/demo.html) but the code there does no longer work:
```
import requests
import tarfile

for filename in ["plikHM_TTTEEE_lowl_lowE_lensing.tar.gz","plikHM_TTTEEE_lowl_lowE_lensing_NS.tar.gz"]:
    github_url = "https://github.com/williamjameshandley/cosmo_example/raw/master/"
    url = github_url + filename
    open(filename, 'wb').write(requests.get(url).content)
    tarfile.open(filename).extractall()

from anesthetic import NestedSamples
nested_root = 'plikHM_TTTEEE_lowl_lowE_lensing_NS/NS_plikHM_TTTEEE_lowl_lowE_lensing'
nested = NestedSamples(root=nested_root)
```
[...]

Issue #62, opened 2019-10-31
Update plot gallery and README


williamjameshandley:
The plot gallery needs updating, and the README would benefit from including some figures that show anesthetic in action.

williamjameshandley:
I have updated the plot gallery for the latest version of the code, which will suffice for 1.3. We'll aim for a more detailed  README in 1.4

PR #168, opened 2021-06-22
axlines method to highlight specific parameter values across all subplots

### LINKED ISSUE #164: vertical and horizontal lines across whole triangle plot
This came up in #153:

> Another addition I have been meaning to get into anesthetic for a while which is somewhat related is the ability to put markers onto a plot, using code similar to
> 
> ```python
> def plot_crosses(axes, params, *args, **kwargs):
>     for param_x in axes.columns:
>         for param_y in axes.index:
>             ax = axes.loc[param_y,param_x]
>             if ax is not None:
>                 ax.axvline(params[param_x], *args, **kwargs)
>                 ax.axhline(params[param_y], *args, **kwargs)
> ```
> 
> where params is a dictionary mapping paramnames to float values. 

As a possible implementation I was wondering whether creating a method `axlines` for the axes dataframe created by `make_?d_axes` could be the way to go, matching the `matplotlib.Axes` methods `axvline` and `axhline` but broadcasting them over all relevant subplots. A code snippet could then look like this:

```python
fig, axes = make_2d_axes(['x0', 'x1'])
ns.plot_2d(axes)
axes.x0.axlines(42, *args, **kwargs)
```

Thoughts?

lukashergt:
This PR addresses #164 and provides the method `axlines` that can be applied to the `axes` dataframe which adds vertical and horizontal lines to all subplots for a given parameter value. This is achieved by subclassing `pandas.Series` and `pandas.DataFrame` (we call the new objects `AxesSeries` and `AxesDataFrame` respectively) and then creating the new method `axlines` which broadcasts matplotlib's `axvline` and `axhline` across all axes.

Fixes #164 

# Checklist:

### DIFF
--- a/anesthetic/plot.py
+++ b/anesthetic/plot.py
@@ -36,6 +36,52 @@ from anesthetic.utils import (sample_compression_1d, quantile,
 from anesthetic.boundary import cut_and_normalise_gaussian
 
 
+class AxesSeries(pandas.Series):
+    @property
+    def _constructor(self):
+        return AxesSeries
+
+    @property
+    def _constructor_expanddim(self):
+        return AxesDataFrame
+
+
+class AxesDataFrame(pandas.DataFrame):
+    @property
+    def _constructor(self):
+        return AxesDataFrame
+
+    @property
+    def _constructor_sliced(self):
+        return AxesSeries
+
+    def axlines(self, params, values, **kwargs):
+        """
+        Add vertical and horizontal lines across all axes for the given
+        parameter, value pairs.
+
+        Parameters
+        ----------
+        params : str, list(str)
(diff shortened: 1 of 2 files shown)


williamjameshandley (approved):
I like this a lot -- the extended dataframe for axes updating is very much in keeping with both anesthetic and matplotlib. 

@dominicanstey might like to have a look at using this functionality once it is merged for streamlining his plots with 'true' value crosshairs.

One thing which we ideally would do is upgrade the [plot gallery](http://htmlpreview.github.io/?https://github.com/williamjameshandley/cosmo_example/blob/master/demos/demo.html)  with these examples, so we should factor this into #62 #132.

PR #170, opened 2021-07-05
Upgrade to quantile function arising from pandas 1.3.0 internals change

lukashergt:
The failing tests in #153, #162, #163, #167, and #168 appear to be a result of problems in pandas itself, introduced in pandas 1.3.0.
Setting pandas requirements to `pandas<1.3.0` to test this hypothesis.
EDIT: Confirmed, setting `pandas<1.3.0` let's tests pass (once conda CI got fixed).

---

The problem in the failing tests turns out, that pandas `quantile` is calling itself with the kwargs `axis`, `numeric_only` and `interpolation`. However anesthetic's inherited `WeightedDataFrame` overwrites that method without allowing for all these kwargs. In the self call this then results in a TypeError.

This is pandas `quantile` method calling itself:

```python
    def quantile(
        self,
        q=0.5,
        axis: Axis = 0,
        numeric_only: bool = True,
        interpolation: str = "linear",
    ):
    """
    [...]
    """
        validate_percentile(q)
    
        if not is_list_like(q):
            # BlockManager.quantile expects listlike, so we wrap and unwrap here
>           res = self.quantile(
                [q], axis=axis, numeric_only=numeric_only, interpolation=interpolation
            )
```

This is the TypeError thrown by pytest:

```python
E           TypeError: quantile() got an unexpected keyword argument 'numeric_only'
```

Adding `**kwargs` to anesthetics `quantile` function solves this issue (see [c174ef0](https://github.com/williamjameshandley/anesthetic/pull/170/commits/c174ef0d6309b8bcd01f5bff3987507f6be1a346)).

---

# Checklist:

### DIFF
--- a/anesthetic/weighted_pandas.py
+++ b/anesthetic/weighted_pandas.py
@@ -127,9 +127,11 @@ class WeightedSeries(_WeightedObject, Series):
         """Weighted standard error of the mean."""
         return np.sqrt(self.var(skipna=skipna)/self.neff())
 
-    def quantile(self, q=0.5, **kwargs):
+    def quantile(self, q=0.5, numeric_only=True, interpolation='linear'):
         """Weighted quantile of the sampled distribution."""
-        return quantile(self.values, q, self.weights, **kwargs)
+        if not numeric_only:
+            raise NotImplementedError("numeric_only kwarg not implemented")
+        return quantile(self.values, q, self.weights, interpolation)
 
     def compress(self, nsamples=None):
         """Reduce the number of samples by discarding low-weights.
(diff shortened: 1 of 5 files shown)


anesthetic/weighted_pandas.py
```diff
@@ -266,7 +266,7 @@ def sem(self, axis=0, skipna=True):
         n = self.neff() if axis == 0 else self.shape[1]
         return np.sqrt(self.var(axis=axis, skipna=skipna)/n)
 
-    def quantile(self, q=0.5, axis=0):
+    def quantile(self, q=0.5, axis=0, **kwargs):
```
lukashergt, line 269: I've put in the generic `**kwargs` here. The actual `pandas.DataFrame.quantile` signature is:

```python
    def quantile(
        self,
        q=0.5,
        axis: Axis = 0,
        numeric_only: bool = True,
        interpolation: str = "linear",
    ):
```

Would it be better to explicitly use `numeric_only` and `interpolation` here, rather than `**kwargs`?

williamjameshandley: Yes. 03f74ef now does this, by implementing the interpolation kwarg, and throwing a `NotImplementedError` if one tries to pass `numeric_only=False`

williamjameshandley: I think that this is preferable to the `**kwargs` solution, since the tests will continue to pick up internal breaking changes like the one that triggered this.

williamjameshandley (approved):
Many thanks for spotting this @lukashergt, and my apologies for taking more than a week to be able to find the time to get round to your pull requests. Let's get this merged so the other pending PRs can be updated too.

PR #172, opened 2021-07-07
patch for getdistreader

### LINKED ISSUE #171: log(likelihood) value from anesthetic does not match getdist
When I try loading the same chains using getdist and anesthetic, the samples.loglikes value from getdist is twice as large as MCMCSamples.logL value from anesthetic. It seems that anesthetic expects getdist to have 2xloglikelihood value, as is also seen in the anesthetic function to_getdist, where the new loglike is -2xlogL of anesthetic samples.

@cmbant has confirmed that getdist loglikes should be the likelihood value rather than chi2, so I am not entirely sure why the likelihoods differ. 

I exchanged emails with @williamjameshandley and he suggested there may have been some backward incompatible change in the relevant column for getdist and that @JesusTorrado might be able to help as well?

lukashergt:
This addresses #171, which seems to suggest that anesthetic is behind GetDist, assuming the second column in the output format to be `chi2` when it really is `-logL`.

Fixes #171 

# Checklist:


williamjameshandley (changes requested):
If this is indeed the default column for getdist, then this looks like all we need. 

However, in order to really isolate this, I think that the best way to prove we've fixed it is to create some explicit tests which load our test files in `tests/example_data/gd.txt` using getdist, and explicitly compares them.

We should ideally also make sure that the example getdist is in the new format (e.g. using -logL rather than chi2).

This is also the appropriate PR to do the same for cobaya and montepython chain reading

lukashergt:
> However, in order to really isolate this, I think that the best way to prove we've fixed it is to create some explicit tests which load our test files in `tests/example_data/gd.txt` using getdist, and explicitly compares them.

Good idea, currently there is a sign difference between getdist and anesthetic when loading the `tests/example_data/gd*` data, which means that either the data is outdated or there is some inconsistency with the getdist documentation (according to the getdist docs the second column should be `-log(L)` which is how I treat it in this PR).
&nbsp;

> We should ideally also make sure that the example getdist is in the new format (e.g. using -logL rather than chi2).

The `tests/example_data/gd*` files indeed seem to be out of date. I expect they were generated using CosmoMC? Could you generate and push new ones, @williamjameshandley?
&nbsp;

> This is also the appropriate PR to do the same for cobaya and montepython chain reading

Yes, I'll start this once #167 is merged.

tests/test_reader.py
```diff
@@ -0,0 +1,105 @@
...
+def test_read_cobayamcmc():
+    np.random.seed(3)
+    mcmc = MCMCSamples(root='./tests/example_data/cb')
+    mcmc.plot_2d(['x0', 'x1'])
+    mcmc.plot_1d(['x0', 'x1'])
+    plt.close("all")
+
+    # compare directly with getdist
+    mcmc_gd = getdist.loadMCSamples(file_root="./tests/example_data/cb")
+    assert_array_almost_equal(mcmc.logL, mcmc_gd.loglikes, decimal=15)
```
lukashergt, line 40: This type of tests works with the new `cobayamcmcreader` and the `cb` example data, but it does _not_ currently work with the `getdistreader` and the `gd` example data.

williamjameshandley:
> The `tests/example_data/gd*` files indeed seem to be out of date. I expect they were generated using CosmoMC? Could you generate and push new ones, @williamjameshandley?

Done in 706a6bc, and updated the automatic generation script.

williamjameshandley (approved):
Many thanks @lukashergt I added a couple of changes to increase the coverage around this portion of code, as well as a removal of a couple of redundant (untested) lines associated with python2 compatibility. If you're happy with these please squash and merge.

lukashergt:
This is on my to-do list, but I first want to verify that the new CosmoMC behaviour is indeed to put `-logL` in the second column. Since the new `gd.txt` data was generated with anesthetic's `gen_data.py` and not with CosmoMC itself, we should double check this manually.

lukashergt:
Ok, I checked with CosmoMC, the second column indeed should be `-logL`, so this PR should take care of that.

tests/test_reader.py
lukashergt: fixed with new example data

williamjameshandley (approved):
Many thanks @lukashergt

PR #173, opened 2021-07-07
Order contour `levels` from outermost to innermost

### LINKED ISSUE #151: 2D fastkde fails when len(levels) != 2 [contains workaround for len(levels)==1]
**Describe the bug**
`plot_2d` with type `fastkde` fails with more or less than 2 `levels`, e.g.
```python
fig, ax = dynesty_results.plot_2d(paramNames, types={'lower':'fastkde'}, levels=[0.68])
fig, ax = dynesty_results.plot_2d(paramNames, types={'lower':'fastkde'}, levels=[0.68, 0.95, 0.975])
```
fail due to two contours being assumed in `utils.py`:
```python
    c0, c1, c2 = contours
ValueError: too many values to unpack (expected 3)
```
and
```python
ValueError: not enough values to unpack (expected 3, got 2)
```
respectively.

**To Reproduce**
MWE:
```python
import anesthetic
import numpy as np

paramNames = ['a1','a2','a3','a4']
limits = {'a1':[0,1], 'a2':[0,1], 'a3':[0,1], 'a4':[0,1]}

np.random.seed(42)
a = np.random.normal(size=(1000,4), scale=0.1)
weights = 1-np.exp(-1/2*(a[:,2]**2+0.5*a[:,3]**2))

dynesty_results = anesthetic.samples.MCMCSamples(a, weights=weights, columns=paramNames, limits=limits)

fig, ax = dynesty_results.plot_2d(paramNames, types={'lower':'fastkde'}, levels=[0.68, 0.95, 0.975])
plt.show()
```

**Expected behavior**
Plot should work and show 1 or 3 contours. (It does work with `kde`.)

**Partial workaround**
If one contour (e.g. 68% only) is desired, `levels=[0.68, 0.68]` produces the desired result (even with matching line styles)
```python
fig, ax = dynesty_results.plot_2d(paramNames, types={'lower':'fastkde'}, levels=[0.68, 0.68], linestyle='dashed', linewidth=4)
```
Figure_1

**Additional context**
I think this is not an urgent bug as usually one wants only one or two contours (then the workaround works), or one can use the `kde` type in many cases.

### LINKED ISSUE #169: contour levels and (e.g.) linestyles arguments use inverted ordering
**Describe the bug**
`samples.plot_2d(paramNames, types={'lower':'fastkde'}, levels=[0.68, 0.95], linestyles=['dashed', 'dotted'])` unintuitively makes the **68%** contour line **dotted**, and the **95%** contour line **dashed**.

**To Reproduce**
MWE:
```python
import anesthetic
import numpy as np

paramNames = ['a1','a2','a3','a4']
limits = {'a1':[0,1], 'a2':[0,1], 'a3':[0,1], 'a4':[0,1]}

np.random.seed(42)
a = np.random.normal(size=(1000,4), scale=0.1)
weights = 1-np.exp(-1/2*(a[:,2]**2+0.5*a[:,3]**2))

samples = anesthetic.samples.MCMCSamples(a, weights=weights, columns=paramNames, limits=limits)
fig, ax = samples.plot_2d(paramNames, types={'lower':'fastkde'}, levels=[0.68, 0.95], linestyles=['dashed', 'dotted'], lw=5)
plt.show()
```

**Expected behavior**
I would expect the n-th contour to be formatted with the n-th linestyle.

**Screenshots**
From the MWE:
Figure_1

**Additional context**
I think this is related to this line in plot.py
```python
levels = iso_probability_contours(pdf, contours=levels)
```
which inverts the levels, and also adds a third level if two levels are given as an input. The inversion of levels makes sense (as `ax.contour` requires increasing levels, and [0.68, 0.95] C.L. correspond to [high, low] PDF values), but it messes with such arguments.

Suggested solution: Can we invert the `levels` input before passing it to `iso_probability_contours`? I.e. for the same result as above the input would need to be `[0.68, 0.95]`. Then levels should be consistent with other arguments. To avoid breaking changes we could check which order the user is using, currently the `levels` _must_ be increasing numbers (as they are not being inverted) and use that check to print a warning and adjust the behaviour.

lukashergt:
This addresses #169, which noted that the input to the `levels` kwarg and that to e.g. the `linestyles` kwarg are inversely ordered to one another when given as lists. Since this is rather counterintuitive we fix this in this PR by sorting the `levels` kwarg from outermost to innermost contour levels. 

Fixes #151 
Fixes #169 

# Checklist:


anesthetic/plot.py
```diff
@@ -529,7 +529,8 @@ def fastkde_contour_plot_2d(ax, data_x, data_y, *args, **kwargs):
 
     levels: list
         amount of mass within each iso-probability contour.
-        optional, default [0.68, 0.95]
+        Has to be ordered from outermost to innermost contour.
+        optional, default [0.95, 0.68]
```
lukashergt, line 534: Do we need more than this docstring? Do we need a DeprecationWarning?

anesthetic/utils.py
```diff
@@ -225,9 +225,9 @@ def unique(a):
     return b
 
 
-def iso_probability_contours(pdf, contours=[0.68, 0.95]):
+def iso_probability_contours(pdf, contours=[0.95, 0.68]):
     """Compute the iso-probability contour values."""
-    contours = [1-p for p in reversed(contours)]
```
williamjameshandley, line 230: I think that we can get round the need for a deprecation warning by putting in a `raise ValueError` statement if the contours are not sorted in the (new) correct order (also in `iso_probability_contours_from_samples`). This will therefore pick up the breaking change in any existing code. Ideally the `ValueError` should also have a note that the breaking change was introduced in version XXX, so that surprised users at least understand where it has come from

williamjameshandley (changes requested):
Many thanks for picking this up. I agree that the new order makes a lot more sense, and have made a suggestion of how to efficiently warn users of the breaking change this bug fix introduces.

anesthetic/utils.py
```diff
@@ -243,12 +253,22 @@ def iso_probability_contours(pdf, contours=[0.68, 0.95]):
     return c
 
 
-def iso_probability_contours_from_samples(pdf, contours=[0.68, 0.95],
+def iso_probability_contours_from_samples(pdf, contours=[0.95, 0.68],
                                           weights=None):
     """Compute the iso-probability contour values."""
+    if not np.all(contours[:-1] > contours[1:]):
```
williamjameshandley, line 259: This check will throw the error if `contours` is a length 1 array. Sadly I think this is actually a legitimate use case, so we probably want this to be a `>=`. This should also be added to the test suite.

williamjameshandley (changes requested):

anesthetic/utils.py
lukashergt: For just one element there is no need to test for the order, right... so I'll just modify the conditional to exclude length 1 arrays.

anesthetic/plot.py
```diff
@@ -772,12 +779,13 @@ def kde_contour_plot_2d(ax, data_x, data_y, *args, **kwargs):
...
-    contours = iso_probability_contours_from_samples(p, weights=w)
+    levels = iso_probability_contours_from_samples(p, contours=levels,
+                                                   weights=w)
 
     if facecolor not in [None, 'None', 'none']:
         linewidths = kwargs.pop('linewidths', 0.5)
         cmap = kwargs.pop('cmap', basic_cmap(facecolor))
-        contf = ax.tricontourf(tri, p, contours, cmap=cmap, zorder=zorder,
+        contf = ax.tricontourf(tri, p, levels=levels, cmap=cmap, zorder=zorder,
                                vmin=0, vmax=p.max(), *args, **kwargs)
```
lukashergt, line 789: This tiny change is in fact almost all that was needed to sort out the problem of having more than 2 contours, thus fixing #151. 

You were right that #114 did cause some of the issues, but the issue that I previously didn't know how to address was the ambiguity of `levels` as arg or as kwarg to `tricontour` (and not to be left out of `iso_probability_contours_from_samples`...).

Does this also solve #69?

williamjameshandley: Excellent!

re #69, not quite -- we need to work out how best to reliably estimate marginal densities in the tails of the distribution (something which I'm convinced should be possible given the nature of nested sampling), but does require some more thinking.

williamjameshandley (approved):
Many thanks @lukashergt, this is a very efficient set of bug fixes!

lukashergt:
Given the slightly breaking nature of this PR (inverted ordering of `levels`), should we release version `2.0.0-beta.9` before merging this PR?

williamjameshandley:
Have released beta.9. I really should get on with #59 so that we can finalise any breaking changes to the API before moving out of beta.

lukashergt:
Thanks @Stefan-Heimersheim for all your input, ordering of `levels` and other kwargs should now match, and more or less than 2 levels should work as expected now, too.

PR #175, opened 2021-07-15
fix `axlines` and `axspans` for axes with `upper=False`

### LINKED ISSUE #174: `axlines` and `axspans` not working when `upper=False`
As the title suggests, using the new `axlines` or `axspans` methods (introduced in #168 and as originally requested in #164) won't work when setting `upper=False` in `make_2d_axes`.

This works:
```python

ns = NestedSamples(root="anesthetic/tests/example_data/pc_250")

fig, axes = make_2d_axes(["x0", "x1", "x2"], upper=True)
fig, axes = ns.plot_2d(axes);
axes.axlines('x0', 0, c='k')
```
This does not work:
```python

ns = NestedSamples(root="anesthetic/tests/example_data/pc_250")

fig, axes = make_2d_axes(["x0", "x1", "x2"], upper=False)
fig, axes = ns.plot_2d(axes);
axes.axlines('x0', 0, c='k')
```
It fails with the error:
```python
AttributeError: 'NoneType' object has no attribute 'axhline'
```

This comes down to `upper=False` turning the unused upper triangle subplots into `None` types. Hence, they will need to be excluded from the axes used by `axlines` and `axspans`.

lukashergt:
For a description of the bug, see #174. This PR aims at fixing that bug.

Fixes #174 

# Checklist:

### DIFF
--- a/anesthetic/plot.py
+++ b/anesthetic/plot.py
@@ -83,10 +83,12 @@ class AxesDataFrame(pandas.DataFrame):
         for i, param in enumerate(params):
             if param in self.columns:
                 for ax in self.loc[:, param]:
-                    ax.axvline(values[i], **kwargs)
+                    if ax is not None:
+                        ax.axvline(values[i], **kwargs)
             if param in self.index:
                 for ax in self.loc[param, self.columns != param]:
-                    ax.axhline(values[i], **kwargs)
+                    if ax is not None:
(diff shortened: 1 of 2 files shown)


lukashergt:
Failing new tests with `upper=False` highlight the issue.

williamjameshandley (approved):
Makes sense to me! Well done for finding and fixing this so quickly.

PR #177, opened 2021-08-31
Copy tex and optionally set label when merging samples

Stefan-Heimersheim:
# Description

Makes `merge_samples_weighted` copy the `tex` dict like `merge_nested_samples` does.

Also adds a `label` kwarg to `merge_samples_weighted` to set a label for the new merged samples.

# Checklist:

**Edit:** Not sure why pytest reports "xfail", does it mean there is a problem?

# Other considerations:

One could generate a label from existing labels by default. In the end I thought the functionality would be not important enough (and potentially confusing if some labels are empty) to justify the extra code
```python
if label is None:
    new_samples.label = 'mergeOf'
    for s in mcmc_samples:
        if s.label is not None:
            new_samples.label += s.label
```

Similarly I considered checking if the `tex` and `limits` values between the keys match but decided to follow `merge_nested_samples` and just overwrite the `tex` values with the one from the last run in the list of runs, and not deal with `limits` (these are particularly annoying as they almost never match if automatically generated). This is the code I would use though if one would like the functionality.
```python
        def update_key(d, key, val):
            if key in d.keys():
                assert d[key] == val, ("Cannot update dict key "+key
                    +" with conflicting values", d[key], val,
                    "Use ignore_keys=True to skip this step")
            else:
                d[key] = val

        for key in new_samples.keys():
            for s in mcmc_samples:
                if key in s.tex.keys():
                    update_key(new_samples.tex, key, s.tex[key])
                if key in s.limits.keys():
                    update_key(new_samples.limits, key, s.limits[key])
```


Stefan-Heimersheim:
I added another small modification allowing `isinstance(samples, Sequence)` or `isinstance(samples, pandas.Series)` so that pandas series can be used for merging samples as well. 

I first though to just replace `Sequence` by `Collection` but that does not distinguish between pandas Series and Dataframe.

williamjameshandley (changes requested):
Nice. This seems relatively straightforward. Could you add tests which 
1. confirm the new tex behaviour
2. check that if you don't pass label, then the label is None

Stefan-Heimersheim:
Sorry for the long delay, I now

* expanded the tex test to check all keys

* added a check in case no label is passed (i.e. is None)

williamjameshandley (approved):
Many thanks Stefan -- this looks great. Please squash and merge into master.

Issue #180, opened 2021-09-21
Improving sample compression and computation time for 1d KDE plots


lukashergt:
For complicated posterior distributions I've often needed to crank up the `ncompress` value and then been frustrated with the computation time, so I've looked into ways of speeding things up. To that end I have timed two lines from the following piece of code from `kde_plot_1d` in `plot.py`:

```python
    x, w = sample_compression_1d(data, weights, ncompress)
    tik = time.time()
    kde = gaussian_kde(x, weights=w)
    print("        KDE time =", time.time() - tik)
    tok = time.time()
    p = kde(x)
    print("   sampling time =", time.time() - tok)
```

And this is what I got for a large dataset with over 100000 samples:

```out
        KDE time = 0.009744882583618164
   sampling time = 85.50073575973511
```

So, it turns out that very little time is spent on computing the KDE. Almost all is spent on evaluating the KDE at the sample points. This leads to a few important conclusions:
1) **The line `kde = gaussian_kde(x, weights=w)` is super fast and should always be run with _all_ sample points, not only with the compressed samples.** This will significantly improve the plotted posterior.
2) The line `p = kde(x)` can be easily parallelized using the `multiprocessing` package by replacing it with:
    ```python
        from multiprocessing import Pool
        x_i = np.array_split(x, nprocesses)
        pool = Pool(processes=nprocesses)
        p_i = pool.map(kde, x_i)
        p = np.concatenate(p_i)
    ```
3) The x-values for plotting don't actually need to come from the original sample points. We could simply create a `np.linspace` instead within appropriate bounds, e.g.:
    ```python
        minx, maxx = ax.get_xlim()
        x = np.linspace(minx, maxx, num_plot_samples)
        p = kde(x)
    ```
    `num_plot_samples` is then the number of plotting samples and can be significantly smaller (e.g. 1000 instead of 100000) than what has been used for `ncompress` without altering the appearance of the posterior.

&nbsp;

This leads me to the following questions:
* Is there a reason to prefer using the original sample points for plotting rather than a `linspace`?
* If we implement point 3, should we differentiate between `ncompress` and some `num_plot_samples` and set the default of `ncompress` to all samples and use a smaller number for `num_plot_samples`? Or should we basically drop the current use of `ncompress` and use the variable name for the number of plotting samples?
* Should I, as outlined in point 2, implement the (possibly optional) parallelization of the call of `p=kde(x)`?
[...]

PR #181, opened 2021-09-22
Improve posterior KDEs for compressed data

lukashergt:
This PR addresses point 1 of issue #180.

The current version of anesthetic compresses the data before passing it to `kde=gaussian_kde(...)` to save time.  However, `gaussian_kde` is not actually the time consuming bit. Only the later _call_ to `kde(...)` is very time consuming. It is better to pass the full (uncompressed) dataset to `kde=gaussian_kde(...)` and then call `kde(...)` only on the compressed data. This won't take much longer, but will render much more precise posteriors.

# Checklist:


lukashergt:
The failing check from the test added in commit No. [2b554e5](https://github.com/williamjameshandley/anesthetic/pull/181/commits/2b554e57577263a9776fc0106d2dcf0cced8cad7) passes with the changes from commit No. [c81f474](https://github.com/williamjameshandley/anesthetic/pull/181/commits/c81f474a4a30867701d29ee7faf6a72edd3a6692), which demonstrates the improved precision we get from using the uncompressed data for `gaussian_kde`.

anesthetic/plot.py
```diff
@@ -816,16 +816,10 @@ def kde_contour_plot_2d(ax, data_x, data_y, *args, **kwargs):
...
     cov = np.cov(data_x, data_y, aweights=weights)
     tri, w = triangular_sample_compression_2d(data_x, data_y, cov,
                                               weights, ncompress)
-    kde = gaussian_kde([tri.x, tri.y], weights=w)
-
-    x, y = kde.resample(ncompress)
-    x = np.concatenate([tri.x, x])
-    y = np.concatenate([tri.y, y])
-    w = np.concatenate([w, np.zeros(ncompress)])
-    tri = scaled_triangulation(x, y, cov)
```
lukashergt, line 828: Was it ok to remove these lines? Tests don't complain...

tests/test_plot.py
```diff
@@ -330,6 +330,17 @@ def test_kde_plot_1d(plot_1d):
...
 
+def test_ncompress_kde_1d():
+    fig, ax = plt.subplots()
+    np.random.seed(0)
+    data = np.random.randn(5000)
+    line0, = kde_plot_1d(ax, data, q='5sigma', ncompress=5000)
+    line1, = kde_plot_1d(ax, data, q='5sigma', ncompress=1000)
+    x2y = interp1d(line0.get_xdata(), line0.get_ydata())
+    assert_allclose(line1.get_ydata(), x2y(line1.get_xdata()))
+    plt.close("all")
```
lukashergt, line 341: This tests the effect of `ncompress` and works nicely for the 1d case, but I can't quite find an equivalent test for the 2d case, where we don't plot the full data, but only the iso contour lines...
Do we need an `ncompress` test for the 2d case? (We didn't have one before...)

williamjameshandley (dismissed):
OK, to get the codecov tests to path I've taken the liberty of putting us closer to 100% coverage -- we'll see how the tests do, but if you're satisfied with this still, please squash and merge!

anesthetic/plot.py
williamjameshandley: I think with the new setup whatever density problem this was trying to resolve is now resolved, and it certainly seems to be doing a better job with fewer lines

lukashergt: Actually, these lines _do_ serve a purpose as the following example shows:

```python
data = np.random.randn(10000, 2)
data[:, 1] = data[:, 0]**2
ns = NestedSamples(data=data, columns=['x', 'y'])

fig = plt.figure()
fig, axes = make_2d_axes(['x', 'y'], fig=fig)
fig, axes = ns.plot_2d(axes)
```

### Without those lines:
image


### With those lines:
image

williamjameshandley: Best reinstate them then unless you have another idea.

williamjameshandley (approved):

lukashergt:
Replaced by #216.

PR #186, opened 2022-02-15
Fix to (near) empty phys_live-birth.txt file reading

williamjameshandley:
# Description

This PR begins by showing that as indicated by #182 the polychord file reading fails where the phys_live-birth.txt file is empty, and also that it fails with the same error when it contains only a single live point (demonstrated by a raised exception in the appropriate test).

I will next push a fix which shows the adjusted tests passing

Fixes #182

# Checklist:

- [X] I have performed a self-review of my own code
- [X] My code is PEP8 compliant (`flake8 anesthetic tests`)
- [X] My code contains compliant docstrings (`pydocstyle --convention=numpy anesthetic`)
- [X] New and existing unit tests pass locally with my changes (`python -m pytest`)
- [x] I have added tests that prove my fix is effective or that my feature works

Issue #182, opened 2021-09-30
Gracefully handle empty `chains/<root>_phys_live-birth.txt`. 


appetrosyan:
Currently, in anesthetic, running trying to use it on sample chains that have an empty `chains/<root>_phys_live-birth.txt` is impossible. However, simply deleting the file allows `anesthetic` to continue in loading the samples. 

I propose replacing the resultant `ValueError` with a UserWarning, or a different type of exception.

PR #187, opened 2022-02-16
Fix for unsorted logL construction bug

williamjameshandley:
# Description

In the process of solving #182 with #186, I uncovered a somewhat nasty bug associated with a premature sort command in `NestedSamples.recompute`, present since #126. This affects arrays that are constructed or loaded which don't have sorted loglikelihood contours. The effect of the bug is to randomise the `logL_birth` contour. Whilst dead points are stored in loglikelihood order, live points are generally not. I believe that bug has remained hidden since 
1. in the most instances for a completed run with reasonable (i.e. default) termination criteria the effect of somewhat randomised live points shouldn't matter 
2. this doesn't affect polychord points due to the [line](https://github.com/williamjameshandley/anesthetic/blob/ad78c426be5b0b69a63314650f8cf3a8d1779437/anesthetic/read/polychordreader.py#L15) `data = np.unique(data, axis=0)` accounting for the fact that PolyChord at the end of a run after deleting all the dead points effectively ignores the live points file.
3. The `RunTimeWarning` behaviour for slightly unsorted loglikelihood contours #159, designed to ignore numerical rounding errors at the end of a run partially masks this.

The final point means that if you have been seeing the warnings `X out of Y samples have logL <= logL_birth,\nZ of which have logL == logL_birth."` then it would be worth re-running anesthetic calculations/plots after this is merged just in case.

First commit demonstrates this with a failing test, remaining commits fix.

# Checklist:

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -3,7 +3,7 @@ anesthetic: nested sampling visualisation
 =========================================
 :anesthetic: nested sampling visualisation
 :Author: Will Handley and Lukas Hergt
-:Version: 2.0.0-beta.11
+:Version: 2.0.0-beta.12
 :Homepage: https://github.com/williamjameshandley/anesthetic
 :Documentation: http://anesthetic.readthedocs.io/
 
--- a/anesthetic/samples.py
+++ b/anesthetic/samples.py
@@ -725,10 +725,12 @@ class NestedSamples(MCMCSamples):
             frame with contours resorted and nlive recomputed
             default: False
         """
-        samples = self.sort_values('logL').reset_index(drop=True)
+        samples = self.copy()
 
         if is_int(logL_birth):
             nlive = logL_birth
+            samples.sort_values('logL', inplace=True)
+            samples.reset_index(drop=True, inplace=True)
             samples['nlive'] = nlive
             descending = np.arange(nlive, 0, -1)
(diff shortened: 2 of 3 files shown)


lukashergt (commented):
Took me a while to get what the problem is. Do I understand correctly, that the actual culprit is the old line 737 in `samples.py`? i.e. 
```python
                samples['logL_birth'] = logL_birth
```
Up to that point `samples` has not had `'logL_birth'` assigned to it (i.e. `samples` had no column called `'logL_birth'`), so the sorting that was previously applied to `samples` was not applied to `logL_birth`. But now `logL_birth` (still in the old ordering) is being assigned to `samples['logL_birth']` which should be in the new ordering. Did I get that right?

williamjameshandley:
Yes, that's exactly right -- subtle but important! 

This is proven by the fact that [2fe452c](https://github.com/williamjameshandley/anesthetic/pull/187/commits/2fe452c58fbbd0f03afaaefaefb5ccf362ebacc3) failed but [2fe452c](https://github.com/williamjameshandley/anesthetic/pull/187/commits/2fe452c58fbbd0f03afaaefaefb5ccf362ebacc3) didn't.

lukashergt (approved):

PR #179, opened 2021-08-31
[Draft] Functions to compute confidence intervals from MCMCSamples or NestedSamples

Stefan-Heimersheim:
# Description

Quick demo of #178, only implements this for `kde` and not tested for posteriors with multiple peaks. Demo works with `tests/test_samples.py`
```python
ns = NestedSamples(root='./tests/example_data/mn'); ns.confidence_1d("x0", levels=[0.95,0.68], ncompress=100000)
```

# Checklist:

- [ ] I have performed a self-review of my own code
- [ ] My code is PEP8 compliant (`flake8 anesthetic tests`)
- [ ] My code contains compliant docstrings (`pydocstyle --convention=numpy anesthetic`)
- [ ] New and existing unit tests pass locally with my changes (`python -m pytest`)
- [ ] I have added tests that prove my fix is effective or that my feature works

Issue #178, opened 2021-08-31
Function to compute iso_probability_contours for given MCMCSamples or NestedSamples object


Stefan-Heimersheim:
**Is your feature request related to a problem? Please describe.**
It is not trivial how to compute confidence intervals from `MCMCSamples` or `NestedSamples` even though we have this functionality in the 1d plots (with `facecolor=True`).

There is a function [iso_probability_contours_from_samples](https://anesthetic.readthedocs.io/en/latest/anesthetic.html?highlight=iso_probability_contours#anesthetic.utils.iso_probability_contours_from_samples) which sounds like it should do that (but does not seem to), the documentation does not explain much and the source code looks mostly like `iso_probability_contours`.

**Describe the solution you'd like**
A function analogous to [plot_1d](https://anesthetic.readthedocs.io/en/latest/anesthetic.html#anesthetic.samples.MCMCSamples.plot_1d) which returns confidence intervals, e.g. `MCMCSamples.confidence_1d(key, type='fastkde', confidence_level=[0.68, 0.95])` could return `[[(1.2,1.4), (1.53,1.55)], [(1.1, 1.6)]]`. Possible types would be `hist`, `kde`, and `fastkde`.
*Edit*: We need to return a list of tuples in case there are multiple peaks

**Describe alternatives you've considered**
In the past I just used the samples to compute the kde manually and compute this, but a built-in functionality would surely be beneficial for many users and also encourage using a sensible definition of confidence levels.

Stefan-Heimersheim:
I just copied and adjusted some code from `kde_plot_1d` to return confidence intervals. Not sure if I understand everything that is going on there though or if this is the optimal way. (Of course ideally there should be a function providing these intervals that is instead called by `kde_plot_1d`.)

lukashergt:
Hi Stefan,

for completeness it's probably good to mention that there of course is the option of calculating the standard deviation with `sample.x0.std()`. That is probably the best thing to use for approximately Gaussian posteriors.

Beyond that we need to specify exactly what we mean by confidence intervals (let's go with 68% confidence intervals in my example plots with a skewed Gaussian), where I see two possible interpretations:

1) **In terms of percentiles/quantiles:**
  For this there already is the option of using e.g. `sample.x0.quantile(0.16)`, which can be performed directly on the sample and therefore does not require any KDE computation. This places the boundaries where the posterior mass of the lower and upper tails make up 16% _each individually_.
[...]

Issue #58, opened 2019-09-11
Improved quantile computation


williamjameshandley:
The current quantile implementation is not particularly high-quality. It would be better to implement the [weighted percentile method](https://en.wikipedia.org/wiki/Percentile#The_weighted_percentile_method) in order to reduce the asymmetry that is currently present.

PR #188, opened 2022-03-22
Implemented Samples.credibility_interval

Stefan-Heimersheim:
Description
===========

Added a `credibility_interval` method to `MCMCSampels`/`NestedSamples`
which computes the credibility / confidence interval.

Fixes \#178 and replaces \#179

Notes:
======

-   Discretization error: Using `np.cumsum`, the value of the CDF at the
    first data point is `== its weight > 0`, but at the last data point
    is `== 1`. This should be negligible for `weights<<1` but for
    significant weights i.e. small sample sizes this can have an effect.
-   Should we add an alias function `confidence_interval`
-   I have also thrown in upper and lower limits which are not
    iso-probability intervals, but rely on the same code

Checklist:
==========

-   [x] I have performed a self-review of my own code
-   [x] My code is PEP8 compliant (`flake8 anesthetic tests`)
-   [x] My code contains compliant docstrings
    (`pydocstyle --convention=numpy anesthetic`)
-   [x] New and existing unit tests pass locally with my changes
    (`python -m pytest`)
-   [x] I have added tests that prove my fix is effective or that my
    feature works

### DIFF
--- a/anesthetic/samples.py
+++ b/anesthetic/samples.py
@@ -348,9 +348,10 @@ class MCMCSamples(WeightedDataFrame):
         return fig, axes
 
     def credibility_interval(self, key, level=0.68, method="iso-probability"):
-        """Compute the marginalised credibility interval, also referred to as
-        confidence interval or iso-probability contours. Also options to
-        compute upper or lower limits instead.
+        """Compute the marginalised credibility interval.
+
+        Also referred to as confidence interval or iso-probability contours.
+        Also options to compute upper or lower limits instead.
 
         Note: This is based on the discrete set of points and weights without
         interpolation. Discretization error is determined by the (normalized)
@@ -369,6 +370,7 @@ class MCMCSamples(WeightedDataFrame):
             samples), or whether to compute one-sided limits "lower-limit" /
             "upper-limit" for which `level` fraction of the samples lie
             above / below the limit.
+
         Returns
         -------
         limit: tuple or float
--- a/anesthetic/samples.py
+++ b/anesthetic/samples.py
@@ -8,6 +8,8 @@ import numpy as np
 import pandas
 import copy
 import warnings
+from scipy.interpolate import interp1d
+from scipy.optimize import minimize_scalar
 from collections.abc import Sequence
 from anesthetic.plot import (make_1d_axes, make_2d_axes, fastkde_plot_1d,
                              kde_plot_1d, hist_plot_1d, scatter_plot_2d,
@@ -344,6 +346,54 @@ class MCMCSamples(WeightedDataFrame):
                     self.plot(ax_, x, y, plot_type=plot_type, *args, **lkwargs)
 
         return fig, axes
+
+    def credibility_interval(self, key, level=0.68, method="iso-probability"):
+        """Compute marginalised credibility interval, also referred to as
+        confidence interval or iso-probability contour.
+        Parameters
+        ----------
+        key: str
+            Variable name
+        level: float
+            confidence level. Default 0.68
+        method: str
+            Whether to calculate iso-probability interval (interval with
+            same probability density at each end, "iso-probability", this
+            is also the smallest interval containing `level` fraction of
+            samples), or whether to compute one-sided limits "lower-limit" /
+            "upper-limit" for which `level` fraction of the samples lie
+            above / below the limit.
+        Returns
+        -------
+        limit: tuple or float
+            Tuple (iso-probability interval) or float (upper/lower limit)
+        """
+        assert level<1, "Level >= 1!"
+        samples = self[key]
+        weights = self.weights
+        # Sort and normalize
+        order = np.argsort(samples)
+        samples = samples[order]
+        weights = weights[order]/np.sum(weights)
+        # Compute inverse cumulative distribution function
+        # Insert 0 and 1 at the first and last value, respectively.
+        CDF = np.append(np.insert(np.cumsum(weights), 0, 0), 1)
+        S = np.array([np.min(samples), *samples, np.max(samples)])
+        invcdf = interp1d(CDF, S)
+        if method=="iso-probability":
+            # Find smallest interval
+            distance = lambda a, level=level: invcdf(a+level)-invcdf(a)
+            a = minimize_scalar(distance, bounds=(0,1-level), method="Bounded")
+            interval = (invcdf(a.x), invcdf(a.x+level))
+        elif method=="lower-limit":
+            # Get value from which we reach the desired level
+            interval = invcdf(1-level)
+        elif method=="upper-limit":
+            # Get value to which we reach the desired level
+            interval = invcdf(level)
+        else:
+            assert False, "unknown method: "+method
+        return interval
 
     def importance_sample(self, logL_new, action='add', inplace=False):
         """Perform importance re-weighting on the log-likelihood.
--- a/anesthetic/samples.py
+++ b/anesthetic/samples.py
@@ -374,7 +374,7 @@ class MCMCSamples(WeightedDataFrame):
         limit: tuple or float
             Tuple (iso-probability interval) or float (upper/lower limit)
         """
-        assert level<1, "Level >= 1!"
+        assert level < 1, "Level >= 1 is not allowed!"
         samples = self[key]
         weights = self.weights
         # Sort and normalize
@@ -386,15 +386,17 @@ class MCMCSamples(WeightedDataFrame):
         CDF = np.append(np.insert(np.cumsum(weights), 0, 0), 1)
         S = np.array([np.min(samples), *samples, np.max(samples)])
         invcdf = interp1d(CDF, S)
-        if method=="iso-probability":
+        if method == "iso-probability":
             # Find smallest interval
-            distance = lambda a, level=level: invcdf(a+level)-invcdf(a)
-            a = minimize_scalar(distance, bounds=(0,1-level), method="Bounded")
+            def distance(a, level=level):
+                return invcdf(a+level)-invcdf(a)
+            a = minimize_scalar(distance, bounds=(0, 1-level),
+                                method="Bounded")
             interval = (invcdf(a.x), invcdf(a.x+level))
-        elif method=="lower-limit":
+        elif method == "lower-limit":
             # Get value from which we reach the desired level
             interval = invcdf(1-level)
-        elif method=="upper-limit":
+        elif method == "upper-limit":
             # Get value to which we reach the desired level
             interval = invcdf(level)
         else:
--- a/anesthetic/samples.py
+++ b/anesthetic/samples.py
@@ -348,8 +348,14 @@ class MCMCSamples(WeightedDataFrame):
         return fig, axes
 
     def credibility_interval(self, key, level=0.68, method="iso-probability"):
-        """Compute marginalised credibility interval, also referred to as
-        confidence interval or iso-probability contour.
+        """Compute the marginalised credibility interval, also referred to as
+        confidence interval or iso-probability contours. Also options to
+        compute upper or lower limits instead.
+
+        Note: This is based on the discrete set of points and weights without
+        interpolation. Discretization error is determined by the (normalized)
+        weights, i.e. is small for large sample sizes with small weights.
+
         Parameters
         ----------
         key: str
--- a/tests/test_samples.py
+++ b/tests/test_samples.py
@@ -874,3 +874,12 @@ def test_plotting_with_integer_names():
     assert_array_equal(samples_1.loc[:, 0], samples_1.iloc[:, 0])
     with pytest.raises(KeyError):
         samples_1['0']
+
+
+def test_credibility_interval():
+    np.random.seed(3)
+    samples = NestedSamples(root='./tests/example_data/pc')
+    assert np.allclose([-0.1001, 0.0951], samples.credibility_interval("x0", level=0.68, method="iso-probability"), atol=0.001)
+    assert np.isclose(0.1633, samples.credibility_interval("x0", level=0.95, method="upper-limit"), atol=0.001)
+    assert np.isclose(-0.0009, samples.credibility_interval("x0", level=0.5, method="lower-limit"), atol=0.001)
+
--- a/tests/test_samples.py
+++ b/tests/test_samples.py
@@ -885,6 +885,6 @@ def test_credibility_interval():
     assert np.isclose(samples.credibility_interval("x0", level=0.95,
                                                    method="upper-limit"),
                       0.1633, atol=0.001)
-    assert np.isclose(samples.credibility_interval("x0", level=0.95,
+    assert np.isclose(samples.credibility_interval("x0", level=0.5,
                                                    method="lower-limit"),
                       -0.0009, atol=0.001)
--- a/tests/test_samples.py
+++ b/tests/test_samples.py
@@ -879,7 +879,12 @@ def test_plotting_with_integer_names():
 def test_credibility_interval():
     np.random.seed(3)
     samples = NestedSamples(root='./tests/example_data/pc')
-    assert np.allclose([-0.1001, 0.0951], samples.credibility_interval("x0", level=0.68, method="iso-probability"), atol=0.001)
-    assert np.isclose(0.1633, samples.credibility_interval("x0", level=0.95, method="upper-limit"), atol=0.001)
-    assert np.isclose(-0.0009, samples.credibility_interval("x0", level=0.5, method="lower-limit"), atol=0.001)
-
+    assert np.allclose(samples.credibility_interval("x0", level=0.68,
+                                                    method="iso-probability"),
+                       [-0.1001, 0.0951], atol=0.001)
+    assert np.isclose(samples.credibility_interval("x0", level=0.95,
+                                                   method="upper-limit"),
+                      0.1633, atol=0.001)
+    assert np.isclose(samples.credibility_interval("x0", level=0.95,
+                                                   method="lower-limit"),
+                      -0.0009, atol=0.001)


anesthetic/samples.py
```diff
@@ -345,6 +347,64 @@ def plot_2d(self, axes, *args, **kwargs):
...
+            samples), or whether to compute one-sided limits "lower-limit" /
+            "upper-limit" for which `level` fraction of the samples lie
+            above / below the limit.
+
+        Returns
+        -------
+        limit: tuple or float
+            Tuple (iso-probability interval) or float (upper/lower limit)
+        """
+        assert level < 1, "Level >= 1 is not allowed!"
```
williamjameshandley, line 379: Could you change the assertions to `raise ValueError`, which is consistent with the rest of the code

anesthetic/samples.py
```diff
@@ -345,6 +347,64 @@ def plot_2d(self, axes, *args, **kwargs):
...
+        weights = self.weights
+        # Sort and normalize
+        order = np.argsort(samples)
+        samples = samples[order]
+        weights = weights[order]/np.sum(weights)
+        # Compute inverse cumulative distribution function
+        # Insert 0 and 1 at the first and last value, respectively.
+        CDF = np.append(np.insert(np.cumsum(weights), 0, 0), 1)
+        S = np.array([np.min(samples), *samples, np.max(samples)])
+        invcdf = interp1d(CDF, S)
```
williamjameshandley, line 390: This is sufficiently general that it deserves its own `anesthetic.utils` function, and to be unified with the `anesthetic.utils.quantile` treatment.

anesthetic/samples.py
```diff
@@ -345,6 +347,64 @@ def plot_2d(self, axes, *args, **kwargs):
...
 
+    def credibility_interval(self, key, level=0.68, method="iso-probability"):
+        """Compute the marginalised credibility interval.
+
+        Also referred to as confidence interval or iso-probability contours.
+        Also options to compute upper or lower limits instead.
+
+        Note: This is based on the discrete set of points and weights without
+        interpolation. Discretization error is determined by the (normalized)
+        weights, i.e. is small for large sample sizes with small weights.
```
williamjameshandley, line 358: Is this still true?

anesthetic/samples.py
```diff
@@ -345,6 +347,64 @@ def plot_2d(self, axes, *args, **kwargs):
...
+        CDF = np.append(np.insert(np.cumsum(weights), 0, 0), 1)
+        S = np.array([np.min(samples), *samples, np.max(samples)])
+        invcdf = interp1d(CDF, S)
+        if method == "iso-probability":
+            # Find smallest interval
+            def distance(a, level=level):
+                return invcdf(a+level)-invcdf(a)
+            a = minimize_scalar(distance, bounds=(0, 1-level),
+                                method="Bounded")
+            interval = (invcdf(a.x), invcdf(a.x+level))
```
williamjameshandley, line 397: This is incredibly neat

tests/test_samples.py
```diff
@@ -885,3 +885,17 @@ def test_plotting_with_integer_names():
...
     with pytest.raises(KeyError):
         samples_1['0']
+
+
+def test_credibility_interval():
+    np.random.seed(3)
+    samples = NestedSamples(root='./tests/example_data/pc')
+    assert np.allclose(samples.credibility_interval("x0", level=0.68,
+                                                    method="iso-probability"),
+                       [-0.1001, 0.0951], atol=0.001)
```
williamjameshandley, line 895: Is it possible to change these to the 'correct' values of -0.1 and 1.1, with an appropriate atol (rather than the noisy versions here)?

williamjameshandley (changes requested):
Excellent work @Stefan-Heimersheim. The changes are only cosmetic/refactoring, so this should be good to go soon. Very thought provoking

Stefan-Heimersheim:
The sorting
```
        order = np.argsort(samples)
        samples = samples[order]
```
breaks with merged data sets, as in #189

anesthetic/samples.py
Stefan-Heimersheim: I have now written up the details with a bunch of plots [here](https://github.com/Stefan-Heimersheim/fastCI), in particular here is the interpolated CDF that is used to determine the interval:
CDF
What I meant with "without interpolation" was more like "without KDE", in fact I am using an interpolation of the CDF. I will adjust the description

anesthetic/samples.py
Stefan-Heimersheim: > This is sufficiently general that it deserves its own anesthetic.utils function

Agree, done

> and to be unified with the anesthetic.utils.quantile treatment.

Agree, but not done yet

anesthetic/samples.py
```diff
@@ -8,6 +8,8 @@
 import pandas
 import copy
 import warnings
+from scipy.interpolate import interp1d
+from scipy.optimize import minimize_scalar
```
lukashergt, line 12: Hi Stefan, going on the `CI / lint details` gives you the error message:

> anesthetic/samples.py:11:1: F401 'scipy.interpolate.interp1d' imported but unused
> anesthetic/samples.py:12:1: F401 'scipy.optimize.minimize_scalar' imported but unused

Are you using `interp1d` and `minimize_scalar` anywhere in `anesthetic/samples`? If not, these lines should be removed.

Stefan-Heimersheim: Thanks @lukashergt !

anesthetic/samples.py
Stefan-Heimersheim: Done

anesthetic/samples.py
williamjameshandley: This is an excellent description. I've added a link to the docstring.

williamjameshandley (approved):
Hi @Stefan-Heimersheim. 

I've unified the CDF functions between quantile and credibility_interval. My weighted ECDF was slightly different from yours (defined by min sample has CDF 0, max sample has CDF 1, and the intermediate points have a symmetric average of the weights either side). This coincides with the unweighted form and is symmetric, but I also have some ideas as to how to extend so that the bounds are appropriately slightly larger than the min and max.

I also added some tests so that the coverage is 100%.

If you're happy with these changes please squash and merge.

Best,
Will

Stefan-Heimersheim:
> My weighted ECDF was slightly different from yours (defined by min sample has CDF 0, max sample has CDF 1, and the intermediate points have a symmetric average of the weights either side).

I am a bit unsure about the new ECDF function:
CDF_illustration
Looking at say CDF(10)-CDF(3) in this example, there are 7 out of 10 samples in the interval [3,10], but the ECDF (blue line) gives a larger value -- is that a more-correct approximation that the version I initially implemented (green/red lines)?

williamjameshandley:
Excellent plot!

I've done some investigating, and these really are second order effects. What we should be most concerned about is whether cdf(b)-cdf(a) approximates the true probability mass in between b and a.

```python
import numpy as np
from scipy.interpolate import interp1d
from scipy.stats import norm

# Generate a dataset
np.random.seed(3)
data = np.random.randn(30)
weights = np.random.rand(30)
i = np.argsort(data)
data = data[i]
weights = weights[i]/weights.sum()

# Symmetric empirical CDF
w_ = np.concatenate([[0.],weights[1:]+weights[:-1]]).cumsum()
w_/=w_[-1]
sym = interp1d(data, w_)

# Left hand empirical CDF
lh = interp1d(data, np.cumsum(weights))

# Right hand empirical CDF
rh = interp1d(data, 1-np.cumsum(weights[::-1])[::-1])

plt.plot(data,sym(data), label='Symmetric')
plt.plot(data,lh(data), label='Left Hand')
plt.plot(data,rh(data),  label='Right Hand')
plt.plot(data,norm.cdf(data),  label='True CDF')
plt.xlabel('x')
plt.ylabel('CDF')
plt.tight_layout()
plt.savefig('CDF.png')
plt.close()
```

CDF

So we see that all perform pretty equivalently relative to the correct answer.

Increasing the number of samples improves performance, but the three empirical curves lie on top of one another, which is different at lower order to the true answer.

CDF

Examining the histograms of the performance relative to the true answer shows them to all perform in an unbiased way

```python
W, SYM, LH, RH, T = [], [], [], [], []
for _ in range(10000):
    a, b = np.sort(np.random.uniform(data.min(),data.max(),size=(2)))
    w = weights[(data < b) & (data > a)].sum()
    W.append(w)
    T.append(norm.cdf(b)-norm.cdf(a))
    SYM.append(sym(b)-sym(a))
    LH.append(lh(b)-lh(a))
    RH.append(rh(b)-rh(a))

W, SYM, LH, RH, T = np.array([W, SYM, LH, RH, T])
        
plt.hist(W-T,bins=100)
plt.hist(SYM-T,bins=100)
plt.hist(LH-T,bins=100,alpha=0.8)
plt.hist(RH-T,bins=100,alpha=0.8)
plt.tight_layout()
plt.savefig('hist.png')
```
hist

Increasing to 1000 points just reduces the spread, but not the difference in the methods.

hist

Unless you can find a stress test which shows any of these to be better or worse than the others, I'd vote for the symmetric one.

Stefan-Heimersheim:
I am confused, clearly in my CDF plot there symmetric method always gives you a steeper gradient, and thus should give always give you a larger CDF(b)-CDF(a) difference.
previous
And here is your plot again without subtracting the True value:
```python
plt.figure(constrained_layout=True)
plt.hist(SYM,bins=100, histtype="step", label="SYM")
plt.hist(LH,bins=100, histtype="step", label="LH")
plt.hist(RH,bins=100, histtype="step", label="RH")
plt.hist(T,bins=100, histtype="step", label="T")
plt.xlabel("CDF(b)-CDF(a)")
plt.legend()
plt.savefig('hist.png')
plt.show()
```
hist
(I did set the weights to `weights = np.random.ones(10)`, I didn't get to double check if that is causing any difference in any of the codes.) Here is the same plot including `T`:
hist

Does the symmetric one do systematically worse? Do both methods have a systematic bias? And what would we expect in theory? Intuitively the LH/RH versions feel like implementations of the basic CDF definition, but I am not sure how to reason about the symmetric formula. I'll try to generalize this benchmark and figure out why we don't see that in your histogram.

williamjameshandley:
I knew I had thought about this before -- In #58, I suggest the use of the [weighted percentile method](https://en.wikipedia.org/wiki/Percentile#The_weighted_percentile_method). Turns out there's no universally agreed approach, but the recommended symmetric (C=1/2) version seems to be the best.

The python one-liner is 
```python
wpct = interp1d(data, (weights.cumsum()-weights/2)/weights.sum())
```

Could you add this option to your plots?

Stefan-Heimersheim:
Sure, here is the case of equal weights -- `wpct` is the average between LH and RH, i.e. the same shifted again:
Figure_1
And thus gives identical results to RH & LH:
hist

But in the case of non-equal weights it looks different, I am not sure yet if all the functions handled the weights correctly or if the differences make sense -- I will double check this as soon as possible.
Figure_2

Stefan-Heimersheim:
I'm just working on checking which method makes sense here. First a correction: The plot in my `fastCI` repo only looked interpolated because I only passed the data points as x values to `plt.plot`, the true sample-based CDF looks more like
this
(This is from a [new branch](https://github.com/Stefan-Heimersheim/fastCI/tree/CDF_approximations) I am working on.)

## Now to the comparison:
* I expected the symmetric `sym` method to over-estimate the CDF based on the figure [a few posts above](https://github.com/williamjameshandley/anesthetic/pull/188#issuecomment-1203561092). `sym` looks identical to the current `anesthetic.utils` implementation.
* I expected the discrete method (image above with the orange step-function-y line) to underestimate the inverse-CDF, as it can always find an interval that _just about_ encloses enough points.

Reminder: These are our competitors:
intuition

Now here come the benchmarks, always using a Gaussian (was easy) with `loc=0, scale=1` (scaling doesn't matter). Runs with 10, 100, and 1000 data points for randomized and uniform weights.
CDF_comparison10False
CDF_comparison100False
CDF_comparison1000False
CDF_comparison10True
CDF_comparison100True
CDF_comparison1000True
And here as a function of number of data points (with a few more steps):
function_of_Ndata
(Note that the only random seed is at the start of the file, I do not know why the neighboring N_data points seem correlated.)

Clearly the symmetric / anesthetic version over-estimates the CDF as expected, but I can't see a trend in the discrete case.

williamjameshandley:
Very impressive! The first figure makes it pretty clear that `sym` version is not suitable, and it does seem that it is the interpolation introducing the bias. 

Is there a discrete version of `sym`/the current anesthetic implementation? (I have a sneaking suspicion that this would be identical to wpct, which would be the one I'm leaning toward at the moment).

Stefan-Heimersheim:
> I would not read too much into the 'correlation' unless it remains upon doubling the density of N_data points.

That's easy, here's the same with `N_data *= 2`
function_of_Ndata

I don't think I can do chi-by-eye here but it doesn't look too concerning.

Stefan-Heimersheim:
Here's the error on a) left CI boundary, b) right CI boundary, and c) credibility interval size, always method values minus true theory value. Curiously _all_ methods (sym, lh, wpct, discrete) seem to underestimate the interval sizes, this is unexpected. `N_data=100`, randomized weights. Code [here](https://github.com/Stefan-Heimersheim/fastCI/blob/test_CDF_functions/CDF_benchmarks/test_CI.py).

bias


size

Stefan-Heimersheim:
Here are 20 example data realizations with CIs from the different methods (color code as [here](https://user-images.githubusercontent.com/40799217/185641693-b6e9cfa8-246d-4b3d-84a7-15580bb39ee9.png)):
example_intervals

I suspect the performance of the discrete method depends strongly on the chosen `level`. With 100 data points I suspect it will do well if `100*level` is an integer vs if `100*level+0.5` is an integer, or vice versa. Whereas the other methods should depend less on the precise value of `level`. I'll test that tomorrow.

Stefan-Heimersheim:
Here's my intuition and reasoning about the 4 CDF cases we are considering. Excuse the anthropomorphisms, I found describing this as a race to find for the shortest interval easier to imagine. The 4 cases were

* Discrete CDF / `dcdf`: Green lines here, step-function-y CDF as shown [here](https://github.com/Stefan-Heimersheim/fastCI/blob/CDF_approximations/illustrations/CDF_illustration.png?raw=true)
* LH / RH: Green and red lines [here](https://user-images.githubusercontent.com/40799217/185641253-cf52ec05-56c0-40c5-a2a9-da3ada09b350.png)
* WPCT: Wikipedia weighted percentile with C=1/2 [here](https://user-images.githubusercontent.com/40799217/185641253-cf52ec05-56c0-40c5-a2a9-da3ada09b350.png)
* SYM / `anest`: Symmetrical CDF as implemented in anesthetic, dashed line [here](https://user-images.githubusercontent.com/40799217/185641253-cf52ec05-56c0-40c5-a2a9-da3ada09b350.png)

Here is my intuition for the shape of these CDFs. Always assuming weights to be normalized. When noted assuming the simplified case of equal and normalized weights `w`, and sometimes considering the specific example with `N_data=10` and `w=0.1`. All CDFs go from CDF(-inf)=0, to CDF(inf)=1.

* DCDF: Step function jumping by `w_i` at the `i`th data point. Highest slope is `(w_i + w_{i+1})/distance` between neighboring points, as you can include both of the jumps in the interval. Generally the minimal interval for any credibility will always "just about" include the edge data points. In the case of equal weights `w=0.1` DCDF will always have an advantage of `0.1` probability level over other methods such as LH, RH, and WPCT except in exceptions (see below).
* LH/RH: Stepwise-linear function. The slope in any section is determined only by the weight of the data point right/left of that section. `Slope12=w2/distance` (LH) or `Slope12=w1/distance` (RH). (`Slope12` is the section in between the 1st and 2nd data point).
* WPCT: Stepwise-linear function. The slope in any section is given by the average of the weights of the data points left and right of that section [1]. `Slope12 = (w1+w2)/2/distance`.

Exception: LH, RH and WPCT also get jumps but only at edge points. LH (WH) gets a jump of `w1` (`wN`) at the left (right) edge, and WPCT has jumps of `w1/2` and `wN/2` at the left and right edges respectively.

* SYM: Same as WPCT, but without any jumps, and all slopes scaled by a factor of `2/(w1+2w2+3w3+...+wN)`, i.e. by a factor of `10/9` in the `w=0.1` case [2]. So it's like the non-jump-part of WPCT but stretched out in y-direction to reach 0 and 1, respectively.

I think In the case of uniform weights, DCDF should be the same as LH/RH/WPCT for `level` rounded down to the nearest sum of weights (except for edge-including-cases). E.g. for `w=0.1` DCDF(0.38)=WPCT(0.3).

Footnotes:
> [1] To see this consider the equation, `weights.cumsum()-weights/2`. The first point will be w1/2, the 2nd point will be w1+w2/2 and so on, the difference between points 1 and 2 is (w1/2+w2/2).
> [2] `w_ = np.concatenate([[0.],weights[1:]+weights[:-1]]).cumsum(); w_/=w_[-1]` --> [0, w1+w2, w1+2\*w2+w3, ...,c]/c, so the difference between point i and i+1 is w_i+w+{i+1} like for wpct but normalized differently by c=w1+2*w2+2*w3+...+wN<2, slightly smaller than `2`, the normalization for wpct.

## Examples:

Assume uniform `w=0.1`. For any value betweem `level = 0.1` to `0.2`, DCDF just selects the two closest points and thus finds the smallest possible distance (except for edge-including-cases). DCDF can just achieve the largest slope out of all the CDFs.
LH and WPCT have parallel CDFs here (no edges) and give us a result as expected.
SYM behaves similarly (except for edge-including-cases) to WPCT but has the scaling factor advantage, I think that  (except for edge-including-cases) the SYM result is WPCT(level*9/10).

Figure_1

Now consider a case where the interval includes the edges:
Figure_2

In this `level=0.85` case, the CI includes an edge point. This means LH gets to include a 0.1 jump and produces a shorter interval than DCDF (going from point 1 to halfway between 8 and 9, whereas DCDF has to go to point 9). The WPCT interval includes the edge as well so gets to include the 0.05 jump, so the interval from point 1 to point 9 provides the remaining 80%. SYM looks like a level 76.5 no-edge WPCT interval as expected (factor 10/9 advantage).

Note that WPCT happens to give the same result as DCDF just by accident because the "remaining 80%" are a multiple of 10%. The `level=0.92` example shows WPCT not matching DCDF.

Figure_3

Well I just realized that the minimization is failing in a couple of cases for DCDF. `sop.minimize_scalar(distance, bounds=(0, 1-level), method="Bounded")` as well as `sop.minimize(distance, (1-level)/2, bounds=[(0,1-level)], method="Nelder-Mead")` seem to fail in some (different) cases. It should be trivial solve this as we know the shape of the function, and know where is should be flat. I.e. we can in theory manually check `invCDF([0.05, 0.15, 0.25 etc.])` to find the minimum.

Stefan-Heimersheim:
Fixed DCDF now, last 2 figures again:
Figure_2
Figure_3

Stefan-Heimersheim:
Okay so it looks like all methods underestimate the credible interval size for these Gaussian examples with uniform weights and `N_data=10`. The behavior for DCDF is as expected.
level-dependent_size_error
level-dependent_boundary_error
And same for randomized weights:
level-dependent_size_error_v2

Stefan-Heimersheim:
We just had a discussion, with Will and David, how we can get a better estimate of the CDF. The key is that the CDFs I drawn above all assign a fixed step in the CDF to each point, most saliently visible in the DCDF function:
img

However it would be great to quantify the uncertainty / distribution of possible CDFs. To realize this we considered that any sample drawn from a PDF corresponds to uniform sample from the CDF y-axis. Before explaining this, note that we assume uniform weights here.

> 1. We think integer weights could be modeled as multiple points with uniform weights and
>2. We think arbitrary weights could be arbitrarily well approximated by many points with integer weights.
> 
> (1) is how e.g. Metropolis Hastings MCMC works, and (2) seems like it would follow, I am not 100% sure if many samples with smaller weights are equivalent to few samples with larger weights, since getting multiple samples at the exact point usually does not happen, but we can deal with this later.

---

Drawing (uniform) `x` samples from a PDF is equivalent to drawing uniform samples `s` in [0,1] and calculating  `x=invCDF(s)`. (You can proof this simple with the probability "chain rule" the PDF of a function `f(x)`, `PDF(f) = d/df CDF(f) = d/dx dx/df CDF(x) = PDF(x) / f'` and insert `f(x)=CDF(x)`.)
truth

So imagine you try to approximate the function `invCDF(s)`, you have drawn `N` unknown samples of `s` (uniform and independent), and get out the `N` numbers `x=f(s)`. The only information you have is that `invCDF` is monotonic.
What we are currently doing is assuming `s=linspace(0,1,N)` and plot the interpolation.
theory
But there is no reason to assume this, and in fact we know what `s` is -- `s` is just a set of `N` samples of `Uniform[0,1]` -- we can sample from this. Notably, we are not generating samples and computing (a new) DCDF, but sampling from the space of functions `CDF(x)` (or `invCDF(x)`).
samples_0
Here are a 1000 different iterations of this:
samples_1
We can plot a linear interpolation of these, although note again that linear interpolation is not giving an unbiased representation.
samples_2
See for example how `CDF(0.8)` is no longer fixed to almost 1 but rather uncertain. In fact we can draw the probability distribution for `CDF(0.8)`, it is simply the distribution of "the largest value of `N` samples of `Uniform[0,1]`".

---

We considered the distribution of the "gaps" between the `s` samples, they should be Dirichlet distributed (`N+1` samples that sum to 1).
samples_gaps

To calculate the distribution of the CDF at the `n`th point (in the weighted case, weights `w_i`), consider it's value from the sum of "gaps" sized `g_i`. Generalizing the idea of integer samples "just repeat the sample `w` times" we suspect the CDF sample should be given by `CDF(x_n) = sum(i=0..n) g_i * w_i`. To be continued ...

Stefan-Heimersheim:
> One thing which occurred to me over the weekend as a potential gotcha with the "just repeat the sample `w` times" approach is that this does depend on the normalisation, i.e. you would get different results with the weights `[1,1,3,5,1,1]` in comparison with `[100,100,300,500,100,100]`, the latter having less noise.

Yeah, this is what I was thinking of too when I mentioned my uncertainty here!

> I am not 100% sure if many samples with smaller weights are equivalent to few samples with larger weights, since getting multiple samples at the exact point usually does not happen

So just to write down my intuitions for both cases (A,B) here:

A) _Multiple samples != few weighted samples:_ Imagine a histogram of 1,000 vs 10,000 samples -- you can pretty clearly draw the PDF, CDF, and read of intervals. But also clearly the accuracy of these will scale with having more samples, so 1,000 samples with weight=10 are not the same.

B) _Oh the other hand,_

1. anesthetic's functions with the `ncompress` option resample uniformly weighted samples from an arbitrarily weights doesn't it? Although looking at the code it returns at most `len(samples)` new samples, so indeed it does not artificially increase the number of samples? ` I'm not entirely sure how `sample_compression_1d()` works if `ncompress=len(x)` and weights are strongly non-uniform, does this (`utils.py:385`) return the correct result?
```python
        x_ = np.random.choice(x, size=ncompress, replace=False)
```

2. MH MCMC returns the same parameter point multiple times in ~70% of cases (if acceptance ~30%), we are usually happy with these? Although we would say this gives a larger autocorrelation length and the effective number of samples is lower? But doesn't using chains with or without "thinning" give you the same result?

---

Anyway, before solving this issue I wanted to focus on the path from this to getting credibility intervals. From the figure above you can -- at least empirically -- derive the probability distribution for CDF(x_i). In between there will necessarily be interpolation (with the convexity/concavity problems discussed above) because we have random samples at fixed positions x_i.
Potentially you could look at it the other way, invCDF(y) is sampled at random positions y but only giving fixed values x_i, this seems maybe more difficult to treat.
repeat

To get a credibility interval we need CDF(b) - CDF(a). In the case of a and b aligning with sampling points, conditional on a given value CDF(a) we can get the distribution of CDF(b) from order statistics arguments (such as [Will's Appendix A2](https://arxiv.org/abs/1506.00171), or potentially your [Wikipedia link](https://en.wikipedia.org/wiki/Order_statistic)). We can (potentially analytically) marginalize over CDF(a) and get a distribution for CDF(b)-CDF(a), i.e. the size of that interval. More generally we could do this for arbitrary points using interpolation and marginalizing over the up to 4 relevant sampling positions.

Generally this will always give us a distribution & uncertainty on the volume of an interval "[3,7] is a 95%+/-3% interval" rather than a uncertainty on the size "[3+/-0.1, 7+/-0.2] is the 95% interval of this distribution". In theory we should be able to translate between these two measures, the same way we could translate from a distribution on f given x to a distribution on x given f in [fgivenx](https://github.com/williamjameshandley/fgivenx).

However I don't think this is necessary. The goal of this code is mainly to be used for large `N_samples` and all this optimization for ~10 samples is not too useful. If a user with 1,000 samples gets the output "[3,7] is the 95%+/-0.4% CI" that's good enough for most cases, and importantly, transparent and with quantified uncertainty.

So I could think of proposing the algorithm

1. Find a guess for the desired interval (say level=95%)  [a,b] using say `wcpt`.
2. Find the distribution of CDF(b) - CDF(a) based on analytical or empirical CDF distributions.
3. Return [a,b] is a 95%+/-0.2% confidence interval, maybe with a warning if the uncertainty is >~ 0.1*(1-level).

Because this should be achievable and I expect [a,b] obtained this way to be close enough (for large N, for practical purposes) to the optimal "[a+/-sigma_a, b+/-sigma_b] = true 95% CI" result.

williamjameshandley:
>Generally this will always give us a distribution & uncertainty on the volume of an interval "[3,7] is a 95%+/-3% interval" rather than a uncertainty on the size "[3+/-0.1, 7+/-0.2] is the 95% interval of this distribution".

We should be able to get either version from a sampling distribution:
1. Assume you know the step heights of the CDF (we have been doing this all along with equal steps)
2. Compute the credibility bounds [a, b] for these step heights as before (e.g. using the current minimisation scheme)
3. Repeat step 1 and 2 over samples drawn from the distribution of CDF values above for nsamples iterations (in practice ~O(12) is usually enough to get a median and spread)
4. you now have a set of [a, b] value for each of the samples from the CDF. These are samples drawn from the distribution, so taking a mean and std of these gives you the "[3+/-0.1, 7+/-0.2]" you're looking for"

You can of course reverse this process, changing [a, b] in the above for 'probability mass between fixed [a, b]' which gives you.

The whole premise of "thinking in samples" is that if you can do a forward analysis to get an output for a fixed set of inputs, then sampling those inputs and repeating the forward analysis to get a set of outputs gives you the sampling distribution of outputs. See [slide 8](https://github.com/williamjameshandley/talks/raw/reach_2022/will_handley_reach_2022.pdf), or [3.1](https://www.cambridge.org/core/books/bayesian-methods-in-cosmology/D7076A1687938B4E4ECF514B94746C94) for more detail.

Stefan-Heimersheim:
> Repeat step 1 and 2 over samples drawn from the distribution of CDF values above for nsamples iterations (in practice ~O(12) is usually enough to get a median and spread)

I see what you mean -- so let's say I do this 12 times with a low-number-of-samples-dataset, using CDFs sampled like above, think points like these:
image from before
Then I might get the results
```
[3,7] [3,7] [3,7] [1,7] [3,7] [5,9] [2,5], [3,7], [3,7], [3,7], [3,7], [3,7]
```
and compute [3+-..., 7+/-...].

Notable I would not use linear interpolation as it is systematically biased (tailed distributions always curve the same way), so the distribution would be quite discrete which might occasionally lead to many samples being equal / make it harder to compute mean and variance -- but that's fine until we run into div by 0 / all numbers identical errors, and then we deal with it.

Secondly (and this is not a bug either but a feature) it might happen (for particularly flat distributions) that we get things like `[3,7]`, `[4,8]`, `[5, 9]` which all are 68%+/-0.1% still leading to a big uncertainty 4+/-1ish. Which is true, the uncertainty on HPD-interval-boundary *is* large. But this would be an argument in favour of using errorbars on the interval size rather than the interval boundary.

---

Anyway, now to implementing something in practice ...

Stefan-Heimersheim:
Let me also note down my thoughts on

> One thing which occurred to me over the weekend as a potential gotcha with the "just repeat the sample w times" approach is that this does depend on the normalisation, i.e. you would get different results with the weights [1,1,3,5,1,1] in comparison with [100,100,300,500,100,100], the latter having less noise.

We should ask ourselves, what do we mean by integer weights? In a typical MCMC it means we have drawn multiple samples at this point. And it holds that the chance of drawing a sample at any point is proportional to the PDF at that point.

Those samples are not independent, so samples do not follow things like the distance-distribution P(distance between 2 random samples). P(2 identical random samples) == 0 but it does occur in MCMC. Is that a problem for our approach above?

previous image
Empirically thinking of the usual image, think of two points at the same point. The y-distance between the CDF samples is a finite (non-zero) value, following some kind of distribution given by order statistics. Thus if we ask for the, say, 1% confidence interval, it might give a zero-sized interval (e.g. [3,3]) sometimes. Now the question is, is that correct? Let's say 5 times we get [3,3] and 2 times we get [3,4]. In the first 5 times the code will return `[3,3] is the 0+/-0% credibility interval` which is correct, and the user will notice they might be asking for an interval too small given their samples. Now in the other two cases the code might return `[3,4] in the 5+/-4% credibility interval`. Again, I think this would be correct and the user would notice what's up.

PS: In the latter `3,4` case is `5+/-4%` computed with the "left" or "right" sample of `3`? Hmm, this is not clear but clearly at most one version can be correct. This problem does not occur if we go with @williamjameshandley's approach for computing `[3+-..., 7+/-...]`-style intervals. So that might be the better option.

---

Let's briefly think of weights [100,100,300,500,100,100]. This would lead to a bunch of "lines of green dots" (as pictured above) bunched up together, and would make the jumps in the CDF much more pronounced. If we were to think of corresponding PDFs (take some interpolation) those could look rather something like this (artist's impression) rather than a Gaussian:
screenshot_2022-10-11_13-07-20
I'm not sure what to make of this, given that MCMC could plausibly give you `[100,100,300,500,100,100]` for a Gaussian.

---

Let's think of how to incorporate fractional weights (which should also solve the issue above). @williamjameshandley mentioned the gaps-sampling idea a few before. The gaps here
gaps image
are sampled from _some_ distribution, and we should be able to generalize this for non-uniform gaps. Important considerations are

* Gap size distribution should somehow scale with weight(s) of sampling points, maybe even be proportional.
* A gap is between two points, see all the discussion on left/right-sided CDF, `wpct` etc. we had above
* The sum of gaps must equal 1, or rather CDF(-inf) must be 0 and CDF(inf) must be 1.
* It's probably related to Will's order statistics things linked above

I'll give this a thought over lunch -- to be continued.

Stefan-Heimersheim:
For uniform weights, the gap distribution seems to be indeed Dirichlet (in particular, with `alpha_i=1`). I think we also tested that when we chatted; here's a plot checking the marginal distribution (M=10 samples). The marginal distribution of one variable of the the Dirichlet is a Beta distribution so `gap_i ~ sst.beta.pdf(gap, a=alpha_i, b=ngaps-alpha_i)`.
```python
import numpy as np
import scipy.stats as sst
N = 100000 #empirical test size
M = 10 #variables
data = np.random.uniform(0,1,size=(M, N))
sorted = np.sort(data, axis=0)
diffs = np.diff(sorted, axis=0)
gaps = np.array([sorted[0], *diffs, 1-sorted[-1]])
ngaps = M+1
plt.hist(gaps.T, bins=1000, histtype="step", density=True)
x = np.linspace(0,1,1000)
f = sst.beta.pdf(x, a=1, b=ngaps-1)
plt.plot(x,f, color="k", lw=2)
plt.show()
```
Figure_1

Just for reference, the sum of two gaps (i.e. gaps like this where we treat weight=2 as 2 samples)
Figure_3
seems to be given by a Dirichlet distribution with alpha_i=weight_i=2.
```python
plt.hist(gaps[6]+gaps[7], bins=1000, histtype="step", density=True)
f = sst.beta.pdf(x, a=2, b=ngaps-2)
```
Figure_2
which suggests that maybe we can do something along the lines of alpha = weights at some point.

---

But the fundamental question is still what do we do with samples (say x=3) occurring twice in the data? Earlier I wrote that this is a problem only if asking for the confidence interval [3,...] or [...,3] but actually it also affects "@williamjameshandley's approach".

> PS: In the latter 3,4 case is 5+/-4% computed with the "left" or "right" sample of 3? Hmm, this is not clear but clearly at most one version can be correct. This problem does not occur if we go with @williamjameshandley's approach for computing [3+-..., 7+/-...]-style intervals. So that might be the better option.

In that one, do you give the "jump" at 3 to both, [3,...] _and_ [...,3]? Does it bias the confidence interval estimates even after we do all this random sampling on the y axis? (making intervals smaller / credibilities larger) What *is* the confidence interval here?

Figure_4

Stefan-Heimersheim:
Okay I just had a chat with @yallup and we concluded that it seems very hard or impossible to work with repeated samples, and to that extend probably also with non-uniform weights.

Fundamentally a "weights=100,200,500,200,100" case as mentioned before does behave fundamentally different in this CDF picture to a "weights=1,2,5,2,1" case, which it should not. In particular, the former will have much more certain less random steps in the CDF, making the jumps much stronger. Here's an example for 10 uniform samples vs the same samples with weight 10 each:
Figure_A
Figure_B

---

So since we fundamentally cannot deal with (integer) weights, or at least, get different results based on the normalization, we just limit ourselves to uniform weights. @yallup pointed out that we can just use anesthetic's compress function to get uniform-weight samples from any data set, so we can call this in `credibility_interval`.

I will implement this.

anesthetic/utils.py
```diff
@@ -89,6 +90,148 @@ def quantile(a, q, w=None, interpolation='linear'):
...
+    samples = np.array(samples)
+    samples.sort()
+    ngaps = len(samples)-1
+    gaps = np.random.dirichlet(np.ones(ngaps))
+    cdf = np.array([0, *np.cumsum(gaps)])
+    assert np.isclose(cdf[-1], 1, atol=1e-9, rtol=1e-9), \
+        "Error: CDF does not reach 1 but "+str(cdf[-1])
+    # Set the last element (tested to be approx 1)
+    # to exactly 1 to avoid interpolation errors
+    cdf[-1] = 1
```
lukashergt, line 104: Would it be more appropriate to normalise as follows?
```python
cdf /= cdf[-1]
```

Or would it be better to handle bound errors within `interp1d`? (see below)

anesthetic/utils.py
```diff
@@ -89,6 +90,148 @@ def quantile(a, q, w=None, interpolation='linear'):
     return quant
 
 
+def sample_cdf(samples, inverse=False, interpolation='linear'):
+    """Sample the empirical cdf for a 1d array."""
+    samples = np.array(samples)
+    samples.sort()
```
lukashergt, line 96: Should be possible to combine these two lines into
```python
samples = np.sort(samples)
```

anesthetic/utils.py
```diff
@@ -89,6 +90,148 @@ def quantile(a, q, w=None, interpolation='linear'):
...
+    cdf = np.array([0, *np.cumsum(gaps)])
+    assert np.isclose(cdf[-1], 1, atol=1e-9, rtol=1e-9), \
+        "Error: CDF does not reach 1 but "+str(cdf[-1])
+    # Set the last element (tested to be approx 1)
+    # to exactly 1 to avoid interpolation errors
+    cdf[-1] = 1
+    if inverse:
+        return interp1d(cdf, samples, kind=interpolation)
+    else:
+        return interp1d(samples, cdf, kind=interpolation)
```
lukashergt, line 108: Should we add `bounds_error=False, fill_value=(0, 1)` for the interpolation from samples to CDF?

Would `fill_value=(min(samples), max(samples))` work for the inverse?

anesthetic/utils.py
```diff
@@ -89,6 +90,148 @@ def quantile(a, q, w=None, interpolation='linear'):
...
+                                                      ddof=1),
+                    np.mean(ci_samples[:, 1]), np.std(ci_samples[:, 1],
+                                                      ddof=1)))
+        if return_covariance:
+            return np.mean(ci_samples, axis=0), \
+                   np.cov(ci_samples, rowvar=False)
+        else:
+            return np.mean(ci_samples, axis=0)
+    else:
+        raise ValueError('ci_samples in unregonized shape')
```
lukashergt, line 232: typo: unregonized -> unrecognised

lukashergt (commented):
Hi @Stefan-Heimersheim,

nice work. See some comments inline. 

It might be nice to also add this as a method of our `anesthetic.samples.Samples` class (with automatic passing of weights etc.).

For code coverage to pass the various cases `'ll'`, `'ul'`, `'et'` need to be tested specifically.

Stefan-Heimersheim:
Thank you @lukashergt for the great feedback, I plan to implement & answer those points within the next two weeks!

anesthetic/utils.py
```diff
@@ -89,6 +90,148 @@ def quantile(a, q, w=None, interpolation='linear'):
...
+        # compress_weights with ncompress<=0 assures weights \in 0,1
+        # Note that this must be done, we cannot handle weights != 1
+        # see this discussion for details:
+        # https://github.com/williamjameshandley/anesthetic/pull/188#issuecomment-1274980982
+        weights = compress_weights(weights, ncompress=-1, u=u)
+        if verbose:
+            print("Compressing weights to", np.sum(weights),
+                  "unit weight samples.")
+        assert np.all(np.logical_or(weights == 0, weights == 1)), \
+               "Unexpected error in compress_weights, weights not binary"
```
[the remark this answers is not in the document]
Stefan-Heimersheim: I have never seen it fail but wanted to make sure -- Good suggestion, I added a test to `test_compress_weights` and removed the assertion.

anesthetic/utils.py
```diff
@@ -89,6 +90,148 @@ def quantile(a, q, w=None, interpolation='linear'):
...
+    cdf[-1] = 1
+    if inverse:
+        return interp1d(cdf, samples, kind=interpolation)
+    else:
+        return interp1d(samples, cdf, kind=interpolation)
+
+
+def credibility_interval(samples, weights=None, level=0.68, method="hpd",
+                         u=None, n_iter=12, return_covariance=False,
+                         verbose=False):
```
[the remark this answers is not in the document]
Stefan-Heimersheim: Good point, updated the docstring!

I would rather not add `interpolation_kwargs` because the interpolation should not matter (thousands of data points). In any case where the interpolation does matter (10-ish points) the method is going to be wrong in any case, and there exists no _good_ interpolation one could choose.

anesthetic/utils.py
```diff
@@ -89,6 +90,148 @@ def quantile(a, q, w=None, interpolation='linear'):
...
+          This only works if the distribution is unimodal.
+        * ``'ll'``/``'ul'``: Lower/upper limit. One-sided limits for which
+          ``level`` fraction of the (equally weighted) samples lie above/below
+          the limit.
+        * ``'et'``: Equal-tailed interval with the same fraction of (equally
+          weighted) samples below and above the interval region.
+
+    u : array, optional
+        Random values to use for reproducible sample compression.
+        Default: ``np.random.rand(len(weights))`` if non-unit weights used.
```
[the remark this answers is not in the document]
Stefan-Heimersheim: > Should the random value array u be part of this function? I feel like this is combining two things (sample compression and computation of the C.I.) into one function, where it would make more sense to handle them separately...?

The only reason for this to be here was to allow to get reproducible outputs, i.e. pass through the `u` argument from `compress_samples`. But I realize that `np.random.seed(0)` before the function call already does this so the `u` is not necessary; deleted.

anesthetic/utils.py
```diff
@@ -89,6 +90,148 @@ def quantile(a, q, w=None, interpolation='linear'):
...
+    n_iter : int, default=12
+        Number of CDF samples to improve `mean` and `std` estimate.
```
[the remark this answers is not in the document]
Stefan-Heimersheim: You're right, done!

anesthetic/utils.py
Stefan-Heimersheim: Honestly it doesn't matter since ` atol=1e-9`, this is just dealing with python float precision. So it's stick to the simpler version to avoid looking like we're doing some actual normalisation

Stefan-Heimersheim:
> For code coverage to pass the various cases 'll', 'ul', 'et' need to be tested specifically.

Right! Added tests & gave everything better names (`'hpd'->'iso-pdf'`, `'ll->'lower-limit'` etc.)

anesthetic/utils.py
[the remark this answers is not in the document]
lukashergt: I agree that for a lot of samples the interpolation method hardly matters. But a lot of the discussion above was specifically about exploring different interpolation options. So I would suggest to pick your favourite as default, but give the user full flexibility. For one, this would allow the user to at least try different interpolation methods to get a sense of how much that might affect the results.

anesthetic/utils.py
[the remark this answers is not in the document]
lukashergt: I agree that we don't have precise knowledge of the population. But I think we can make our results more robust to variability coming from the sample compression (now that we got rid of the `u`). If we move the line
```python 
weights = compress_weights(weights, ncompress=-1)
```
into our iteration loop, then each iteration will use a slightly different subsample, giving us a handle at the (sub-)sample variance (or maybe it is more accurate to call this the compression variance...?).

tests/test_utils.py
```diff
@@ -172,3 +176,59 @@ def test_p_values_from_sample():
 
     ks_results = insertion_p_value(ns.insertion[nlive:-nlive], nlive, batch=1)
     assert ks_results['p-value'] > 0.05
+
+
+def test_credibility_interval():
```
lukashergt, line 181: I'd suggest compactifying `test_credibility_interval` as follows. The parametrised tests ensure we don't miss any cases (currently we miss `return_covarianc==True and 'limit' in method`...).

I also changed the cov tests a bit, making them a bit more permissive such that they are more robust to different seeds and to a different number of iterations. After all, the values you tested cov against were not theory values, right? They just happened to work for that seed and for that number of iterations?

Something along the lines of this:

```python
@pytest.mark.parametrize('method', ['iso-pdf', 'equal-tailed',
                                    'lower-limit', 'upper-limit'])
@pytest.mark.parametrize('return_covariance', [True, False])
def test_credibility_interval(method, return_covariance):
    np.random.seed(0)
    normal_samples = np.random.normal(loc=2, scale=0.1, size=10000)
    ci = credibility_interval(
        normal_samples,
        level=0.84 if 'limit' in method else 0.68,
        method=method,
        return_covariance=return_covariance,
    )
    if return_covariance:
        ci, cov = ci
        assert np.ndim(cov) == 0 if 'limit' in method else 2
        assert np.all(np.abs(cov) < 1e-4)
        assert np.all(np.abs(cov) > 1e-8)
    if method in ['iso-pdf', 'equal-tailed']:
        lower, upper = ci
    else:
        upper = ci if 'upper' in method else 2.1
        lower = ci if 'lower' in method else 1.9
    assert_allclose(lower, 1.9, atol=0.02)
    assert_allclose(upper, 2.1, atol=0.02)
```

tests/test_utils.py
```diff
@@ -172,3 +176,59 @@ def test_p_values_from_sample():
...
+    assert_allclose(CDF(2), 0.5, atol=0.1)
+
+    samples = read_chains('./tests/example_data/pc')
+    mean2, cov2 = credibility_interval(samples["x0"], level=0.68,
+                                       weights=samples.get_weights(),
+                                       method="iso-pdf",
+                                       return_covariance=True)
+    assert_allclose(mean2[0], -0.1, atol=0.01)
+    assert_allclose(mean2[1], 0.1, atol=0.01)
+    assert_allclose(cov2, [[5e-5, 5e-5], [5e-5, 1e-4]], rtol=0.5)
```
lukashergt, line 204: I think it would be cleaner to stick to the `normal_samples` for this test in `test_utils.py`, and put tests using the nested sampling example data into the test in `test_samples.py`.

tests/test_utils.py
```diff
@@ -172,3 +176,59 @@ def test_p_values_from_sample():
...
+                                     return_covariance=True)
+    assert_allclose(mean[0], 1.9, atol=0.01)
+    assert_allclose(mean[1], 2.1, atol=0.01)
+    assert_allclose(cov, [[7e-6, 6e-6], [6e-6, 8e-6]], rtol=0.5)
+
+    # Test non-inverse sample_cdf
+    CDF = sample_cdf(normal_samples, inverse=False)
+    assert_allclose(CDF(-np.inf), 0)
+    assert_allclose(CDF(-100), 0)
+    assert_allclose(CDF(2), 0.5, atol=0.1)
```
lukashergt, line 195: This should go into a separate test `test_sample_cdf`.

lukashergt (commented):
Nice, thanks @Stefan-Heimersheim. A few more points going into finer details inline.

On the topic of sample variance, I read up a bit on bootstrapping, and the Bayesian bootstrap in particular (see e.g. [the corresponding Wikipedia article](https://en.wikipedia.org/w/index.php?title=Bootstrapping_(statistics)&section=11#Bayesian_bootstrap)). Is that essentially what we have implemented?

anesthetic/utils.py
```diff
@@ -89,6 +90,151 @@ def quantile(a, q, w=None, interpolation='linear'):
...
+          lie above/below the limit.
+        * ``'equal-tailed'``: Equal-tailed interval with the same fraction of
+          (equally weighted) samples below and above the interval region.
+
+    return_covariance: bool, default=False
+        Return the covariance of the sampled limits, in addition to the mean
+    nsamples : int, default=12
+        Number of CDF samples to improve `mean` and `std` estimate.
+    verbose: bool, default=False
+        Print information and outputs.
```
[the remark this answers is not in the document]
Stefan-Heimersheim: Removed; mainly useful for debugging

anesthetic/utils.py
```diff
@@ -89,6 +90,151 @@ def quantile(a, q, w=None, interpolation='linear'):
...
+                                                      ddof=1),
+                    np.mean(ci_samples[:, 1]), np.std(ci_samples[:, 1],
+                                                      ddof=1)))
+        if return_covariance:
+            return np.mean(ci_samples, axis=0), \
+                   np.cov(ci_samples, rowvar=False)
+        else:
+            return np.mean(ci_samples, axis=0)
+    else:
+        raise ValueError('ci_samples in unrecognised shape')
```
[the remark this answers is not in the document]
Stefan-Heimersheim: I don't think it _should_ ever not match, but I was cautious of maybe this could break if you come up with something crazy to pass.

Turned it into an else now.

anesthetic/utils.py
```diff
@@ -89,6 +90,151 @@ def quantile(a, q, w=None, interpolation='linear'):
...
+    else:
+        weights = np.array(weights.copy())
+
+    # Convert samples to unit weight not the case
+    if not np.all(np.logical_or(weights == 0, weights == 1)):
+        # compress_weights with ncompress<=0 assures weights \in 0,1
+        # Note that this must be done, we cannot handle weights != 1
+        # see this discussion for details:
+        # https://github.com/williamjameshandley/anesthetic/pull/188#issuecomment-1274980982
+        weights = compress_weights(weights, ncompress=-1)
```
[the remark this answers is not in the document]
Stefan-Heimersheim: We need `ncompress=-1` because the method does not work for weights other than 1. In practice this shouldn't matter because the (original) point of this method is to deal with huge numbers of samples.

anesthetic/utils.py
```diff
@@ -89,6 +90,151 @@ def quantile(a, q, w=None, interpolation='linear'):
...
+        * ``'lower-limit'``/``'upper-limit'``: Lower/upper limit. One-sided
+          limits for which ``level`` fraction of the (equally weighted) samples
+          lie above/below the limit.
+        * ``'equal-tailed'``: Equal-tailed interval with the same fraction of
+          (equally weighted) samples below and above the interval region.
+
+    return_covariance: bool, default=False
+        Return the covariance of the sampled limits, in addition to the mean
+    nsamples : int, default=12
+        Number of CDF samples to improve `mean` and `std` estimate.
```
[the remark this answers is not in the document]
Stefan-Heimersheim: Intuitively, I expect other uncertainties (finite amount of samples) to outweigh the `nsamples=12` error, and this seems to be the case for the two examples below. By default `nsamples=1` was fine, we just added a few more because it was a very low-hanging fruit.

I'm less certain about the covariance returned. But I also have no idea how many `nsamples` are good, I don't want to make this needlessly slow.

Here's `normal_samples` for different `nsamples` values
```
1 samples: [1.90273801 2.09455387]
2 samples: [1.90269773 2.09849627]
4 samples: [1.89775164 2.09207853]
8 samples: [1.90027277 2.09531987]
12 samples: [1.89883413 2.09379086]
120 samples: [1.89997834 2.09506267]
1200 samples: [1.89981016 2.09489933]
```

Here's ./tests/example_data/pc "x0"]
```
1 samples: [-0.1988116   0.18185223]
2 samples: [-0.17449234  0.1926399 ]
4 samples: [-0.17046529  0.17364958]
8 samples: [-0.18633088  0.17586416]
12 samples: [-0.19075288  0.17333953]
120 samples: [-0.18451308  0.17946344]
1200 samples: [-0.18678313  0.1744095 ]
```

anesthetic/samples.py
```diff
@@ -443,6 +443,62 @@ def importance_sample(self, logL_new, action='add', inplace=False):
         else:
             return samples.__finalize__(self, "importance_sample")
 
+    def credibility_interval(self, column, level=0.68, method="iso-pdf",
+                             return_covariance=False, nsamples=12,
+                             verbose=False):
```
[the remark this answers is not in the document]
Stefan-Heimersheim: That _would_ be nice, however I have no idea how to make MultiIndices pandas magic -- would you potentially be happy to implement that?

tests/test_utils.py
Stefan-Heimersheim: This is great, I've replaced my function with this!

> After all, the values you tested cov against were not theory values, right? They just happened to work for that seed and for that number of iterations?

Yep

Stefan-Heimersheim:
Thanks for all the suggestions @lukashergt! I've implemented everything now, except for the MultiIndex. I think that would have to be implemented in `weighted_pandas` similar to the other functions, but I'm really not sure how that works.

anesthetic/utils.py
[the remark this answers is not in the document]
lukashergt: `samples.compress` will always return equally weighted samples (i.e. samples with all weights equal to 1), also for `ncompress=True` or for `ncompress=100000`. 

So are equal weights the only requirement? Or is there the additional requirement that there must not be repeated samples?

If only equal weights are needed, then `ncompress=True` will lead to a more conservative (and hence trustworthy...?) compression. 

If samples are not allowed to repeat, then indeed `ncompress=-1` is needed.

anesthetic/samples.py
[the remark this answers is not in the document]
lukashergt: I'll take a look, but won't have time immediately.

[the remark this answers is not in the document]
lukashergt: Ok, I've had a look at this. This is what things look like at the moment:

image

`credibility_interval` called as a method from a `Samples` instance now returns a dataframe with the limit(s) for all parameters. Depending on `method` the index contains only `lower`, only `upper`, or both. To get the limits from only a subset of the parameters, just call the method from a subset of the dataframe. For `return_covariance=True` a tuple of two dataframes is returned, one for the intervals, the other for the covs. The dataframe of covs is a multi-indexed dataframe.

@Stefan-Heimersheim, @williamjameshandley what do you think?

anesthetic/utils.py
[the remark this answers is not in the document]
lukashergt: With #285 we will need to change `ncompress=-1` to `ncompress='equal'`.

Stefan-Heimersheim:
Thanks @lukashergt for implementing the MultiIndex! That's everything done; as you suggested I'll take a look at the code tonight and we can get this merged.

Stefan-Heimersheim:
I think this is fantastic, thanks for implementing the extension!

The only unintuitive case is calling it for a single variable, you have to do
```
samples[["noise"]].credibility_interval()
```
rather than
```
samples["noise"].credibility_interval()
```
but our other functions like `mean()` and `std()` work in either case. But not a big deal and happy to leave it this way if the extra functionality is a hassle.

lukashergt:
> I think this is fantastic, thanks for implementing the extension!
> 
> The only unintuitive case is calling it for a single variable, you have to do
> 
> ```
> samples[["noise"]].credibility_interval()
> ```
> 
> rather than
> 
> ```
> samples["noise"].credibility_interval()
> ```
> 
> but our other functions like `mean()` and `std()` work in either case. But not a big deal and happy to leave it this way if the extra functionality is a hassle.

`samples.x0.credibility_interval(...)` now returns the same as `credibility_interval(samples.x0, weights=samples.get_weights(), ...)`.

lukashergt:
@williamjameshandley, since @Stefan-Heimersheim and I have both added functionality here, would you like to review?

williamjameshandley (changes requested):

Stefan-Heimersheim:
> I would strongly prefer for there not to be an assert statement

I agree, and we do not need to check `cdf[-1]` because this is literally a sum over `np.random.dirichlet` which by [definition](https://numpy.org/doc/stable/reference/random/generated/numpy.random.dirichlet.html) sums to 1.

lukashergt (approved):
Anything else for this PR, @williamjameshandley?

williamjameshandley (approved):
Many thanks @Stefan-Heimersheim. Please squash and merge.

williamjameshandley:
Should merge #398 first.

PR #192, opened 2022-06-16
Check for NaNs in logL

Ormorod:
Check for NaNs (more generally, null values) in logL values of MCMCSamples.


williamjameshandley:
I'm not convinced this is the best way to deal with NaNs. At the moment, with this change any code (e.g. cobaya) which produced NaN likelihoods would just fail to load, when what users would likely want to do in reality is to drop them as unphysical (and perhaps throw a warning). This would be more consistent with the code block in `NestedSamples.recompute()`, which warns if the birth likelihoods are not sorted correctly. Upon further thought I think that that function is the correct place for this check (rather than MCMCSamples, where users might legitimately want to flag logLs with nans)

anesthetic/samples.py
```diff
@@ -764,6 +764,10 @@ def recompute(self, logL_birth=None, inplace=False):
 
         samples.tex['nlive'] = r'$n_{\rm live}$'
         samples.beta = samples._beta
+
+        if np.any(pandas.isnull(samples.logL)):
+            warnings.warn("Null value encountered in logL.", RuntimeWarning)
```
lukashergt, line 769: I would prefer `np.isnan` over `pandas.isnull` and `NaN` over `Null`, which I think are more pythonic. The pandas "null" comes from R...

lukashergt: Actually, pandas can apparently do more... So I'll revise my comment to: I would prefer `pandas.isna` over `pandas.isnull` and `NaN` over `Null`.

williamjameshandley (dismissed):
OK, I've updated the strategy somewhat so that if it discovers NaNs, it warns the user and then drops those rows. I think that anything else is too dangerous as NaNs royally foul up the nlive calculation. After experimenting with ways to try and do this with NaNs (and failing), I have updated the nlive calculation so that it is slightly more robust and cleaner.

If you're happy with these changes, please squash and merge.

Ormorod:
I'm happy (and agree that it's best to drop the NaN samples).

Just made the small change requested by the linting workflow, and I think the word "samples" would make the error message clearer.

williamjameshandley (approved):

williamjameshandley:
I agree that samples is much clearer. Please squash and merge.

Ormorod:
I haven't got the squash and merge button

Ormorod:
Thanks - my github stats have been suffering as commits on forks don't count 😭

PR #194, opened 2022-07-27
WeightedDataFrame plotting

williamjameshandley:
# Description

This is the first of hopefully several big PRs over the next few weeks to move us closer to anesthetic 2.0 release.

This PR goes some way to resolving #59. 

At the moment, seasoned users of `pandas` would be surprised by the behaviour of weighted data frames, since if they tried to run any of these
```python
df = WeightedDataFrame(...)
df.plot.boxplot()
df.hist()
df.plot(kind='scatter')
df.x.plot.kde()
```
the resulting code would not incorporate the weights correctly.

This PR fixes the above behaviour. It achieves it by overriding the `pandas.plotting._matplotlib` backend, returning the same behaviour for `pd.DataFrames`, but updating the behaviour for `anesthetic.weighted_pandas.WeightedDataFrames`.

At the moment `anesthetic.NestedSamples` and `anesthetic.MCMCSamples` override some (but importantly not all) of the above behaviour, by creating their own `MCMCSamples.plot`. This again is a surprise, and unless @lukashergt, @Stefan-Heimersheim, @htjb, @dominicanstey or any other regular users seriously object, it is my intention to use the major version update to make backward incompatible changes to the `.plot` commands (leaving deprecated versions available for the next few minor version updates). 

# Checklist:


anesthetic/__init__.py
```diff
@@ -11,6 +11,34 @@
...
+
+
+# Override the two places where _get_plot_backend is defined
+pandas.plotting._core._get_plot_backend = \
+        _anesthetic_override(pandas.plotting._core._get_plot_backend)
+pandas.plotting._misc._get_plot_backend = \
+        _anesthetic_override(pandas.plotting._misc._get_plot_backend)
+
+# Set anesthetic._matplotlib as the actual backend
+pandas.options.plotting.backend = 'anesthetic._matplotlib'
```
williamjameshandley, line 40: This code sets up the anesthetic._matplotlib backend as a replacement for matplotlib. In principle it should behave exactly the same as the current matplotlib backend for normal pandas arrays, but behave correctly for `WeightedDataFrame`s

anesthetic/_code_utils.py
```diff
@@ -0,0 +1,53 @@
+def replace_inner_function(outer, new_inner):
+    """Replace a nested function code object used by outer with new_inner.
+
+    The replacement new_inner must use the same name and must at most use the
+    same closures as the original.
+
+    Code taken from https://stackoverflow.com/a/27550237
```
williamjameshandley, line 7: This is a very fancy bit of python taken from the above link. It allows one to monkey patch a function within a function. This is only necessary for modifying the `pandas.plotting._matplotlib.boxplot.boxplot` command, which is infuriatingly written as a function rather than a class.

anesthetic/_matplotlib/__init__.py
```diff
@@ -0,0 +1,43 @@
+"""anesthetic override of matplotlib backend for weighted samples."""
```
williamjameshandley, line 1: This imports all of the things it needs to from the `matplotlib` backend, and replaces the relevant functions and classes with functions defined in `anesthetic._matplotlib`

anesthetic/_matplotlib/boxplot.py
```diff
@@ -0,0 +1,129 @@
...
+
+def _bxpstat(y, weights, whis):
+    q1, med, q3 = quantile(y, [0.25, 0.5, 0.75], weights)
+    bxpstat = {}
+    bxpstat['med'] = med
+    bxpstat['q1'] = q1
+    bxpstat['q3'] = q3
+    bxpstat['whislo'] = q1 - whis*(q3-q1)
+    bxpstat['whishi'] = q3 + whis*(q3-q1)
+    return bxpstat
```
williamjameshandley, line 19: Compute weighted boxplot statistics.

anesthetic/_matplotlib/boxplot.py
```diff
@@ -0,0 +1,129 @@
...
+
+        weights = kwds.pop("weights", None)
+
+        if weights is None:
+            bp = ax.boxplot(y, **kwds)
+        else:
+            whis = kwds.pop("whis", 1.5)
+            kwds['showfliers'] = False
+            y = np.atleast_2d(y)
+            bp = ax.bxp([_bxpstat(yi, weights, whis) for yi in y], **kwds)
```
williamjameshandley, line 43: New code: weighted boxplots using `ax.bxp(stats,...)` rather than `ax.boxplot`

anesthetic/_matplotlib/boxplot.py
```diff
@@ -0,0 +1,129 @@
...
+        if return_type == "dict":
+            return bp, bp
+        elif return_type == "both":
+            return cls.BP(ax=ax, lines=bp), bp
+        else:
+            return ax, bp
+
+    def _make_plot(self):
+        if isinstance(self.data, _WeightedObject):
+            self.kwds['weights'] = self.data.weights
```
williamjameshandley, line 54: New code to set weights as a keyword argument to be passed to the `@classmethod` `_plot`. This idiom occurs regularly throughout the `anesthetic._matplotlib` backend.

anesthetic/_matplotlib/boxplot.py
```diff
@@ -0,0 +1,129 @@
...
+        return super()._make_plot()
+
+
+def boxplot(data, *args, **kwds):
+    # noqa: disable=D103
+    if isinstance(data, _WeightedObject):
+        kwds['weights'] = data.weights
```
williamjameshandley, line 61: New code

anesthetic/_matplotlib/boxplot.py
```diff
@@ -0,0 +1,129 @@
...
+
+    def create_plot_group():
```
williamjameshandley, line 63: This is a fancy piece of code to be passed to `replace_inner_function`

anesthetic/_matplotlib/boxplot.py
```diff
@@ -0,0 +1,129 @@
...
+            if weights is None:
+                values = [np.asarray(remove_na_arraylike(v), dtype=object)
+                          for v in values]
+                bp = ax.boxplot(values, **kwds)
+            else:
+                whis = kwds.pop("whis", 1.5)
+                kwds['showfliers'] = False
+                from anesthetic._matplotlib.boxplot import _bxpstat
+                bp = ax.bxp([_bxpstat(v, weights, whis) for v in values],
+                            **kwds)
```
williamjameshandley, line 90: New code: compute weighted boxplots

anesthetic/_matplotlib/boxplot.py
```diff
@@ -0,0 +1,129 @@
...
+            # Return axes in multiplot case, maybe revisit later # 985
+            if return_type == "dict":
+                return bp
+            elif return_type == "both":
+                return BoxPlot.BP(ax=ax, lines=bp)
+            else:
+                return ax
+
+    boxplot = replace_inner_function(pandas.plotting._matplotlib.boxplot,
+                                     create_plot_group.__code__.co_consts[1])
```
williamjameshandley, line 117: Use the rest of the original `pandas.plotting._matplotlib.boxplot` function, but overwrite the inner `plot_group` function which calls the unweighted `ax.boxplot`

anesthetic/_matplotlib/boxplot.py
```diff
@@ -0,0 +1,129 @@
...
+    return boxplot(data, *args, **kwds)
+
+
+def boxplot_frame(data, *args, **kwds):
+    # noqa: disable=D103
+    if isinstance(data, _WeightedObject):
+        kwds['weights'] = data.weights
```
williamjameshandley, line 124: New code

anesthetic/_matplotlib/core.py
```diff
@@ -0,0 +1,21 @@
...
+from pandas.plotting._matplotlib.core import HexBinPlot as _HexBinPlot
+from anesthetic.weighted_pandas import _WeightedObject
+
+
+class ScatterPlot(_ScatterPlot):
+    # noqa: disable=D101
+    def __init__(self, data, x, y, s=None, c=None, **kwargs) -> None:
+        if isinstance(data, _WeightedObject):
+            data = data.compress()
+            kwargs['alpha'] = kwargs.get('alpha', 0.5)
```
williamjameshandley, line 11: New code. Compute weighted scatter plots by compressing to the channel capacity, and switching to alpha of 0.5 (the same convention as `scatter_matrix`), so that repeated/overlapping points come out with greater density.

anesthetic/_matplotlib/core.py
```diff
@@ -0,0 +1,21 @@
...
+        super().__init__(data, x, y, s, c, **kwargs)
+
+
+class HexBinPlot(_HexBinPlot):
+    # noqa: disable=D101
+    def __init__(self, data, x, y, C=None, **kwargs) -> None:
+        if isinstance(data, _WeightedObject):
+            C = '__weights'
+            data[C] = data.weights
```
williamjameshandley, line 20: Compute a weighted hexbin plot by using the `C` functionality to accumulate weight rather than counts in each hexbin.

anesthetic/_matplotlib/hist.py
```diff
@@ -0,0 +1,68 @@
...
+
+        hist, bins = np.histogram(
+            values, bins=self.bins, range=self.kwds.get("range", None),
+            weights=self.kwds.get("weights", None)
+        )
+        return bins
+
+    def _make_plot(self):
+        if isinstance(self.data, _WeightedObject):
+            self.kwds['weights'] = self.data.weights
```
williamjameshandley, line 28: New code

anesthetic/_matplotlib/hist.py
```diff
@@ -0,0 +1,68 @@
...
+    # noqa: disable=D101
+    def _calculate_bins(self, data):
+        nd_values = data._convert(datetime=True)._get_numeric_data()
+        values = np.ravel(nd_values)
+        values = values[~isna(values)]
+
+        hist, bins = np.histogram(
+            values, bins=self.bins, range=self.kwds.get("range", None),
+            weights=self.kwds.get("weights", None)
+        )
```
williamjameshandley, line 23: Compute a weighted histogram by passing the weights keyword.

anesthetic/_matplotlib/hist.py
```diff
@@ -0,0 +1,68 @@
...
+        ind=None,
+        column_num=None,
+        stacking_id=None,
+        **kwds,
+    ):
+        from scipy.stats import gaussian_kde
+        weights = kwds.pop("weights", None)
+
+        y = remove_na_arraylike(y)
+        gkde = gaussian_kde(y, bw_method=bw_method, weights=weights)
```
williamjameshandley, line 50: Compute a weighted gaussian_kde by passing the weights keyword

anesthetic/_matplotlib/hist.py
```diff
@@ -0,0 +1,68 @@
...
+
+        y = gkde.evaluate(ind)
+        lines = MPLPlot._plot(ax, ind, y, style=style, **kwds)
+        return lines
+
+
+def hist_frame(data, *args, **kwds):
+    # noqa: disable=D103
+    if isinstance(data, _WeightedObject):
+        kwds['weights'] = data.weights
```
williamjameshandley, line 60: New code

anesthetic/_matplotlib/hist.py
```diff
@@ -0,0 +1,68 @@
...
+    return pandas.plotting._matplotlib.hist_frame(data, *args, **kwds)
+
+
+def hist_series(data, *args, **kwds):
+    # noqa: disable=D103
+    if isinstance(data, _WeightedObject):
+        kwds['weights'] = data.weights
```
williamjameshandley, line 67: New code

anesthetic/_matplotlib/misc.py
```diff
@@ -0,0 +1,16 @@
+import pandas.plotting._matplotlib.misc as misc
+from anesthetic.weighted_pandas import _WeightedObject
+
+
+def scatter_matrix(frame, *args, **kwargs):
+    # noqa: disable=D103
+    if isinstance(frame, _WeightedObject):
+        frame = frame.compress()
```
williamjameshandley, line 8: Compute a weighted scatter matrix by compressing to the channel capacity (could technically do slightly better than this by incorporating weights into the KDE/hist diagonal, but this would involve re-writing the whole function, when we have more functionality in `plot_2d`

anesthetic/utils.py
```diff
@@ -76,7 +76,7 @@ def quantile(a, q, w=None, interpolation='linear'):
     w = np.array(list(w))  # Necessary to convert pandas arrays
     i = np.argsort(a)
     c = np.cumsum(w[i[1:]]+w[i[:-1]])
-    c /= c[-1]
+    c = c/c[-1]
```
williamjameshandley, line 79: Needed to make this trivial change to avoid an error if the weights are integers and the `numpy.ufunc` complaints.

anesthetic/weighted_pandas.py
```diff
@@ -53,10 +53,6 @@ def sample(self, *args, **kwargs):
         """Weighted sample."""
         return super().sample(weights=self.weights, *args, **kwargs)
 
-    def hist(self, *args, **kwargs):
-        """Weighted histogram of the sampled distribution."""
-        return super().hist(weights=self.weights, *args, **kwargs)
-
```
williamjameshandley, line 59: No longer needed now we have the new backend.

tests/test_weighted_pandas.py
```diff
@@ -207,19 +209,6 @@ def test_WeightedDataFrame_quantile():
...
-    axes = df[['A', 'B']].hist(bins=20, density=True)
-    for ax in axes.flatten():
-        assert len(ax.patches) == 20
-        norm = 0
-        for patch in ax.patches:
-            norm += patch.get_height() * patch.get_width()
-        assert norm == pytest.approx(1)
-    plt.close("all")
-
-
```
williamjameshandley, line 222: No longer needed now we have the new backend

tests/test_weighted_pandas.py
```diff
@@ -412,18 +401,6 @@ def test_WeightedSeries_quantile():
...
-    series = test_WeightedSeries_constructor()
-    ax = series.hist(bins=20, density=True)
-    assert len(ax.patches) == 20
-    norm = 0
-    for patch in ax.patches:
-        norm += patch.get_height() * patch.get_width()
-    assert norm == pytest.approx(1)
-    plt.close("all")
-
-
```
williamjameshandley, line 426: No longer needed now we have the new backend.

tests/test_weighted_pandas.py
```diff
@@ -465,3 +442,163 @@ def test_WeightedSeries_nan():
...
+    plt.close("all")
+
+
+def test_bootstrap_plot():
+    df, wdf = mcmc_run()
+    bootstrap_plot(wdf.x)
+    plt.close("all")
+
+
+def test_BoxPlot():
```
williamjameshandley, line 547: The boxplot 'tests' don't actually test for anything, since it's hard to extract the values of the weighted boxplots for comparison with the unweighted version (and they give slightly different answers).

williamjameshandley (commented):
Some notes for the reviewer, highlight the significant changes (most of the rest is necessary boiler plate to overwrite the original.

anesthetic/__init__.py
```diff
@@ -11,6 +11,34 @@
...
+
+# Override the two places where _get_plot_backend is defined
+pandas.plotting._core._get_plot_backend = \
+        _anesthetic_override(pandas.plotting._core._get_plot_backend)
+pandas.plotting._misc._get_plot_backend = \
+        _anesthetic_override(pandas.plotting._misc._get_plot_backend)
+
+# Set anesthetic.plotting._matplotlib as the actual backend
+pandas.options.plotting.backend = 'anesthetic.plotting._matplotlib'
+
```
williamjameshandley, line 41: This code sets up the anesthetic._matplotlib backend as a replacement for matplotlib. In principle it should behave exactly the same as the current matplotlib backend for normal pandas arrays, but behave correctly for WeightedDataFrames

anesthetic/plotting/_matplotlib/boxplot.py
```diff
@@ -0,0 +1,122 @@
...
+
+def _bxpstat(y, weights, whis):
+    q1, med, q3 = quantile(y, [0.25, 0.5, 0.75], weights)
+    bxpstat = {}
+    bxpstat['med'] = med
+    bxpstat['q1'] = q1
+    bxpstat['q3'] = q3
+    bxpstat['whislo'] = q1 - whis*(q3-q1)
+    bxpstat['whishi'] = q3 + whis*(q3-q1)
+    return bxpstat
```
williamjameshandley, line 19: Compute weighted boxplot statistics.

anesthetic/plotting/_matplotlib/__init__.py
```diff
@@ -0,0 +1,43 @@
+"""anesthetic override of matplotlib backend for weighted samples."""
```
williamjameshandley, line 1: This imports all of the things it needs to from the matplotlib backend, and replaces the relevant functions and classes with functions defined in anesthetic._matplotlib

anesthetic/plotting/_matplotlib/boxplot.py
```diff
@@ -0,0 +1,122 @@
...
+
+        weights = kwds.pop("weights", None)
+
+        if weights is None:
+            bp = ax.boxplot(y, **kwds)
+        else:
+            whis = kwds.pop("whis", 1.5)
+            kwds['showfliers'] = False
+            y = np.atleast_2d(y)
+            bp = ax.bxp([_bxpstat(yi, weights, whis) for yi in y], **kwds)
```
williamjameshandley, line 43: New code: weighted boxplots using ax.bxp(stats,...) rather than ax.boxplot

anesthetic/plotting/_matplotlib/boxplot.py
```diff
@@ -0,0 +1,122 @@
...
+            return bp, bp
+        elif return_type == "both":
+            return cls.BP(ax=ax, lines=bp), bp
+        else:
+            return ax, bp
+
+
+def boxplot(data, *args, **kwds):
+    # noqa: disable=D103
+    _get_weights(kwds, data)
```
williamjameshandley, line 55: New code to set weights as a keyword argument to be passed to the @classmethod _plot. This idiom occurs regularly throughout the anesthetic.plotting._matplotlib backend.

anesthetic/plotting/_matplotlib/boxplot.py
```diff
@@ -0,0 +1,122 @@
...
+            if weights is None:
+                values = [np.asarray(remove_na_arraylike(v), dtype=object)
+                          for v in values]
+                bp = ax.boxplot(values, **kwds)
+            else:
+                whis = kwds.pop("whis", 1.5)
+                kwds['showfliers'] = False
+                from anesthetic.plotting._matplotlib.boxplot import _bxpstat
+                bp = ax.bxp([_bxpstat(v, weights, whis) for v in values],
+                            **kwds)
```
williamjameshandley, line 84: New code

anesthetic/plotting/_matplotlib/boxplot.py
```diff
@@ -0,0 +1,122 @@
...
+    _get_weights(kwds, data)
+
+    def create_plot_group():
+        fontsize = None  # pragma: no cover
+        kwds = None  # pragma: no cover
+        maybe_color_bp = None  # pragma: no cover
+        return_type = None  # pragma: no cover
+        rot = None  # pragma: no cover
+
+        def plot_group(keys, values, ax, **kwargs):  # pragma: no cover
```
williamjameshandley, line 64: This is a fancy piece of code to be passed to replace_inner_function

anesthetic/plotting/_matplotlib/boxplot.py
```diff
@@ -0,0 +1,122 @@
...
+            # Return axes in multiplot case, maybe revisit later # 985
+            if return_type == "dict":
+                return bp
+            elif return_type == "both":
+                return BoxPlot.BP(ax=ax, lines=bp)
+            else:
+                return ax
+
+    boxplot = replace_inner_function(pandas.plotting._matplotlib.boxplot,
+                                     create_plot_group.__code__.co_consts[1])
```
williamjameshandley, line 111: Use the rest of the original pandas.plotting._matplotlib.boxplot function, but overwrite the inner plot_group function which calls the unweighted ax.boxplot

anesthetic/plotting/_matplotlib/boxplot.py
```diff
@@ -0,0 +1,122 @@
...
+    return boxplot(data, *args, **kwds)
+
+
+def boxplot_frame(data, *args, **kwds):
+    # noqa: disable=D103
+    _get_weights(kwds, data)
```
williamjameshandley, line 117: New code

anesthetic/plotting/_matplotlib/core.py
```diff
@@ -0,0 +1,45 @@
...
+    if isinstance(data, _WeightedObject):
+        return data.compress(kwargs.pop('ncompress', None))
+    else:
+        return data
+
+
+class _CompressedMPLPlot(MPLPlot):
+    def __init__(self, data, *args, **kwargs):
+        data = _compress_weights(kwargs, data)
+        super().__init__(data, *args, **kwargs)
```
williamjameshandley, line 28: This automates compression of weighted samples into unweighted samples for scatter and bootstrap plots.

anesthetic/plotting/_matplotlib/core.py
```diff
@@ -0,0 +1,45 @@
...
+
+def _get_weights(kwargs, data):
+    if isinstance(data, _WeightedObject):
+        kwargs['weights'] = data.weights
+
+
+class _WeightedMPLPlot(MPLPlot):
+    def __init__(self, data, *args, **kwargs):
+        _get_weights(kwargs, data)
+        super().__init__(data, *args, **kwargs)
```
williamjameshandley, line 15: This automates passing the weights to histogram and kde plots

anesthetic/plotting/_matplotlib/core.py
```diff
@@ -0,0 +1,45 @@
...
+            kwargs['alpha'] = kwargs.get('alpha', 0.5)
+        super().__init__(data, x, y, s, c, **kwargs)
+
+
+class HexBinPlot(_HexBinPlot):
+    # noqa: disable=D101
+    def __init__(self, data, x, y, C=None, **kwargs) -> None:
+        if isinstance(data, _WeightedObject):
+            C = '__weights'
+            data[C] = data.weights
```
williamjameshandley, line 44: Compute a weighted hexbin plot by using the C functionality to accumulate weight rather than counts in each hexbin.

anesthetic/plotting/_matplotlib/hist.py
```diff
@@ -0,0 +1,71 @@
...
+                pass
+            weights = np.ravel(weights)
+            weights = weights[~isna(values)]
+
+        values = values[~isna(values)]
+
+        hist, bins = np.histogram(
+            values, bins=self.bins, range=self.kwds.get("range", None),
+            weights=weights
+        )
```
williamjameshandley, line 33: Compute a weighted histogram by passing the weights keyword.

anesthetic/plotting/_matplotlib/hist.py
```diff
@@ -0,0 +1,71 @@
...
+        values = np.ravel(nd_values)
+        weights = self.kwds.get("weights", None)
+        if weights is not None:
+            try:
+                weights = np.broadcast_to(weights[:, None], nd_values.shape)
+            except ValueError:
+                print(nd_values.shape, weights.shape)
+                pass
+            weights = np.ravel(weights)
+            weights = weights[~isna(values)]
```
williamjameshandley, line 26: Ensures any nan weights are also masked when plotting with subplots=True

anesthetic/plotting/_matplotlib/hist.py
```diff
@@ -0,0 +1,71 @@
...
+        ind=None,
+        column_num=None,
+        stacking_id=None,
+        **kwds,
+    ):
+        from scipy.stats import gaussian_kde
+        weights = kwds.pop("weights", None)
+
+        y = remove_na_arraylike(y)
+        gkde = gaussian_kde(y, bw_method=bw_method, weights=weights)
```
williamjameshandley, line 55: Compute a weighted gaussian_kde by passing the weights keyword

anesthetic/plotting/_matplotlib/misc.py
```diff
@@ -0,0 +1,14 @@
+import pandas.plotting._matplotlib.misc as misc
+from anesthetic.plotting._matplotlib.core import _compress_weights
+
+
+def scatter_matrix(frame, *args, **kwargs):
+    # noqa: disable=D103
+    frame = _compress_weights(kwargs, frame)
```
williamjameshandley, line 7: Compute a weighted scatter matrix by compressing to the channel capacity (could technically do slightly better than this by incorporating weights into the KDE/hist diagonal, but this would involve re-writing the whole function, when we have more functionality in plot_2d

anesthetic/plotting/_matplotlib/hist.py
```diff
@@ -0,0 +1,71 @@
...
+
+        y = gkde.evaluate(ind)
+        lines = MPLPlot._plot(ax, ind, y, style=style, **kwds)
+        return lines
+
+
+def hist_frame(data, *args, **kwds):
+    # noqa: disable=D103
+    _get_weights(kwds, data)
```
williamjameshandley, line 64: new code

anesthetic/plotting/_matplotlib/hist.py
```diff
@@ -0,0 +1,71 @@
...
+    return pandas.plotting._matplotlib.hist_frame(data, *args, **kwds)
+
+
+def hist_series(data, *args, **kwds):
+    # noqa: disable=D103
+    _get_weights(kwds, data)
```
williamjameshandley, line 70: new code

williamjameshandley:
CI lint currently fails due to a PEP update. When #195 is merged this will be fixed

anesthetic/plotting/_matplotlib/boxplot.py
lukashergt: Why `kwds` instead of the usual `kwargs`?

anesthetic/plotting/_matplotlib/boxplot.py
lukashergt: This is a curious nesting of functions and I'm getting a bit confused by the use of `kwds` and `kwargs`, then setting `kwds` to `None`, but then doing `kwds.pop`... Is this all intentional?

anesthetic/plotting/_matplotlib/hist.py
lukashergt: Why the `KdePlot` in the `hist.py` file?

anesthetic/utils.py
lukashergt: Yes, I've got this one in a few branches, too.

anesthetic/plotting/_matplotlib/core.py
lukashergt: Looks like `C` is what will be used with a colourbar. What type of normalisation would we want by default for that?

lukashergt: Why do `HistPlot` and `KDEPlot` get their own python file, but `ScatterPlot` and `HexBinPlot` are put in `core.py`?

anesthetic/plotting/_matplotlib/misc.py
lukashergt: Where does this get used? 

> could technically do slightly better than this by incorporating weights into the KDE/hist diagonal

Not sure what you mean here.

lukashergt (changes requested):
The new tests only test the new behaviour of the weighted data frames directly, but not for `MCMCSamples` or `NestedSamples`. Those actually fail because they come with the `plot` method that overwrites the `plot` attribute of `WeightedDataFrame`. 

We should add further tests to `test_samples.py` and find a new name for the `plot` method. Maybe `_plot`? 

We should also avoid for loops in tests. We should rewrite the for loops with pytests parametrised tests.

tests/test_weighted_pandas.py
```diff
@@ -1,9 +1,11 @@
 from anesthetic.weighted_pandas import WeightedDataFrame, WeightedSeries
+from anesthetic.utils import channel_capacity
 from pandas import Series, DataFrame
 import pytest
 import numpy as np
 from numpy.testing import assert_array_equal, assert_allclose
 import matplotlib.pyplot as plt
+from pandas.plotting import scatter_matrix, bootstrap_plot
```
lukashergt, line 8: Shouldn't this be `from anesthetic.plotting import scatter_matrix, bootstrap_plot`?

anesthetic/plotting/_matplotlib/boxplot.py
williamjameshandley: Yes, `ax.boxplot` doesn't accept weights, so you have to manually compute the weighted boxplots statistics and use the underlying `ax.bxp` which matplotlib's `ax.boxplot` calls underneath.

anesthetic/plotting/_matplotlib/__init__.py
williamjameshandley: These are indeed included, since we import the `PLOT_CLASSES` from pandas, and then overwrite the code which needs modification. Since I don't _think_ the other classes (namely `line`, `bar`, `barh`, `area` and `pie`).

```python
>>> from anesthetic.plotting._matplotlib import PLOT_CLASSES
>>> print(PLOT_CLASSES)
{'line': pandas.plotting._matplotlib.core.LinePlot,
 'bar': pandas.plotting._matplotlib.core.BarPlot,
 'barh': pandas.plotting._matplotlib.core.BarhPlot,
 'box': anesthetic.plotting._matplotlib.boxplot.BoxPlot,
 'hist': anesthetic.plotting._matplotlib.hist.HistPlot,
 'kde': anesthetic.plotting._matplotlib.hist.KdePlot,
 'area': pandas.plotting._matplotlib.core.AreaPlot,
 'pie': pandas.plotting._matplotlib.core.PiePlot,
 'scatter': anesthetic.plotting._matplotlib.core.ScatterPlot,
 'hexbin': anesthetic.plotting._matplotlib.core.HexBinPlot}
```

Giving it a bit more thought, there may be some instances of e.g. pie when computing down a column you might want weighted versions of these -- will get back to you here.

anesthetic/plotting/_matplotlib/boxplot.py
williamjameshandley: This is consistent with the matplotlib code it replaces, so it will be easier to diff if we make the changes as minimal as possible.

anesthetic/plotting/_matplotlib/boxplot.py
williamjameshandley: The fancy code here means that it is essential to distinguish between the two. I agree this code is very spaghettified, but the alternative is to have a much larger copy-paste job of the [entire function](https://github.com/pandas-dev/pandas/blob/v1.4.3/pandas/plotting/_matplotlib/boxplot.py#L270-L407) (which is a little long, and therefore less maintainable)

anesthetic/plotting/_matplotlib/core.py
williamjameshandley: This is consistent with the `pandas.plotting._matplotlib` structure, see e.g. [ScatterPlot](https://github.com/pandas-dev/pandas/blob/v1.4.3/pandas/plotting/_matplotlib/core.py#L1055-#L1140).

anesthetic/plotting/_matplotlib/hist.py
williamjameshandley: Again, consistent with [`pandas.plotting._matplotlib`](https://github.com/pandas-dev/pandas/blob/v1.4.3/pandas/plotting/_matplotlib/hist.py#L169-#L228)

anesthetic/plotting/_matplotlib/misc.py
williamjameshandley: `scatter_matrix` is the pandas equivalent of our `plot_2d` (but with less functionality). it is used in `tests.test_weighted_pandas.test_scatter_matrix`. Implementation wise I could be slightly smarter by incorporating weights for the diagonal plots of the kde and hist, rather than just compressing to equally weighted samples as we do at the moment.

```python
>>> from pandas.plotting import scatter_matrix
>>>
>>> scatter_matrix(wdf)                                                                      
```

tests/test_weighted_pandas.py
williamjameshandley: You can see something of the intricacy of the backend system here. `pandas.plotting.scatter_matrix` calls the backend. Originally this would then have directed to `pandas.plotting._matplotlib.misc.scatter_plot`, but now this directs to `anesthetic.plotting._matplotlib.misc.scatter_plot`. It's important that it does this, since this way if you call native pandas functions on weighted data frames/series it still uses the weights, whilst still doing the unweighted job when there are no weights.

williamjameshandley:
Many thanks @lukashergt for this thorough check of the code.

>The new tests only test the new behaviour of the weighted data frames directly, but not for MCMCSamples or NestedSamples. Those actually fail because they come with the plot method that overwrites the plot attribute of WeightedDataFrame. Try for example ns.plot.hexbin().

Keen eyes! Addressing this is the next PR (which is ready to go once I've adjusted all your changes). This is the main breaking change I've been wanted to make before bumping the major version number. Let's address this there.

I will now get on with the pie plots and parametrised tests.

tests/test_weighted_pandas.py
lukashergt: Ouf, that's subtle, nice. 

Are we testing for this, i.e. are we testing that `scatter_matrix` and `bootstrap_plot` still behave as usual for **un**weighted samples?

If not, is there an easy way of doing that?

williamjameshandley:
Parametrized tests-wise, it's proving a bit hard to see how we can do that -- the for loops are actually over a set of plots which a single command produces (e.g. the `axes` produced by `axes = df.hist()`), so it's not clear to me that they fit here. I have however tried out some new [`pytest.fixture`s](https://docs.pytest.org/en/6.2.x/fixture.html), which are the proper way to generate thing which are continually reused.

tests/test_weighted_pandas.py
williamjameshandley: Good point -- I have added a test which checks that the `scatter_matrix` call is indeed different from the `pandas.plotting._matplotlib.misc.scatter_matrix` call.

anesthetic/plotting/_matplotlib/core.py
lukashergt: Do we have a test for this?

anesthetic/plotting/_matplotlib/core.py
lukashergt: > Looks like `C` is what will be used with a colourbar. What type of normalisation would we want by default for that?

How do we prefer to treat the normalisation for colourbars? 

For the 1-dim case we have a kwarg `density`, with `density=False` normalising such that the min is 0 and the max is 1, and with `density=True` normalising such that the integral is 1.

anesthetic/plotting/_matplotlib/misc.py
lukashergt: Hmm, difficult one. I'm fine either way. Maybe not so bad to have the 1D hist be the projection from the 2D scatter realisation.

On another note, how does `scatter_matrix` handle the split into 1D diagonal plots and 2D off-diagonal plots? Is there something better than our approach with `twin` axes?

williamjameshandley:
> Looks like C is what will be used with a colourbar. What type of normalisation would we want by default for that?

Excellent question. This triggered some code diving into `ax.hexbin`, but I can get the unweighted and weighted calculations to line up exactly if (a) the unweighted version is passed `mincnt=1`, which masks out 0 weight bins (which the `C` version has to do apparently), and (b) the weighted version as a default `reduce_C_function` as `np.sum`, so that the normalisation is by the sum of the weights (rather than the mean). I think given this means an mcmc repeated row weight now gives the same answer, this is the normalisation of least surprise for a weighted data frame.

williamjameshandley:
I've now had a close look at `bar`, `barh`, `area`, `line` and `pie`, and I don't think these plots really have a weighted counterpart. I'd propose to leave them as-is unless somebody finds a use case.

anesthetic/plotting/_matplotlib/misc.py
williamjameshandley: To recount our analogue discussion just now -- I think that (a) scatter_plot is not a particularly good way to do it (b) it would be good to revisit whether the `twin` axes is the best way to achieve what we want given that (c) the axes twinning is primarily designed to ensure that yticks appear on the top left corner of a plot_2d.

williamjameshandley: To recount our analogue discussion just now -- I think that (a) scatter_plot is not a particularly good way to do it (b) it would be good to revisit whether the `twin` axes is the best way to achieve what we want given that (c) the axes twinning is primarily designed to ensure that yticks appear on the top left corner of a plot_2d.

tests/test_weighted_pandas.py
```diff
@@ -465,3 +413,181 @@ def test_WeightedSeries_nan():
...
+
+
+def test_WeightedDataFrame_hist(mcmc_df, mcmc_wdf):
+    df_axes = mcmc_df.hist()
+    wdf_axes = mcmc_wdf.hist()
+    for df_ax, wdf_ax in zip(df_axes.flatten(), wdf_axes.flatten()):
+        for df_patch, wdf_patch in zip(df_ax.patches, wdf_ax.patches):
+            assert df_patch.get_height() == wdf_patch.get_height()
+            assert df_patch.get_width() == wdf_patch.get_width()
+            assert df_patch.get_xy() == wdf_patch.get_xy()
```
lukashergt, line 451: Instead of the `for` loops, here, it might be nicer to simply compare lists, e.g. for comparing heights:
```python
    df_axes = mcmc_df.hist().flatten()
    wdf_axes = mcmc_wdf.hist().flatten()
    df_heights = list(p.get_height() for ax in df_axes for p in ax.patches)
    wdf_heights = list(p.get_height() for ax in wdf_axes for p in ax.patches)
    assert df_heights == wdf_heights
```
In case of failure this will flag the list elements that are not equal.

lukashergt:
> Parametrized tests-wise, it's proving a bit hard to see how we can do that -- the for loops are actually over a set of plots which a single command produces (e.g. the `axes` produced by `axes = df.hist()`), so it's not clear to me that they fit here. I have however tried out some new [`pytest.fixture`s](https://docs.pytest.org/en/6.2.x/fixture.html), which are the proper way to generate thing which are continually reused.

I agree that parametrized tests are not the correct thing for this. However, for such cases I think it is better to turn things into lists and check for equality of those lists. For an example see [my inline comment](https://github.com/williamjameshandley/anesthetic/pull/194#discussion_r939598842).

lukashergt:
> I've now had a close look at `bar`, `barh`, `area`, `line` and `pie`, and I don't think these plots really have a weighted counterpart. I'd propose to leave them as-is unless somebody finds a use case.

One thing to consider for these cases is what we believe to be the correct handling of the MultiIndex. 

Example:
```python
wdf.x.plot.line()
```
image

Note the `(#, weight)` tuple that is used for the x-labels. Is that the desired behaviour?

anesthetic/plotting/_matplotlib/core.py
williamjameshandley: No, but it is covered by anything that derives from it (e.g. `ScatterPLot`). Since it isn't a public function, I wouldn't consider this warranting a test, since in theory the underlying implementation could change without changing the public behaviour (and underscore prefixes are meant to indicate this to users).

anesthetic/plotting/_matplotlib/core.py
williamjameshandley: Ah, I see. At the moment I think we probably want `WeightedDataFrame` to be consistent with what `DataFrame` does, namely to normalise the colour to the maximum weight. If we wanted to implement a `density` keyword, we should implement our own `anesthetic.plot.hexbin_plot_2d` wrapper as we do for KDE and hists.

anesthetic/plotting/_matplotlib/core.py
lukashergt: My question was more regarding your comment rather than this specific class/method, i.e. whether the compressed samples match the expected unweighted samples. But I guess this is automatically tested by testing scatter and bootstrap plots...?

tests/test_weighted_pandas.py
williamjameshandley: OK -- now does this [6e3f4b0](https://github.com/williamjameshandley/anesthetic/pull/194/commits/6e3f4b0993b46a24867ae93f3ca84ee8061c2b80). I agree this is much neater.

williamjameshandley:
[20711fe](https://github.com/williamjameshandley/anesthetic/pull/194/commits/20711fe1173dc04052dcc42b0d2fcaecd968931e) now introduces functionality for area, line, bar, barh and pie plots, and so addresses the below


> Note the `(#, weight)` tuple that is used for the x-labels. Is that the desired behaviour?

No! This has always annoyed me e.g. when doing `samples.nlive.plot()` to get a line plot of the number of live points over the run. Now it's much more sensible.

This update will also be much more robust when #198 is merged.



> I've now had a close look at `bar`, `barh`, `area`, `line` and `pie`, and I don't think these plots really have a weighted counterpart. I'd propose to leave them as-is unless somebody finds a use case.

Having implemented the tests, I can now imagine a couple of use cases. The plots are now much more readable with the updated index labels.

anesthetic/plotting/_matplotlib/core.py
```diff
@@ -0,0 +1,101 @@
...
+        if isinstance(self.data, _WeightedObject):
+            return np.squeeze(self.data.index.droplevel('weights').names)
+        else:
+            return super()._get_index_name()
+
+    def _get_xticks(self, convert_period: bool = False):
+        if isinstance(self.data, _WeightedObject):
+            return self.data.index.droplevel('weights')._mpl_repr()
+        else:
+            return super()._get_xticks(convert_period)
```
lukashergt, line 33: This does seem to fix the issue of having unwelcome `weights` on the x-axis. However, this also seems to remove the x-ticklabels when we actually want them. Try:
```python 
wdf = WeightedDataFrame(ns)
wdf.plot.hexbin('x0', 'x1')
```

image


The x-ticklabels are gone, even though we would like to have the labels from `'x0'`.

lukashergt: Actually, no, this seems to be the (ugly) default behaviour of pandas, so nevermind...

lukashergt (approved):
Looks good. I reckon there are probably still things to iron out, but best to move on, we might catch those with the upcoming PRs.

williamjameshandley:
Many thanks @lukashergt

PR #198, opened 2022-08-07
WeightedPandas index improvements

williamjameshandley:
# Description

This provides a few simultaneous upgrades to WeightedPandas, which were sparked by some of the indexing considerations in #194. These are

1. `WeightedDataFrame` and `WeightedSeries` always have a weights column in their `MultiIndex`. If no weights are passed, the column is filled with 1's (integers).
2. More complicated `MultiIndex`es are now possible (previously it was assumed there would be only one other 'iteration' column).
3. The 'other' column is now no longer called `'#'`, but will retain the name of the previous index/indices (if present at all).

The above changes make the code more robust and extensible, particularly when it comes to extracting/dropping weights (we can assume that `df.index.get_level_values('weights')` and `df.index.droplevel('weights')` will always work), but also making sure that code like `(a*b).mean()` is only possible between weighted objects if they have the same weights.

# Checklist:

### DIFF
--- a/anesthetic/weighted_pandas.py
+++ b/anesthetic/weighted_pandas.py
@@ -292,7 +292,9 @@ class WeightedDataFrame(_WeightedObject, DataFrame):
         if not numeric_only:
             raise NotImplementedError("numeric_only kwarg not implemented")
         if axis == 0:
-            data = np.array([c.quantile(q) for _, c in self.iteritems()])
+            data = np.array([c.quantile(q, interpolation=interpolation,
+                                        numeric_only=numeric_only)
+                             for _, c in self.iteritems()])
             if np.isscalar(q):
                 return Series(data, index=self.columns)
             else:
@@ -319,12 +321,12 @@ class WeightedDataFrame(_WeightedObject, DataFrame):
         df.index = df.index.droplevel('weights')
         return df
 
-    def _reduce(self, op, name, **kwargs):
-        answer = super()._reduce(op, name, **kwargs)
-        if kwargs['axis'] == 0:
-            answer = Series(answer).droplevel('weights')
-            answer.name = None
-        return answer
+    #def _reduce(self, op, name, **kwargs):
+    #    answer = super()._reduce(op, name, **kwargs)
+    #    print(kwargs['axis'])
+    #    if self._get_axis_number(kwargs['axis']) == 0 and 'weights' in self.index.names:
+    #        answer = Series(answer).droplevel('weights', 0)
+    #    return answer
 
     @property
     def _constructor_sliced(self):
@@ -334,14 +336,12 @@ class WeightedDataFrame(_WeightedObject, DataFrame):
     def _constructor(self):
         return WeightedDataFrame
 
-    def _ixs(self, i, axis=0): 
-        answer = super()._ixs(i, axis=axis)
-        #print("columns", self.columns)
-        #print("index", self.index)
-        
-        #if self._get_axis_number(axis) == 0:
-        #    answer = Series(answer.droplevel('weights'))
-        return answer
+    #def _ixs(self, i, axis=0): 
+    #    answer = super()._ixs(i, axis=axis)
+    #    
+    #    if self._get_axis_number(axis) == 0:
+    #        answer = Series(answer.droplevel('weights'))
+    #    return answer
 
 
 
--- a/anesthetic/weighted_pandas.py
+++ b/anesthetic/weighted_pandas.py
@@ -6,6 +6,7 @@ from pandas.util import hash_pandas_object
 from numpy.ma import masked_array
 from anesthetic.utils import (compress_weights, channel_capacity, quantile,
                               temporary_seed)
+from inspect import signature
 
 
 class _WeightedObject(object):
@@ -14,17 +15,16 @@ class _WeightedObject(object):
     def __init__(self, *args, **kwargs):
         weights = kwargs.pop('weights', None)
         super().__init__(*args, **kwargs)
-        if weights is None:
-            if 'weights' in self.index.names:
-                weights = self.weights
-            else:
-                weights = np.ones_like(self.index, int)
-        self._set_weights(weights)
+        if weights is not None and 'weights' not in self.index.names:
+            self._set_weights(weights)
 
     @property
     def weights(self):
         """Sample weights."""
-        return self.index.get_level_values('weights').to_numpy()
+        if 'weights' in self.index.names:
+            return self.index.get_level_values('weights').to_numpy()
+        else:
+            return np.ones_like(self.index)
 
     @weights.setter
     def weights(self, weights):
@@ -47,13 +47,6 @@ class _WeightedObject(object):
             answer._set_weights(weights)
             return answer
 
-    @property
-    def _rand(self):
-        """Random number for consistent compression."""
-        seed = hash_pandas_object(self.index).sum() % 2**32
-        with temporary_seed(seed):
-            return np.random.rand(len(self))
-
     def std(self, *args, **kwargs):
         """Weighted standard deviation of the sampled distribution."""
         return np.sqrt(self.var(*args, **kwargs))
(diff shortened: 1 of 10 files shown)


anesthetic/weighted_pandas.py
```diff
@@ -14,21 +14,38 @@ class _WeightedObject(object):
...
+
+    def reset_index(self, level=None, drop=False, inplace=False, *args):
+        weights = self.weights
+        answer = super().reset_index(level=level, drop=drop,
+                                     inplace=inplace, *args)
+        if inplace:
+            self._set_weights(weights)
+        else:
+            answer._set_weights(weights)
+            return answer
```
lukashergt, line 48: Is `reset_index` behaving the way we'd expect it to? Try e.g.:
```python
print(ns.head())
ns.reset_index(level=0, inplace=True)
print(ns.head())
```

anesthetic/weighted_pandas.py
```diff
@@ -298,9 +315,16 @@ def compress(self, nsamples=None):
...
-        df.index = df.index.get_level_values('#')
+        df.index = df.index.droplevel('weights')
         return df
 
+    def _reduce(self, op, name, **kwargs):
+        answer = super()._reduce(op, name, **kwargs)
+        if kwargs['axis'] == 0:
+            answer = Series(answer).droplevel('weights')
+            answer.name = None
+        return answer
```
lukashergt, line 326: Why is this needed?

Why is this not needed for `WeightedSeries`?

tests/test_weighted_pandas.py
```diff
@@ -465,3 +466,18 @@ def test_WeightedSeries_nan():
...
+    index_1 = np.arange(N)
+    index_2 = np.random.randint(0, 1000, N)
+    index = MultiIndex.from_arrays([index_1, index_2], names=['A', 'B'])
+    data = np.random.rand(N, 5)
+    weights = np.random.rand(N)
+    wdf = WeightedDataFrame(data, weights=weights, index=index)
+
+    assert wdf.index.names == ['A', 'B', 'weights']
+    assert_allclose(np.array([*wdf.index]).T, [index_1, index_2, weights])
+    assert wdf.reset_index().index.names == [None, 'weights']
```
lukashergt, line 483: `reset_index` is only tested for its default arguments `level=None, drop=False, inplace=False`, here. Can we test the alternative arguments, too?

anesthetic/weighted_pandas.py
williamjameshandley: I believe so. The key is to compares this to the results from `df = DataFrame(ns).copy()`, which I believe matches up. `reset_index` can be a bit counterintuitive at first.

anesthetic/weighted_pandas.py
williamjameshandley: Removed in latest update

williamjameshandley:
OK, fairly major update.

It turns out that the 'nice to have' idea of default unit weights in the weighted index when `weights=None` is passed is not going to work. The main issue is that when transposing `samples.T.index` (i.e. the original `samples.columns`) would now have unit weights. This is ugly, plays havoc with index merging, and potentially leads to unusual behaviour in any code which calls a transpose (which we cannot easily/should not forbid). 

Fundamentally `samples.mean(axis=1)` should give the same thing as `samples.T.mean()`. More importantly `samples.T.mean(axis=1)` should give the same as `samples.mean()`. In the original code, these were different, with the former giving the unweighted mean by transferring the weights to a column where the hard-coded `axis=0`  therefore skipped.

The new updates now aim tie weighted sums to any axis with `'weights'` in a MultiIndex, and the above sums work (and are tested to work).

I would value feedback from @lukashergt as to the new user interface, in particular the now slightly inconsistent `.weights` getter and setter and `__init__` function, which only extracts weights from an index for a `WeightedDataFrame`, rather than from a column. It should be noted that `reset_index` also performs similarly, but this is consistent with pandas.

anesthetic/weighted_pandas.py
```diff
@@ -143,6 +154,18 @@ def compress(self, nsamples=None):
...
+    def _rand(self):
+        """Random number for consistent compression."""
+        seed = hash_pandas_object(self.index).sum() % 2**32
+        with temporary_seed(seed):
+            return np.random.rand(len(self))
+
+    @property
+    def _weighted(self):
+        """Determine if weights are actually present."""
+        return 'weights' in self.index.names
```
lukashergt, line 167: * This doesn't need to be private, right? We could call it `isweighted`, which is the typical way of indicating a boolean return.
   (Same question for `WeightedDataFrame`)
* I think I would prefer this to be a method rather than a property, since it is a method for `WeightedDataFrame`.

anesthetic/weighted_pandas.py
```diff
@@ -155,131 +178,161 @@ def _constructor_expanddim(self):
 class WeightedDataFrame(_WeightedObject, DataFrame):
     """Weighted version of pandas.DataFrame."""
 
-    def mean(self, axis=0, skipna=True):
+    def _weights(self, axis):
+        return self._get_axis(axis).get_level_values('weights').to_numpy()
```
lukashergt, line 182: * Should we call this `get_weights`?
* Should `axis` have a default value? Or is it better to force input to ensure correct usage?

anesthetic/weighted_pandas.py
```diff
@@ -14,21 +14,38 @@ class _WeightedObject(object):
...
-            self.index = [self.index.get_level_values(0), weights]
-            self.index.set_names(['#', 'weights'], inplace=True)
+        self._set_weights(weights)
+
+    def _set_weights(self, weights):
+        self.index = [self.index.get_level_values(name)
+                      for name in self.index.names if name != 'weights'] \
+                          + [weights]
+        names = self.index.names[:-1] + ['weights']
+        self.index.set_names(names, inplace=True)
```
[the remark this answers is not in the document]
lukashergt: Should we maybe simply get rid of the `@weights.setter` entirely and just have the methods `get_weights(axis)` and `set_weights(axis)`? Possibly such methods would be better located in `WeightedSeries` and `WeightedDataFrame`?

tests/test_weighted_pandas.py
```diff
@@ -76,124 +84,220 @@ def test_WeightedDataFrame_key(frame):
...
+    assert_allclose(correl, frame.corr()['A'])
 
     correl = frame.corrwith(frame[['A', 'B']])
-    assert isinstance(correl, Series)
+    assert isinstance(correl, WeightedSeries)
+    assert not correl._weighted
+    assert_array_equal(correl.index, frame.columns)
     assert_allclose(correl['A'], 1, atol=1e-2)
     assert_allclose(correl['B'], 1, atol=1e-2)
     assert np.isnan(correl['C'])
```
lukashergt, line 174: What about a test case where for `frame.corrwith(other)`:
* `frame` is unweighted?
* `other` is unweighted?

williamjameshandley:
OK, the new API with `get_weights` and `set_weights`, minimising `@property`s and conforming to pandas copying scheme is now finished.

One problem this has revealed is that in the 'correct' copying scheme, metadata is not deep copied [in pandas](https://stackoverflow.com/questions/50372509/why-are-attributes-lost-after-copying-a-pandas-dataframe). This is even more reason to implement #214 and #215, which I'll get on with now. If we're happy with the tex dictionary doing something potentially funny in copying until then, this is ready to be merged.

tests/test_weighted_pandas.py
williamjameshandley: At the moment this fails with `ValueError: cannot join with no overlapping index names`. I think that this is the correct behaviour, since it forbids correlating weighted and unweighted series (when strictly any combination (addition, multiplication, correlation) of series for weighted operations should have the same weights).

williamjameshandley: I've added a test that checks for this failure.

tests/test_weighted_pandas.py
williamjameshandley: Now added!

anesthetic/weighted_pandas.py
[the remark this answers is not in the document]
williamjameshandley: Done

anesthetic/weighted_pandas.py
williamjameshandley: Agreed. Done

anesthetic/weighted_pandas.py
williamjameshandley: Agreed on both counts. Done.

lukashergt (dismissed):
This looks great, just one minor question/comment.

anesthetic/weighted_pandas.py
```diff
@@ -14,28 +15,59 @@ class _WeightedObject(object):
...
+            elif inplace:
+                result = self
+            else:
+                result = self.copy()
+        else:
+            result = self.set_axis([self._get_axis(axis).get_level_values(name)
+                                    for name in self._get_axis(axis).names
+                                    if name != 'weights'] + [weights],
+                                   axis=axis)
+            names = result._get_axis(axis).names[:-1] + ['weights']
```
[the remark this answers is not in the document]
williamjameshandley: As far as I can tell, nothing assumes that weight is the last index. In the above function `set_weights` puts it there, but nothing else assumes it (e.g. calling `droplevel('weights')` or `get_weights`. For indexing it is much more convenient to have weights as the last. However, it might be surprising to users if `set_weights` reordered a multiindex which they had been manipulating, so I have adjusted the function so

a) If weights is present, it keeps the same location in the multiindex
b) If weights is not present, it is appended last by default (for convenience)
c) the result, self syntax is more consistent with usual inplace optional functions.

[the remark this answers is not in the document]
williamjameshandley: Tests now added to check this behaviour, as well as a `level` kwarg if you want to insert weights at a specific level (defaulting to the back).

lukashergt (approved):
Great work, please squash and merge.

williamjameshandley:
Many thanks for your help reviewing @lukashergt

williamjameshandley:
This one may prove fiddly to merge into your active branches -- let me know if you need any help.

PR #207, opened 2022-08-09
MCMC plotting

### LINKED ISSUE #59: Closer mirroring of pandas

williamjameshandley:
# Description

This follows on from #194, resolving the issue with `MCMCSamples.plot` overwriting `WeightedPandas.plot`. It removes the `MCMCSamples.plot` function entirely, with the existing `plot_2d` and `plot_1d` interfaces using `WeightedSamples.plot` in a pandas-like way

Fixes #59 

# Checklist:


anesthetic/weighted_pandas.py
```diff
@@ -294,7 +294,8 @@ def compress(self, nsamples=None):
         data = np.repeat(self.to_numpy(), i, axis=0)
         index = self.index.repeat(i)
         df = DataFrame(data=data, index=index, columns=self.columns)
-        df.index = df.index.get_level_values('#')
+        if self.index.nlevels == 2:
+            df.index = df.index.get_level_values('#')
```
lukashergt, line 298: weird, possibly to be undone in #198

anesthetic/samples.py
```diff
@@ -113,132 +116,6 @@ def _set_automatic_limits(self):
...
-
-        paramname_y: str
-            Choice of parameter to plot on y-coordinate from self.columns.
-            optional, if not provided, or the same as paramname_x, then 1D plot
-            produced.
-
-        plot_type: str
-            Must be in {'kde', 'scatter', 'hist', 'fastkde'} for 2D plots and
-            in {'kde', 'hist', 'fastkde', 'astropyhist'} for 1D plots.
-            optional, (Default: 'kde')
```
lukashergt, line 137: This should now be called `kind` and put into `plot_1d` docstring.

tests/test_samples.py
```diff
@@ -911,3 +920,19 @@ def test_logL_list():
...
+    samples.x0.plot.kde(subplots=True)
+
+    samples.plot.hist_2d('x0', 'x1')
+    samples.plot.kde_2d('x0', 'x1')
+    samples.plot.fastkde_2d('x0', 'x1')
+    samples.plot.kde_1d()
+    samples.plot.fastkde_1d()
+    samples.plot.hist_1d()
+
+    plt.close("all")
```
lukashergt, line 938: * Does this belong here? Or rather in `test_plot.py` or even `test_weighted_pandas.py`?
* Should these tests use a `WeightedDataFrame` instead of `NestedSamples` as a baseline?
* Can we make these tests a bit more meaningful, beyond just testing the existence of those methods?

anesthetic/samples.py
```diff
@@ -289,8 +174,8 @@ def plot_2d(self, axes, *args, **kwargs):
             this is used for creating the plot. Otherwise a new set of axes are
             created using the list or lists of strings.
 
-        types: dict, optional
-            What type (or types) of plots to produce. Takes the keys 'diagonal'
+        kind: dict, optional
+            What kinds of plots to produce. Takes the keys 'diagonal'
             for the 1D plots and 'lower' and 'upper' for the 2D plots.
             The options for 'diagonal are:
                 - 'kde'
```
lukashergt, line 181: 1) Can we allow both strings and dicts for `kind` and have strings be broadcast across possible options for `lower`, `diagonal`, and `upper`? For example:
   * `'kde_1d'` --> `{'diagonal': 'kde_1d'}`
   * `'kde_2d'` --> `{'lower': 'kde_2d'}`
   * `'kde'` --> `{'diagonal': 'kde_1d', 'lower': 'kde_2d'}`
   * equivalently for `'hist'` instead of `'kde'`
2) `kind` or `kinds`? I'm ultimately happy with whichever feels to make more sense when you are trying to sort point (1) out.

anesthetic/samples.py
williamjameshandley: Updated in 0a3b7e0

anesthetic/samples.py
williamjameshandley: I agree. I've added in this string-based option in 0a3b7e0. For now I've gone with allowing both `'kind'` and `'kinds'` as synonyms.

tests/test_samples.py
```diff
@@ -911,3 +936,23 @@ def test_logL_list():
...
+    samples.plot.kde_1d()
+    samples.plot.hist_1d()
+
+    try:
+        samples.plot.fastkde_2d('x0', 'x1')
+        samples.plot.fastkde_1d()
+    except ImportError:
+        pass
+
+    plt.close("all")
```
lukashergt, line 958: * delete this and hope for 100% coverage
* create an issue that this is not perfect pandas behaviour

anesthetic/samples.py
```diff
@@ -319,33 +230,58 @@ def plot_2d(self, axes, *args, **kwargs):
...
 
         """
-        default_types = {'diagonal': 'kde', 'lower': 'kde', 'upper': 'scatter'}
-        types = kwargs.pop('types', default_types)
+        kind = kwargs.pop('kind', 'default')
+        kind = kwargs.pop('kinds', kind)
+        try:
+            kind = self.plot_2d_default_kinds.get(kind, kind)
+        except TypeError:
+            pass
```
lukashergt, line 238: This doesn't fully work, make failing test and fix.
Maybe simply check for a string and use the dictionary.

tests/test_samples.py
lukashergt: > Does this belong here? Or rather in `test_plot.py` or even `test_weighted_pandas.py`?
> Should these tests use a `WeightedDataFrame` instead of `NestedSamples` as a baseline?

Renamed to `test_samples_dot_plot` to make clear these tests are meant primarily to test for the pandas `.plot` attribute. These are not meant to be thorough tests of the plotting functions.

> Can we make these tests a bit more meaningful, beyond just testing the existence of those methods?

Done in [aa24d72](https://github.com/williamjameshandley/anesthetic/pull/207/commits/aa24d724a9e7f6ab005b3e441b5394325b2165d0).

tests/test_samples.py
lukashergt: > * delete this and hope for 100% coverage

Failed. Implemented more thorough tests in [aa24d72](https://github.com/williamjameshandley/anesthetic/pull/207/commits/aa24d724a9e7f6ab005b3e441b5394325b2165d0) instead.

anesthetic/samples.py
lukashergt: Done.

lukashergt (approved):

williamjameshandley:
Many thanks @lukashergt for reviewing and finishing the tests. I've made an issue #211 to remind us that at some point we should fix the corner case, but this shouldn't hold us up.

Issue #208, opened 2022-08-10
Should we change the way we read files?


lukashergt:
Our current way of reading chain files (be it from PolyChord, MultiNest, Cobaya, ...) is not how pandas would do it. In pandas you use functions such as `pandas.read_csv`. 

* Should we have an `anesthetic.read` function? 
* Should it be just `anesthetic.read` or more specifically `anesthetic.read_polychord`, `anesthetic.read_multinest`, etc. or yet rather `anesthetic.read_mcmc` and `anesthetic.read_ns`?

williamjameshandley:
Unless anybody objects, we should go with the philosophy of 'make it look more like pandas' (the third option would be to use `@classmethod`s and have `samples = NestedSamples.read(root)`, but this is not what pandas does, and I now seem to be remembering that when I tried this it did not gel well with subclassing `DataFrame`s. 

Another virtue of `read_chains` would be that it would simplify the `__init__` functions to make it easier for future users to modify.

Issue #200, opened 2022-08-08
Consider removing user ability to set `.limits`


williamjameshandley:
One of the interfacing problems in #105 is that we would rather encourage users to set limits at the prior level by importance sampling (e.g. `samples.importance_sample(samples.x0>0, action='mask')`) in comparison with setting the `.limits`, which should in principle be extractable purely from the samples themselves. We would propose to remove the `.limits` and at the very least replace with `._limits`

lukashergt:
It would also be good to crack the anesthetic plotting limits once and for all, i.e. automatically determine the appropriate plotting range that 
a) ensures that only the relevant bits get plotted, i.e. all non-transparent things, and not crazy long tails,
b) to that end it needs to reliably compute the important plotting range (difficult case being e.g. multimodal distributions),
c) and upon adding a second set of samples to a first set, the plot range should be only expanded, and should not loose the first plots.

PR #209, opened 2022-08-10
create parent class `Samples` for child classes `MCMCSamples` and `NestedSamples`

### LINKED ISSUE #199: Create more general `Samples` that is parent to both `MCMCSamples` and `NestedSamples`
Currently `MCMCSamples` is the parent of `NestedSamples`. For many things this works quite nicely, since `NestedSamples` tends to be the one with more features. However, there are a few things that are actually quite specific for `MCMCSamples`, e.g. coming from different chains. Hence, it would be nice to have a more general `Samples` class that is parent to both `MCMCSamples` and `NestedSamples`, and to have some MCMC specific functionality to the new `MCMCSamples`.

In that context, it would be nice to have the following MCMC features:
* Move the `burn_in` to the new `MCMCSamples` (has nothing to do with `NestedSamples`.
* Keep track of the chain number for each sample point (allows separating the complete dataset into separate chains).
* Have some functions for convergence statistics of MCMC chains.
* Others?

lukashergt:
This splits what previously was `MCMCSamples` into a more general `Samples` and a new `MCMCSamples` which inherits from `Samples`. `NestedSamples` also only inherits from `Samples` and no longer from `MCMCSamples`.

In the long run, this setup should allow us to keep/create more MCMC specific functionality separate from NS. Examples are `burn_in`, keeping track of separate MCMC chains, convergence criteria for MCMC.

Fixes #199 

# Checklist:


anesthetic/samples.py
```diff
@@ -20,22 +21,19 @@
 from anesthetic.weighted_pandas import WeightedDataFrame, WeightedSeries
 
 
-class MCMCSamples(WeightedDataFrame):
-    """Storage and plotting tools for MCMC samples.
+class Samples(WeightedDataFrame):
+    """Storage and plotting tools for general samples.
 
     Extends the pandas.DataFrame by providing plotting methods and
     standardising sample storage.
```
williamjameshandley, line 28: Could we now add a bit more detail about exactly what it extends, e.g. the logL and tex functions (no need to document limits in light of the upcoming #200 upgrades)

anesthetic/samples.py
```diff
@@ -62,57 +60,56 @@ class MCMCSamples(WeightedDataFrame):
...
-    burn_in: int or float
-        Discards the first integer number of nsamples if int
-        or the first fraction of nsamples if float.
-        Only works if `root` provided and if chains are GetDist compatible.
-        default: False
-
     """
 
+    _metadata = WeightedDataFrame._metadata + ['tex', 'limits', 'root',
+                                               'label']
```
williamjameshandley, line 66: Should root be here if we're removing the reading functionality? This may likely change if we implement #208

anesthetic/samples.py
```diff
@@ -389,39 +386,95 @@ def importance_sample(self, logL_new, action='add', inplace=False):
...
-        return new
+    Extends the pandas.DataFrame by providing plotting methods and
+    standardising sample storage.
 
-    _metadata = WeightedDataFrame._metadata + ['tex', 'limits',
-                                               'root', 'label']
+    Example plotting commands include
+        - ``mcmc.plot_1d(['paramA', 'paramB'])``
+        - ``mcmc.plot_2d(['paramA', 'paramB'])``
+        - ``mcmc.plot_2d([['paramA', 'paramB'], ['paramC', 'paramD']])``
```
williamjameshandley, line 399: We should update this docstring so that it makes clearer what MCMCSamples adds

williamjameshandley (changes requested):
A relatively straightforward reorganisation, which definitely seperates out the logically separate portions of MCMCSamples and NestedSamples. I suspect that #208 will make the code even cleaner (reducing the duplication of root reading). 

It also opens the way to a refactor into additional classes in MetropolisSamples or HamiltonianSamples. On the nested sampling side, we might like to think about having a class for samples where you have birth contours (and therefore have much more functionality), and one where you just have nlive (for legacy chains/nested samplers which don't support this). Not sure what we'd call that though. BasicNestedSamples/SimpleNestedSamples and NestedSamples?

I think my only major request to consider in addition to the inline comments is whether we should change the tests which currently check `MCMCSamples` to `Samples`, except for where we're explicitly testing MCMC functionality. This may prove difficult as most of our test suite is reliant on loading from chains, although of course you could always write a fixture which loaded an MCMCSamples/NestedSamples object and then downgraded them to Samples.

anesthetic/samples.py
lukashergt: Agreed, updated in [02aae9e](https://github.com/williamjameshandley/anesthetic/pull/209/commits/02aae9edc34c2a861cc250c66dcf77e17821ccf5).

anesthetic/samples.py
lukashergt: > We should update this docstring so that it makes clearer what MCMCSamples adds

So far it barely adds anything, but that might change in the future.
I've adapted the docstring to match the `NestedSamples` one for now in [69b05dd](https://github.com/williamjameshandley/anesthetic/pull/209/commits/69b05dd529e480feb175e83ebdbaca843d708305).

anesthetic/samples.py
lukashergt: Isn't that essentially covered by the corresponding kwargs as listed under `Parameters`?

williamjameshandley: Agreed.

williamjameshandley (approved):

lukashergt:
> I think my only major request to consider in addition to the inline comments is whether we should change the tests which currently check `MCMCSamples` to `Samples`, except for where we're explicitly testing MCMC functionality. This may prove difficult as most of our test suite is reliant on loading from chains, although of course you could always write a fixture which loaded an MCMCSamples/NestedSamples object and then downgraded them to Samples.

I've implemented this for a big part in [57ce686](https://github.com/williamjameshandley/anesthetic/pull/209/commits/57ce686f379c8459edec380cea601bd56cbd621f). The later tests indeed rely too much on loading from the example data chains, so I have left them as is.

williamjameshandley (approved):

Issue #201, opened 2022-08-08
Explicit user separation of ncompress and nplot


williamjameshandley:
As discussed in #180 and #94, there is a difference between the number of plotting points (nplot) and the number of points we use to estimate kdes (ncompress). Runtime goes as `nplot*ncompress`, so any automated reduction in one can improve accuracy in the other for a given computational time. In principle one only needs a few targeted plotting points guided by the posterior/tails (e.g. extra plotting points at the edges, and do not need to overpopulate within 1 sigma as nested sampling automatically does). This would be a breaking change to the current ncompress interface

PR #212, opened 2022-08-11
remove `limits` attribute and `xmin` and `xmax` kwargs

### LINKED ISSUE #200: Consider removing user ability to set `.limits`

lukashergt:
This PR addresses #200 and 
1) removes the `limits` attribute of `Samples`, `MCMCSamples`, and `NestedSamples`, and
2) removes the `xmin` and `xmax` (and `ymin` and `ymax` kwargs passed to the plotting functions.

Fixes #200 

# Checklist:

### DIFF
--- a/anesthetic/plot.py
+++ b/anesthetic/plot.py
@@ -390,6 +390,13 @@ def fastkde_plot_1d(ax, data, *args, **kwargs):
         values at which to draw iso-probability lines.
         optional, default [0.95, 0.68]
 
+    q: int or float or tuple
+        Quantile to determine the data range to be plotted.
+        - 0: full data range, i.e. q=0 --> quantile range (0, 1)
+        - int: `q`-sigma data range, e.g. q=1 --> quantile range (0.16, 0.84)
+        - float: percentile, e.g. q=0.68 --> quantile range  (0.16, 0.84)
+        - tuple: quantile range, e.g. (0.16, 0.84)
+
     facecolor: bool or string
         If set to True then the 1d plot will be shaded with the value of the
         ``color`` kwarg. Set to a string such as 'blue', 'k', 'r', 'C1' ect.
@@ -435,7 +442,7 @@ def fastkde_plot_1d(ax, data, *args, **kwargs):
     except NameError:
         raise ImportError("You need to install fastkde to use fastkde")
     p /= p.max()
-    i = ((x > quantile(x, q[0], p)) & (x < quantile(x, q[1], p)))
+    i = ((x > quantile(x, q[0], p)) & (x < quantile(x, q[-1], p)))
 
     area = np.trapz(x=x[i], y=p[i]) if density else 1
     ans = ax.plot(x[i], p[i]/area, color=color, *args, **kwargs)
@@ -479,6 +486,13 @@ def kde_plot_1d(ax, data, *args, **kwargs):
     levels: list
         values at which to draw iso-probability lines.
         optional, default [0.95, 0.68]
+
+    q: int or float or tuple
+        Quantile to determine the data range to be plotted.
+        - 0: full data range, i.e. q=0 --> quantile range (0, 1)
+        - int: `q`-sigma data range, e.g. q=1 --> quantile range (0.16, 0.84)
+        - float: percentile, e.g. q=0.68 --> quantile range  (0.16, 0.84)
+        - tuple: quantile range, e.g. (0.16, 0.84)
 
     facecolor: bool or string
         If set to True then the 1d plot will be shaded with the value of the
@@ -529,7 +543,7 @@ def kde_plot_1d(ax, data, *args, **kwargs):
     kde = gaussian_kde(x, weights=w, bw_method=bw_method)
     p = kde(x)
     p /= p.max()
-    i = ((x > quantile(x, q[0], w)) & (x < quantile(x, q[1], w)))
+    i = ((x > quantile(x, q[0], w)) & (x < quantile(x, q[-1], w)))
     bw = np.sqrt(kde.covariance[0, 0])
     pp = cut_and_normalise_gaussian(x[i], p[i], bw=bw,
                                     xmin=data.min(), xmax=data.max())
@@ -570,6 +584,13 @@ def hist_plot_1d(ax, data, *args, **kwargs):
     weights: np.array, optional
         Sample weights.
 
+    q: int or float or tuple
+        Quantile to determine the data range to be plotted.
+        - 0: full data range, i.e. q=0 --> quantile range (0, 1)
+        - int: `q`-sigma data range, e.g. q=1 --> quantile range (0.16, 0.84)
+        - float: percentile, e.g. q=0.68 --> quantile range  (0.16, 0.84)
+        - tuple: quantile range, e.g. (0.16, 0.84)
+
     Returns
     -------
     patches : list or list of lists
@@ -596,7 +617,7 @@ def hist_plot_1d(ax, data, *args, **kwargs):
     q = kwargs.pop('q', 5)
     q = quantile_plot_interval(q=q)
     xmin = quantile(data, q[0], weights)
-    xmax = quantile(data, q[1], weights)
+    xmax = quantile(data, q[-1], weights)
 
     if bins in ['knuth', 'freedman', 'blocks']:
         try:
@@ -850,6 +871,13 @@ def hist_plot_2d(ax, data_x, data_y, *args, **kwargs):
         mass. If None defaults to usual matplotlib.axes.Axes.hist2d colouring.
         optional, default None
 
+    q: int or float or tuple
+        Quantile to determine the data range to be plotted.
+        - 0: full data range, i.e. q=0 --> quantile range (0, 1)
+        - int: `q`-sigma data range, e.g. q=1 --> quantile range (0.16, 0.84)
+        - float: percentile, e.g. q=0.68 --> quantile range  (0.16, 0.84)
+        - tuple: quantile range, e.g. (0.16, 0.84)
+
     Returns
     -------
     c: matplotlib.collections.QuadMesh
@@ -868,9 +896,9 @@ def hist_plot_2d(ax, data_x, data_y, *args, **kwargs):
     q = kwargs.pop('q', 5)
     q = quantile_plot_interval(q=q)
     xmin = quantile(data_x, q[0], weights)
-    xmax = quantile(data_x, q[1], weights)
+    xmax = quantile(data_x, q[-1], weights)
     ymin = quantile(data_y, q[0], weights)
-    ymax = quantile(data_y, q[1], weights)
+    ymax = quantile(data_y, q[-1], weights)
     rge = kwargs.pop('range', ((xmin, xmax), (ymin, ymax)))
 
     if levels is None:
@@ -986,7 +1014,7 @@ def quantile_plot_interval(q):
         if q > 0.5:
             q = 1 - q
         q = (q, 1-q)
-    return q
+    return tuple(np.sort(q))
 
 
 def normalize_kwargs(kwargs, alias_mapping=None, drop=None):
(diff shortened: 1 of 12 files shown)


anesthetic/plot.py
```diff
@@ -505,59 +496,52 @@ def kde_plot_1d(ax, data, *args, **kwargs):
...
     i = ((x > quantile(x, q[0], w)) & (x < quantile(x, q[1], w)))
-    if xmin is not None:
-        i = i & (x > xmin)
-    if xmax is not None:
-        i = i & (x < xmax)
-    sigma = np.sqrt(kde.covariance[0, 0])
-    pp = cut_and_normalise_gaussian(x[i], p[i], sigma, xmin, xmax)
+    bw = np.sqrt(kde.covariance[0, 0])
+    pp = cut_and_normalise_gaussian(x[i], p[i], bw=bw,
+                                    xmin=data.min(), xmax=data.max())
```
williamjameshandley, line 541: If I understand it, this is the first substantive change -- defining limits by data. All the rest of the above is just making the ordering of kwarg processing consistent.

anesthetic/plot.py
```diff
@@ -686,23 +665,26 @@ def fastkde_contour_plot_2d(ax, data_x, data_y, *args, **kwargs):
                                            color=['c'],
                                            facecolor=['fc'],
                                            edgecolor=['ec']))
+
+    if len(data_x) == 0 or len(data_y) == 0:
+        return np.zeros(0), np.zeros(0), np.zeros((0, 0))
```
williamjameshandley, line 670: This will need to be covered by the tests, or removed if you don't think its necessary. Possibly by `samples[:0].plot.plot_kde_1d()`?

anesthetic/plot.py
```diff
@@ -790,31 +768,30 @@ def kde_contour_plot_2d(ax, data_x, data_y, *args, **kwargs):
...
                                            facecolor=['fc'],
                                            edgecolor=['ec']))
-    bw_method = kwargs.pop('bw_method', None)
-    xmin = kwargs.pop('xmin', None)
-    xmax = kwargs.pop('xmax', None)
-    ymin = kwargs.pop('ymin', None)
-    ymax = kwargs.pop('ymax', None)
+
+    if len(data_x) == 0 or len(data_y) == 0:
+        return np.zeros(0), np.zeros(0), np.zeros((0, 0))
```
williamjameshandley, line 773: ditto

anesthetic/plot.py
```diff
@@ -505,59 +496,52 @@ def kde_plot_1d(ax, data, *args, **kwargs):
...
 
-    if weights is not None:
-        data = data[weights != 0]
-        weights = weights[weights != 0]
-
     x, w = sample_compression_1d(data, weights, ncompress)
     kde = gaussian_kde(x, weights=w, bw_method=bw_method)
     p = kde(x)
     p /= p.max()
     i = ((x > quantile(x, q[0], w)) & (x < quantile(x, q[1], w)))
```
williamjameshandley, line 538: Could we raise an issue to improve the quantile treatment for defining x's, so that we drive 'five sigma quantiles' by iso-probability credibility regions, rather than cdfs?

anesthetic/plot.py
```diff
@@ -1038,6 +996,8 @@ def quantile_plot_interval(q):
                   '4sigma': 0.999936657516334,
                   '5sigma': 0.999999426696856}
         q = (1 - sigmas[q]) / 2
+    elif isinstance(q, int) and q >= 1:
+        q = (1 - erf(q / np.sqrt(2))) / 2
     if isinstance(q, float) or isinstance(q, int):
```
williamjameshandley, line 1001: This is a nice addition! Should we add q to the docstrings officially? Tests look good too.

anesthetic/read/cobayamcmcreader.py
```diff
@@ -32,28 +32,21 @@ def paramnames(self):
...
         """Read <root>_1.txt in getdist format."""
         data = np.array([])
         for chains_file in self.chains_files:
             data_i = np.loadtxt(chains_file)
             if burn_in:
-                if 0 < burn_in < 1:
+                if isinstance(burn_in, float) and 0 < burn_in < 1:
                     index = int(len(data_i) * burn_in)
-                elif type(burn_in) is int and 1 < burn_in < len(data_i):
+                elif isinstance(burn_in, int) and 1 < burn_in < len(data_i):
```
williamjameshandley, line 43: Does it need to be an integer? Pandas slicing will probably handle floats, and there's no need to restrict this to users (it would be annoying to have to wrap an `int(...)` around an automatically computed argument. You're also missing out the case where burn_in=1.

anesthetic/read/getdistreader.py
```diff
@@ -36,41 +36,21 @@ def paramnames(self):
...
         """Read <root>_1.txt in getdist format."""
         data = np.array([])
         for chains_file in self.chains_files:
             data_i = np.loadtxt(chains_file)
             if burn_in:
-                if 0 < burn_in < 1:
+                if isinstance(burn_in, float) and 0 < burn_in < 1:
                     index = int(len(data_i) * burn_in)
-                elif type(burn_in) is int and 1 < burn_in < len(data_i):
+                elif isinstance(burn_in, int) and 1 < burn_in < len(data_i):
```
williamjameshandley, line 47: see above. Annoying this code is repeated -- hopefully something we'll resolve with a better inheritance hierarchy in the resolution to #208

anesthetic/plot.py
```diff
@@ -362,10 +363,6 @@ def make_2d_axes(params, **kwargs):
                 for a in ax_:
                     a.tick_params('x', bottom=False, top=False,
                                   labelbottom=False, labeltop=False)
-            else:
-                raise ValueError(
-                    "ticks=%s was requested, but ticks can only be one of "
-                    "['outer', 'inner', None]." % ticks)
```
williamjameshandley, line 368: I think I agree that this can be safely deleted, and the new test seems to confirm.

tests/test_plot.py
```diff
@@ -267,26 +272,12 @@ def test_kde_plot_1d(plot_1d):
...
-        xmax = 0.5
-        line, = plot_1d(ax, data, xmax=xmax)
-        assert (line.get_xdata() <= xmax).all()
-
-        # Check xmin and xmax
-        line, = plot_1d(ax, data, xmin=xmin, xmax=xmax)
-        assert (line.get_xdata() <= xmax).all()
-        assert (line.get_xdata() >= xmin).all()
-        plt.close("all")
-
```
williamjameshandley, line 285: Always makes me uncomfortable to see tests deleted, but here we can!

tests/test_plot.py
```diff
@@ -670,24 +632,27 @@ def test_scatter_plot_2d():
...
-    assert q2 == 0.5 + sigmas[1] / 2
+    qi1, qi2 = quantile_plot_interval(q=sigmas[0])
+    qs1, qs2 = quantile_plot_interval(q=sigmas[1])
+    assert qi1 == pytest.approx(0.5 - sigmas[2] / 2)
+    assert qi2 == pytest.approx(0.5 + sigmas[2] / 2)
+    assert qs1 == pytest.approx(0.5 - sigmas[2] / 2)
+    assert qs2 == pytest.approx(0.5 + sigmas[2] / 2)
 
 
-@pytest.mark.parametrize('floats', [0, 1, 0.1, 0.9])
```
williamjameshandley, line 690: The removal of `1` here because of its overlap with the new integer framework is a bit of a problem. How else could users specify 'please plot everything'? Perhaps a float (although that does also seem a bit dangerous)?

tests/test_samples.py
```diff
@@ -407,7 +402,7 @@ def test_plot_1d_colours():
 def test_astropyhist():
     np.random.seed(3)
     ns = NestedSamples(root='./tests/example_data/pc')
-    ns.plot_1d(['x0', 'x1', 'x2', 'x3'], kind='hist_1d', plotter='astropyhist')
```
williamjameshandley, line 410: Might some people want to use the astropyhist plotter for non-knuth based reasons? Should we test both?

tests/test_utils.py
```diff
@@ -9,7 +9,14 @@
...
                               triangular_sample_compression_2d,
-                              insertion_p_value)
+                              insertion_p_value, compress_weights)
+
+
+def test_compress_weights():
+    w = compress_weights(w=np.ones(10), u=None)
+    assert_array_equal(w, np.ones(10))
+    w = compress_weights(w=None, u=np.random.rand(10))
+    assert_array_equal(w, np.ones(10))
```
williamjameshandley, line 19: Always good to see new tests. Is there a specific reason we needed this for this PR?

williamjameshandley (commented):
Looking good -- only some minor comments to consider. Very nice to see so much extraneous code being removed. This should be much neater. I guess the next step to consider in plotting is #201

anesthetic/plot.py
lukashergt: I was actually trying to actively hit this, and it turned out that it is actually inaccessible...

anesthetic/plot.py
lukashergt: See issue #213

anesthetic/plot.py
lukashergt: Yes, although for most cases, the `limits` were previously automatically computed as `data.min(), data.max()`, so it is not a dramatic change.

anesthetic/plot.py
lukashergt: Sorry, this is not a new introduction from me, but a reordering to have more consistency between the plotting functions. 

Is this needed? I was tempted to drop it, but can't claim to know why it was introduced in the first place... In what scenario do you have an empty column? 

I think this should be and is all handled higher um in `Samples.plot_2d`, see e.g. this line:
```python
                    if x in self and y in self and lkwargs['kind'] is not None:
```

I'll drop it. The tests don't cover it after all...

anesthetic/plot.py
lukashergt: ditto

tests/test_plot.py
lukashergt: Yes, I agree. And it might mean encountering issues further along, but we are fundamentally changing the way we are handling data vs plotting limits, and these tests no longer cover that.

tests/test_plot.py
lukashergt: The "please plot everything" is still covered by the `0`. So far `0` and `1` were synonymous, both returning `(0, 1)`...

tests/test_utils.py
lukashergt: The usual issue with dropping a bunch of lines: the code coverage can drop causing the CI checks to fail. This test helps bring the coverage back up, such that CI checks pass.

anesthetic/plot.py
lukashergt: Removed in [6d0d8a3](https://github.com/williamjameshandley/anesthetic/pull/212/commits/6d0d8a3b823849e405569a0ea01bf17016f06064).

anesthetic/plot.py
lukashergt: Removed in [6d0d8a3](https://github.com/williamjameshandley/anesthetic/pull/212/commits/6d0d8a3b823849e405569a0ea01bf17016f06064).

tests/test_samples.py
lukashergt: I think the only added functionality of astropyhist is in `'knuth'`, `'freedman'` and `'blocks'`, which are all tested here. Note that it is the `bins` arg that indicates the use of astropyhist.

anesthetic/plot.py
lukashergt: Added in [62f1d9c](https://github.com/williamjameshandley/anesthetic/pull/212/commits/62f1d9c928c108ac8b913865c0be1a7bea779293).

williamjameshandley (dismissed):

williamjameshandley (approved):

PR #216, opened 2022-08-12
Have separate `ncompress` and `nplot`

### LINKED ISSUE #201: Explicit user separation of ncompress and nplot

lukashergt:
This PR 
* introduces an `nplot` kwarg that determines the number of plotting points
  - reduces computation cost for KDE plots which goes as `ncompress x nplot`

Replaces #181 

Fixes #201

# Checklist:


anesthetic/plot.py
```diff
@@ -539,27 +544,28 @@ def kde_plot_1d(ax, data, *args, **kwargs):
...
-    ans = ax.plot(x[i], pp/area, color=color, *args, **kwargs)
+    area = np.trapz(x=x, y=pp) if density else 1
+    ans = ax.plot(x, pp/area, color=color, *args, **kwargs)
 
     if facecolor and facecolor not in [None, 'None', 'none']:
         if facecolor is True:
             facecolor = color
-        c = iso_probability_contours_from_samples(pp, contours=levels,
-                                                  weights=w)
+        c = iso_probability_contours_from_samples(pp, contours=levels)
```
williamjameshandley, line 564: This should be iso_probability_contours, since they are no longer samples

lukashergt: Done in [2a04ff3](https://github.com/williamjameshandley/anesthetic/pull/216/commits/2a04ff36263a4667f2028bdae25ba3fe31b7ee8b).

I've also taken away the option to not provide weights to `iso_probability_contours_from_samples`. This should help emphasize that this function is only meant for weighted samples.

anesthetic/plot.py
```diff
@@ -481,7 +481,12 @@ def kde_plot_1d(ax, data, *args, **kwargs):
         Sample weights.
 
     ncompress: int, optional
-        Degree of compression. Default 1000
+        Degree of compression.
+        Default 10000
```
williamjameshandley, line 485: Should the default be `None`, which compresses 'optimally' to the channel capacity? 10000 seems a little arbitrary, and at this value makes one wonder if we even need compression at all! How does the speed compare with no compression on a cosmological run?.

anesthetic/plot.py
williamjameshandley: This isn't quite correct -- you can provide unweighted samples. The issue is that the plotting points are not necessarily samples, so you need a different algorithm to compute the iso probability contours.

lukashergt: Reverted

anesthetic/plot.py
lukashergt: > Should the default be None, which compresses 'optimally' to the channel capacity? 10000 seems a little arbitrary, and at this value makes one wonder if we even need compression at all! 

Agreed on all accounts. We have a few decisions to make:
1) Should `nplot` be the same for 1D and 2D? I think it would be better to have separate `nplot_1d` and `nplot_2d`, since aesthetically they have very different minimum requirements and speedwise 1D is much faster (regardless of `ncompress`) than 2D.
2) Which default `ncompress` for 1D? From the timings I'd say to either go with channel capacity or with no compression at all.
3) Which default `ncompress` for 2D? No compression is definitely not an option, too slow. We could go with:
   * `ncompress=1000`: current default, good for quick look, for paper plot typically insufficient for non-standard-gaussian contours
   * `ncompress=int(ns.neff())`: potentially enough for paper plots, but maybe too slow (especially when plotting multiple contours)

Thoughts?



> How does the speed compare with no compression on a cosmological run?.

See timings below.



&nbsp;

### `kde_plot_1d`

For the 1D case `ncompress` is negligible for computation time for an 86042 sample cosmology run:
```python
ns.plot_2d(axes, kind='kde_1d', ...)
```
* `nplot=200, ncompress=1000`: 421ms +- 9ms
* `nplot=200, ncompress=10000`: 480ms +- 10ms
* `nplot=200, ncompress=int(ns.neff())`: 917ms +- 56ms   (`ncompress=16874`)
* `nplot=200, ncompress=100000`: 1290ms += 20ms   (`ncompress=86042`)



&nbsp;

### `kde_plot_2d`

For the 2D case `ncompress` is more important regarding computation time. For an 86042 sample cosmology run:
```python
ns.plot_2d(axes, kind='kde_2d', ...)
```
* `nplot=1000, ncompress=1000`: 3.09s +- 0.08s
* `nplot=1000, ncompress=10000`: 6.58s +- 0.20s
* `nplot=1000, ncompress=int(ns.neff())`: 9.29s +- 0.22s   (`ncompress=16874`)
* `nplot=1000, ncompress=100000`: 32.5s +- 0.30s   (`ncompress=86042`)

williamjameshandley: Many thanks for these excellent data!

> 1\. Should `nplot` be the same for 1D and 2D? I think it would be better to have separate `nplot_1d` and `nplot_2d`, since aesthetically they have very different minimum requirements and speedwise 1D is much faster (regardless of `ncompress`) than 2D.

No. I agree.


> 2\. Which default `ncompress` for 1D? From the timings I'd say to either go with channel capacity or with no compression at all.

Dealer's choice here. Tempting to go with the less interfering of the two options (i.e. no compression).


> 3\. Which default `ncompress` for 2D? No compression is definitely not an option, too slow. We could go with:
>     
>     * `ncompress=1000`: current default, good for quick look, for paper plot typically insufficient for non-standard-gaussian contours
>     * `ncompress=int(ns.neff())`: potentially enough for paper plots, but maybe too slow (especially when plotting multiple contours)

Here my answer is a bit more involved.

- The current commit switches from a tricontour-based samples approach to a meshgrid-based one. 
- The advantage of the previous setup was that using samples to plot 1 and 2-sigma contours naturally focuses points on the region that we need (which meshgrid doesn't, particularly for correlated distributions). In principle, one would only need ~O(20) plotting samples to adequately populate 2-sigma contours. If we had this set-up we could set nplot for an 2 sigma plot to be 370 = (1/1-sigmas['3sigma']), or something proportional (which would claw back a factor of 3 at least). I imagine that this could be set _much_ lower in practice (perhaps only going half a sigma more than necessary).
- The disadvantage was the creation of artefacts, since a triangulation along a banana doesn't handle the non-convex negative space very well.

What we should do is to combine the two approaches, namely to have a meshgrid superimposed upon samples for plotting purposes. The only thing you need to be careful with is the `iso_probability_contours_from_samples`, which historically we were able to use by setting the meshgrid weights which are passed in to zero. Previously we used random samples rather than a meshgrid, but an advantage of the meshgrid would be to remove jagged edges in the event the quantile reached the edge.

However, I think we can go one better than random/meshgrid. We definitely want to avoid jagged edges, so having four corner points on boundaries of the plot is a good idea. The meshgrid/random samples could also be in the decorrelated space which scaled_triangulation uses.

The following strategy looks to be a good way of populating the space with plotting points with fewer than 200 samples.

```python
from matplotlib.tri import Triangulation, UniformTriRefiner
import numpy as np

samples = NestedSamples(root='./tests/example_data/pc')

n = 100
x = samples.compress(n).x0
y = samples.compress(n).x1
xmin, xmax = samples.x0.quantile([0.00001, 0.99999])
ymin, ymax = samples.x1.quantile([0.00001, 0.99999])

x = (x-xmin)/(xmax-xmin)
y = (y-ymin)/(ymax-ymin)

x = np.concatenate([x, [0,1,0,1]])
y = np.concatenate([y, [0,1,1,0]])
x = np.concatenate([x, np.random.rand(50)])
y = np.concatenate([y, np.random.rand(50)])

tri = Triangulation(x, y)
x = xmin + (xmax-xmin)*x
y = ymin + (ymax-ymin)*y
tri = Triangulation(x, y, tri.triangles)

ax = samples.plot.scatter('x0','x1')
ax.triplot(tri, color='k', lw=0.5)
ax.figure.tight_layout()
```
tri

I think that in principle one could be even smarter than this, e.g. using temperature to spread the plotting samples, but for now this should be better than what we're currently doing.

williamjameshandley (changes requested):
Many thanks for these excellent data!

> 1\. Should `nplot` be the same for 1D and 2D? I think it would be better to have separate `nplot_1d` and `nplot_2d`, since aesthetically they have very different minimum requirements and speedwise 1D is much faster (regardless of `ncompress`) than 2D.

No. I agree.


> 2\. Which default `ncompress` for 1D? From the timings I'd say to either go with channel capacity or with no compression at all.

Dealer's choice here. Tempting to go with the less interfering of the two options (i.e. no compression).


> 3\. Which default `ncompress` for 2D? No compression is definitely not an option, too slow. We could go with:
>     
>     * `ncompress=1000`: current default, good for quick look, for paper plot typically insufficient for non-standard-gaussian contours
>     * `ncompress=int(ns.neff())`: potentially enough for paper plots, but maybe too slow (especially when plotting multiple contours)

Here my answer is a bit more involved.

- The current commit switches from a tricontour-based samples approach to a meshgrid-based one. 
- The advantage of the previous setup was that using samples to plot 1 and 2-sigma contours naturally focuses points on the region that we need (which meshgrid doesn't, particularly for correlated distributions). In principle, one would only need ~O(20) plotting samples to adequately populate 2-sigma contours. If we had this set-up we could set nplot for an 2 sigma plot to be 370 = (1/1-sigmas['3sigma']), or something proportional (which would claw back a factor of 3 at least). I imagine that this could be set _much_ lower in practice (perhaps only going half a sigma more than necessary).
- The disadvantage was the creation of artefacts, since a triangulation along a banana doesn't handle the non-convex negative space very well.

What we should do is to combine the two approaches, namely to have a meshgrid superimposed upon samples for plotting purposes. The only thing you need to be careful with is the `iso_probability_contours_from_samples`, which historically we were able to used by setting the meshgrid weights which are passed in to zero. Previously we used random samples rather than a meshgrid, but an advantage of the meshgrid would be to remove jagged edges in the event the quantile reached the edge.

However, I think we can go one better with random+meshgrid. We definitely want to avoid jagged edges, so having four corner points on boundaries of the plot is a good idea.

The following strategy looks to be a good way of populating the space with plotting points with fewer than 200 samples.

```python
from matplotlib.tri import Triangulation, UniformTriRefiner
import numpy as np

samples = NestedSamples(root='./tests/example_data/pc')

n = 100
x = samples.compress(n).x0
y = samples.compress(n).x1
xmin, xmax = samples.x0.quantile([0.00001, 0.99999])
ymin, ymax = samples.x1.quantile([0.00001, 0.99999])

x = (x-xmin)/(xmax-xmin)
y = (y-ymin)/(ymax-ymin)

x = np.concatenate([x, [0,1,0,1]])
y = np.concatenate([y, [0,1,1,0]])
x = np.concatenate([x, np.random.rand(50)])
y = np.concatenate([y, np.random.rand(50)])

tri = Triangulation(x, y)
x = xmin + (xmax-xmin)*x
y = ymin + (ymax-ymin)*y
tri = Triangulation(x, y, tri.triangles)

ax = samples.plot.scatter('x0','x1')
ax.triplot(tri, color='k', lw=0.5)
ax.figure.tight_layout()
```
tri

I think that in principle one could be even smarter than this, e.g. using temperature to spread the plotting samples, but for now this should be better than what we're currently doing.

lukashergt:
> Here my answer is a bit more involved.

I think your answers are more to the question "What should we use as default `nplot`?" rather than "What should we use as default `ncompress`?" The timings I've listed are very much about `ncompress`, not about `nplot`. 

I've left the `nplot` at the previous default of 1000, but we can probably reduce that without any problems. I was just being conservative there. The only reason that I increased the default for `ncompress` is that (a) it makes a lot of sense in the 1D case and (b) increasing `ncompress` has gotten cheaper given a fixed `nplot`.


> The advantage of the previous setup was that using samples to plot 1 and 2-sigma contours naturally focuses points on the region that we need (which meshgrid doesn't, particularly for correlated distributions). 

I'm not sure I agree, which is partly the point of this PR. We end up with a lot of unneeded points within the 1-sigma contours, which would be better used for an overall increase in plot point density. We don't need the plotting points in the centre, we need them at the 1-to-2-sigma and the 2-to-3-sigma intersections. We need the centre points for the KDE computation (which is completely independent of the meshgrid), but not for the plotting, matplotlib's `contour` does the interpolations quite well. And especially when it comes to n-sigma contours, we would rather have a few points spread across the whole plotting area, rather than a lot of points only in the centre.

williamjameshandley:
> I think your answers are more to the question "What should we use as default `nplot`?" rather than "What should we use as default `ncompress`?" The timings I've listed are very much about `ncompress`, not about `nplot`.

I agree -- I got a little carried away since t ~ nplot * ncompress (I assume), so as you say any amount we can reduce nplot by will give us a bit more headroom with ncompress. 

In an ideal world, we would get each plot down to the sub-second level by default, which is a reasonable latency for interactive work. Obviously this is machine dependent, so if you're on a slow system 3s per plot is OK, but would ideally be lower. I would therefore go with `ncompress=1000` as a default, and we'll try and communicate in documentation that this can be increased for production plots.

I remember you saying that you had a branch with parallelisation, which would be good to get in at some point (though is not needed for 2.0.0).

> I'm not sure I agree, which is partly the point of this PR. We end up with a lot of unneeded points within the 1-sigma contours, which would be better used for an overall increase in plot point density. We don't need the plotting points in the centre, we need them at the 1-to-2-sigma and the 2-to-3-sigma intersections. We need the centre points for the KDE computation (which is completely independent of the meshgrid), but not for the plotting, matplotlib's `contour` does the interpolations quite well. And especially when it comes to n-sigma contours, we would rather have a few points spread across the whole plotting area, rather than a lot of points only in the centre.

I definitely see what you're saying here. More than a few points within 1sigma is wasteful. I wonder if there's any way to concentrate plotting points around the contours. It's worth bearing in mind that what the full set of dead points looks like as a triangulation

```python
from matplotlib.tri import Triangulation
import numpy as np

samples = NestedSamples(root='./tests/example_data/pc')

n = 100
x = samples.compress(n).x0
y = samples.compress(n).x1
xmin, xmax = samples.x0.quantile([0.00001, 0.99999])
ymin, ymax = samples.x1.quantile([0.00001, 0.99999])

x = (x-xmin)/(xmax-xmin)
y = (y-ymin)/(ymax-ymin)

x = np.concatenate([x, [0,1,0,1]])
y = np.concatenate([y, [0,1,1,0]])
x = np.concatenate([x, np.random.rand(50)])
y = np.concatenate([y, np.random.rand(50)])

tri = Triangulation(x, y)
x = xmin + (xmax-xmin)*x
y = ymin + (ymax-ymin)*y
tri = Triangulation(x, y, tri.triangles)

ax = samples.plot.scatter('x0','x1')
ax.triplot(tri, color='k', lw=0.5)
ax.figure.tight_layout()
```

tri

lukashergt:
> In an ideal world, we would get each plot down to the sub-second level by default

That is the case for a single 2D KDE plot. I've edited my timing report to clarify that those times were for a lower cosmo triangle, i.e. 15 2D KDE plots...

lukashergt:
> I remember you saying that you had a branch with parallelisation, which would be good to get in at some point (though is not needed for 2.0.0).

Unfortunately, with my latest work it turns out that this is not as useful anymore. The parallelisation works on the plotting side of things, but now that the plot points don't increase with the KDE points, the parallelisation doesn't really become efficient...

williamjameshandley:
> > In an ideal world, we would get each plot down to the sub-second level by default
> 
> That is the case for a single 2D KDE plot. I've edited my timing report to clarify that those times were for a lower cosmo triangle, i.e. 15 2D KDE plots...

Ah, then that's much more acceptable! I think 'sub second per plot' is still a reasonable upper bound to imagine, so we can use that to guide how efficient we want it to be.


> Unfortunately, with my latest work it turns out that this is not as useful anymore. The parallelisation works on the plotting side of things, but now that the plot points don't increase with the KDE points, the parallelisation doesn't really become efficient...

The other option is of course to parallelise with e.g. multiprocessing over the for loops in the plot_2d command.


> I wonder if there's any way to concentrate plotting points around the contours. It's worth bearing in mind that what the full set of dead points looks like as a triangulation

One way to partially concentrate efforts around the two sigma contours is to work with dead points, but to 'terminate early' (e.g. when only half the evidence has accrued). Subsampling this (without weights) within the quantile region gives an interesting grid of `nplot=200`:

```python
from matplotlib.tri import Triangulation, UniformTriRefiner
from anesthetic.utils import scaled_triangulation
import numpy as np

samples = NestedSamples(root='./tests/example_data/pc')

n = 200
xmin, xmax = samples.x0.quantile([0.00001, 0.99999])
ymin, ymax = samples.x1.quantile([0.00001, 0.99999])
mask = (samples.x0>xmin) & (samples.x0<xmax) & (samples.x1>ymin) & (samples.x1<ymax)
mask &= (samples.get_weights().cumsum()< 0.1 * samples.get_weights().sum())
subsamples = samples[mask].droplevel('weights').sample(n)
x = subsamples.x0
y = subsamples.x1

x = (x-xmin)/(xmax-xmin)
y = (y-ymin)/(ymax-ymin)
x = np.concatenate([x, [0,1,0,1]])
y = np.concatenate([y, [0,1,1,0]])
tri = Triangulation(x, y)
x = xmin + (xmax-xmin)*x
y = ymin + (ymax-ymin)*y
tri = Triangulation(x, y, tri.triangles)

ax = samples.plot.scatter('x0','x1')
ax.triplot(tri, color='k', lw=0.5)
ax.figure.tight_layout()

```
tri

lukashergt:
Just a log entry of the current n-sigma performance compared to the master branch. This works really well, now. And performance-wise this is miles better than anything we had previously thanks to `nplot` being decoupled from `ncompress`. In the current PR we can use all the data for the KDE and it is still much faster than the master branch which only uses about a quarter of the data.

## master branch

```python
ns = NestedSamples(root="finite_inflation_curved_space_data/cb3-0_pc1-20_cl2-9/lcdm/pcs2d1000_cl_hf_p18_TTTEEE_lowl_lowE_BK15_lcdm/pcs2d1000_TTTEEE_lowl_lowE_BK15_lcdm_polychord_raw/pcs2d1000_TTTEEE_lowl_lowE_BK15_lcdm")
fig = plt.figure(figsize=(10, 10))
fig, axes = make_2d_axes(['omega_b', 'omega_cdm', 'theta_s_1e2', 'tau_reio', 'logA', 'n_s'], tex=ns.tex, fig=fig, ticks='inner', upper=False, diagonal=False);
%time _ = ns.plot_2d(axes, label='spam', kind='kde_2d', ncompress=20000, levels=erf(np.arange(1, 6)[::-1]/np.sqrt(2)));
```
Wall time: 1min 50s
image


&nbsp;

## this PR

```python
ns = NestedSamples(root="finite_inflation_curved_space_data/cb3-0_pc1-20_cl2-9/lcdm/pcs2d1000_cl_hf_p18_TTTEEE_lowl_lowE_BK15_lcdm/pcs2d1000_TTTEEE_lowl_lowE_BK15_lcdm_polychord_raw/pcs2d1000_TTTEEE_lowl_lowE_BK15_lcdm")
fig = plt.figure(figsize=(10, 10))
fig, axes = make_2d_axes(['omega_b', 'omega_cdm', 'theta_s_1e2', 'tau_reio', 'logA', 'n_s'], tex=ns.tex, fig=fig, ticks='inner', upper=False, diagonal=False);
%time _ = ns.plot_2d(axes, label='spam', kind='kde_2d', nplot=1000, ncompress=100000, levels=erf(np.arange(1, 6)[::-1]/np.sqrt(2)));
```
Wall time: 35.1 s
image

williamjameshandley:
Very impressive! 

This conversation has been making me consider whether what we're doing at the `triangular_sample_compression_2d` step is as effective as it could be -- it effectively compresses onto a triangulation on the deadpoints, which will cover the tails well (hence the surprisingly good 5 sigma performance), but will also over-cover the center. However, designing a compression scheme better tailored for marginal 2d plots is probably beyond the scope of this PR.

I think for this PR we should be focusing on the API shift, which to my understanding is shifting from `ncompress` to `(ncompress, nplot_1d, nplot_2d)`, vs `(ncompress, nplot)`. The advantage of the former is that it logically separates `nplot` between the two cases, so that if a user passes in `nplot` as a global keyword to improve speed it doesn't degrade 1d kde.

I am definitely coming round to the meshgrid for plotting purposes. plotting a 2d grid of equivalent size shows that one does not gain a great deal:

```python
for l in np.linspace(xmin, xmax, int(np.ceil(np.sqrt(len(x))))):
    ax.plot([l, l], [ymin, ymax], 'r', lw=0.5)

for l in np.linspace(ymin, ymax, int(np.ceil(np.sqrt(len(x))))):
    ax.plot([xmin, xmax], [l, l], 'r', lw=0.5)
```

tri

My only concern is how well meshgrid will perform on a knife-edge like the excellent example in #193. That use-case alone could still merit the combined approach.

lukashergt:
> My only concern is how well meshgrid will perform on a knife-edge like the excellent example in #193. That use-case alone could still merit the combined approach.

This was indeed something I wanted to look into more, too. The thing is, that that example in #193 is a very difficult case for KDE and neither the previous master default settings, nor the default settings of this PR will be good enough (see below). But to some extent that is simply down to this being a difficult case. Cranking up `ncompress` and `nplot` solves those issues. 

There might be some merit to the combined approach in such an extreme corner case. However, I didn't manage to get it to work, I ended up with some weird spiky/fuzzy contours. Maybe weights gone wrong somewhere...? Does `tricontour` care whether its plotting points are uniformly distributed or not? 

Should we leave such explorations to a future PR? This contains everything relevant for `2.0.0` (once I've updated the distinction between `nplot_1d` and `nplot_2d`.


# master default

image

# PR default

image

# master tweaked 

```python
ncompress=10000
```
image


# PR tweaked

```python
ncompress=10000
nplot=10000
```
image

anesthetic/utils.py
```diff
@@ -387,6 +387,8 @@ def sample_compression_1d(x, w=None, n=1000):
     x, w, array-like
         Compressed samples and weights
     """
+    if n is None:
+        return x, w
```
williamjameshandley, line 391: This change to the default differs from `compress_weights` and therefore `WeightedDataFrame.compress`. There, if `n=None` it compresses to the channel capacity. I see the logic of `None` being 'do not compress', but equally there's something convenient about calling `.compress()` and it compressing optimally. Whatever we choose, we should be consistent across compression functions.

williamjameshandley (commented):
I agree that we should press on with `nplot_1d` and `nplot_2d`, and leave tinkering with plotting to a future date.

anesthetic/utils.py
lukashergt: Should we expand with a `bool` option? `True` for channel capacity, `False` for no compression?

williamjameshandley: I like that -- could we have that convention for all compression functions?

lukashergt: implemented in [aba18fa](https://github.com/williamjameshandley/anesthetic/pull/216/commits/aba18fa7b048c5aad7251aa5b2d0caf4248c4303)

williamjameshandley (approved):

williamjameshandley:
Excellent -- let's get this merged. 2.0.0 is getting close. Does this also close #181?

Could you have a look through to see if there are any other plotting-related breaking changes? The only one I can think of is changing `fig, axes` outputs to just `axes` ones. Did we settle on whether we wanted any changes to the axes themselves?

lukashergt:
> Could you have a look through to see if there are any other plotting-related breaking changes? The only one I can think of is changing `fig, axes` outputs to just `axes` ones. 

Addressed in #223.

> Did we settle on whether we wanted any changes to the axes themselves?

Did you have specific examples in mind?

PR #217, opened 2022-08-14
Change to reading system

### LINKED ISSUE #208: Should we change the way we read files?

williamjameshandley:
# Description

This PR implements the new function `read_chains(...)` which replaces the previous `NestedSamples(root=...)`. `read_chains` calls the functions `read_polychord`, `read_multinest`, `read_getdist` etc. The classes are removed, so it is easier for people to write new functions for loading.

Closes #208 

# Checklist:


lukashergt (changes requested):
I love how this reduces the initialisation methods for `MCMCSamples` and `NestedSamples`.

We should add some tests, though, that check that the reading works as intended. Previously we were doing checks for `.yaml` file or `.paramnames` file ahead of reading. Now we only have a `try... except...` block, which can silently fail, as the currently failing tests highlight. In this case it is the incorrect reading of Cobaya data with the GetDist reader.

It would be good to check for correctness for each of the readers for the following:
* weights
* column names
* tex labels

anesthetic/read/chain.py
```diff
@@ -0,0 +1,29 @@
...
+    root: str
+        root name
+
+    Returns
+    -------
+    NestedSamples or MCMCSamples depending on auto-detection
+
+    """
+    errors = []
+    for read in [read_polychord, read_multinest, read_cobaya, read_getdist]:
```
lukashergt, line 22: Note the importance of the order here: `read_cobaya` has to come before `read_getdist`.

anesthetic/read/cobaya.py
```diff
@@ -0,0 +1,80 @@
...
+    This is the data file of the first chain. It should have as many
+    columns as there are parameters (sampled and derived) plus an
+    additional two corresponding to the weights (first column) and the
+    logposterior (second column). The first line should start with a # and
+    should list the parameter names corresponding to the columns. This
+    will be used as label in the pandas array.
+    """
+    with open(root + ".1.txt") as f:
+        header = f.readline()[1:]
+        paramnames = header.split()[2:]
```
lukashergt, line 21: Only split off the weights and likelihood values. Don't split off the last columns. For the example data here that might be tempting, but these columns will actually be different for runs with multiple likelihoods.

anesthetic/read/cobaya.py
```diff
@@ -0,0 +1,80 @@
...
+
+    """
+    burn_in = kwargs.pop('burn_in', None)
+
+    pattern = '.[0-9]*.txt'
+    chains_files = sorted(glob.glob(root + pattern))
+    if len(chains_files) > 1:
+        chain_numbers = [cf.split('.')[-2] for cf in chains_files]
+    else:
+        chain_numbers = [0]
```
lukashergt, line 51: I've adjusted this such that the chain numbers in `MCMCSamples` match the ones in the `.txt` files.

williamjameshandley: Nice catch!

anesthetic/read/cobaya.py
williamjameshandley: Agreed.

anesthetic/read/chain.py
williamjameshandley: This is subtle. Well spotted.

tests/test_gui.py
```diff
@@ -35,6 +35,7 @@ def test_gui():
     plotter.temperature.slider.set_val(2)
     assert plotter.temperature() == 100
     plotter.type.buttons.set_active(0)
+    plotter.samples
```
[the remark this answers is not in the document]
williamjameshandley: No. Removed in 4618df9eb02790627801156a39403a9581755a66

tests/test_reader.py
```diff
@@ -5,30 +5,28 @@
...
-    assert reader.paramnames() == (None, {})
-    with pytest.raises(NotImplementedError):
-        reader.samples()
-
-
 def test_read_getdist():
     np.random.seed(3)
-    mcmc = MCMCSamples(root='./tests/example_data/gd')
+    mcmc = read_getdist('./tests/example_data/gd')
+    assert 'chain' in mcmc
```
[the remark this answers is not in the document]
williamjameshandley: Done in f149bc4

anesthetic/read/chain.py
```diff
@@ -0,0 +1,29 @@
...
+    """Auto-detect chain type and read from file.
+
+    Parameters
+    ----------
+    root: str
+        root name
+
+    Returns
+    -------
+    NestedSamples or MCMCSamples depending on auto-detection
```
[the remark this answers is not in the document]
williamjameshandley: I've added burn in to 74fe4135bfe9276763d68755ac113f8acb80986d, as well as a note that all other args and kwargs are forwarded onto `NestedSamples` or `MCMCSamples` constructors. `burn_in` is the only argument specific to reading at the moment (note that `MCMCSamples` no longer has it now we've moved reading out of the constructor).

williamjameshandley:
> It would be good to check for correctness for each of the readers for the following:
> 
> * weights
> * column names
> * tex labels

It is _always_ a good idea to ask for this as a reviewer. Basically none of the weights or tex were being read in correctly, and unexpected behaviour was occurring like the fact that `glob.glob` doesn't sort. 

I have now added tests to all of the readers, and upgraded the `glob.glob` to a proper regex so that we're now no longer limited to 10 chains and we read the chain number from the regex for getdist and cobaya.

I have also dropped montepython support. This worked in a while, since when trying to `from montepython import analyze` it fails because of a bug in where it fails at an `import io_mp` step due to a syntax error (possibly from the upgrade from `python2` to `python3`), which has gone unfixed for years, so I'm not sure that even this functionality is ever used. It can be loaded as a gedist chain, and I have added a test which does this.

anesthetic/read/chain.py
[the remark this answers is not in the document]
lukashergt: > (note that `MCMCSamples` no longer has it now we've moved reading out of the constructor)

In the long run it might be nice to have some sort of `remove_burn_in` method, that correctly handles the `'chain'` key to remove from all chains, which slicing cannot do.

anesthetic/read/cobayamcmcreader.py
```diff
@@ -1,57 +0,0 @@
...
-        will be used as label in the pandas array. Uses GetDist's
-        `loadMCSamples` to infer the tex dictionary.
-        """
-        try:
-            with open(self.root + ".1.txt") as f:
-                header = f.readline()[1:]
-                paramnames = header.split()[2:]
-                s = loadMCSamples(file_root=self.root)
-                tex = {p.name: '$' + p.label + '$' for p in s.paramNames.names}
-                return paramnames, tex
```
lukashergt, line 31: Currently `read_cobaya` does not read in any TEX data. Was that dropped because of the upcoming #215?

anesthetic/read/chain.py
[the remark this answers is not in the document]
williamjameshandley: Definitely! Added as a reminder in #219

anesthetic/read/cobayamcmcreader.py
williamjameshandley: Best not to presume -- although I am now making good progress again with #215 (I may be able to get the original idea to work).

I didn't particularly like the dependency on getdist for file reading (ideally we'd do it natively), but you're right that this is an unnecessary removal of functionality. Restored and tested

anesthetic/read/cobaya.py
```diff
@@ -20,7 +20,12 @@ def read_paramnames(root):
     with open(root + ".1.txt") as f:
         header = f.readline()[1:]
         paramnames = header.split()[2:]
-        return paramnames
+        try:
+            s = loadMCSamples(file_root=self.root)
+            tex = {p.name: '$' + p.label + '$' for p in s.paramNames.names}
+            return paramnames, tex
+        except NameError:
+            return paramnames, {}
```
lukashergt, line 28: This is not actually working, yet. `loadMCSamples` was not imported and `self` does not exist, so this was silently failing. Probably better to put `import getdist` into the try block and only allow failing with `ImportError`.

tests/test_reader.py
```diff
@@ -61,10 +61,12 @@ def test_read_cobayamcmc():
         np.loadtxt("./tests/example_data/cb.2.txt", usecols=0)
     ))
     assert_array_equal(mcmc.weights, w)
-    params = rp_cb("./tests/example_data/cb")
+    params, tex = rp_cb("./tests/example_data/cb")
     params.append('logL')
     params.append('chain')
     assert_array_equal(mcmc.columns, params)
+    tex['chain'] = r'$n_\mathrm{chain}$'
+    assert mcmc.tex == tex
```
lukashergt, line 69: It might be better to have an explicit list of tex labels to test both `rp_cb` and `mcmc` against (and probably similarly for `read_getdist`), since this did not catch the failed reading in the current setup.

anesthetic/read/cobaya.py
williamjameshandley: Agreed. With the latest merge of master this is now corrected and tested.

tests/test_reader.py
williamjameshandley: Agreed. We no longer call the readers, and test explicitly against the tex. Tying the tex labels to a column in #215 should hopefully make this much safer.

williamjameshandley:
> I did some important changes in [b0d7baf](https://github.com/williamjameshandley/anesthetic/pull/217/commits/b0d7baf26ae6635d557275f906e2d9c23f7066b2) which seems to have been reverted by a master merge...

I can't spot them -- in changing to regular expressions from globbing I moved it back to the original style, but I thought I retained all the fixes to the errors which you spotted (ordering of chain reading, keeping all the cobaya columns, synchronising the chain index with the file index)

lukashergt:
> > I did some important changes in [b0d7baf](https://github.com/williamjameshandley/anesthetic/pull/217/commits/b0d7baf26ae6635d557275f906e2d9c23f7066b2) which seems to have been reverted by a master merge...
> 
> I can't spot them -- in changing to regular expressions from globbing I moved it back to the original style, but I thought I retained all the fixes to the errors which you spotted (ordering of chain reading, keeping all the cobaya columns, synchronising the chain index with the file index)

Not sure what was going on, but now things seem to be working. Neat!

lukashergt (approved):
Nice, not sure in what order the current PRs should best be merged...?

williamjameshandley:
Let's try this one first (you may need to re-approve #210 if it requires a manual merge).

Many thanks @lukashergt

PR #218, opened 2022-08-15
Issue with plot labels

williamjameshandley:
# Description

#207 introduced a bug whereby `samples.plot_2d(...)` no longer has the correct tex in all of the labels. This arises because `samples.plot(...)` defaults to using the column label, which can overwrite the labels set up by `make_2d_axes`. This fixes that by passing `xlabel` explicitly through `samples.plot` in `samples.plot_2d(...)`

# Checklist:


lukashergt (approved):
Nice quick bug catch.

PR #221, opened 2022-08-20
100% coverage

williamjameshandley:
# Description

With the latest merge of #217, we are near to 100% coverage.

This PR makes the relatively strong assessment "If it's not being tested, we don't know what bug it's fixing", and removes sections of code which guard cases for which I can't see how these would occur in production code. If these re-emerge we should be able to construct a failing test from this branch and resurrect the fix from the diff of this PR.

# Checklist:

### DIFF
--- a/tests/test_reader.py
+++ b/tests/test_reader.py
@@ -45,6 +45,17 @@ def test_read_getdist():
     mcmc.plot_1d(['x0', 'x1', 'x2', 'x3'])
     plt.close("all")
 
+    os.rename('./tests/example_data/gd.paramnames',
+              './tests/example_data/gd.paramnames_')
+    mcmc = read_getdist('./tests/example_data/gd')
+    os.rename('./tests/example_data/gd.paramnames_',
+              './tests/example_data/gd.paramnames')
+
+    params = [0, 1, 2, 3, 4, 'logL', 'chain']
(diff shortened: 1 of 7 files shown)


tests/test_reader.py
```diff
@@ -45,6 +45,17 @@ def test_read_getdist():
     mcmc.plot_1d(['x0', 'x1', 'x2', 'x3'])
     plt.close("all")
 
+    os.rename('./tests/example_data/gd.paramnames',
+              './tests/example_data/gd.paramnames_')
+    mcmc = read_getdist('./tests/example_data/gd')
+    os.rename('./tests/example_data/gd.paramnames_',
+              './tests/example_data/gd.paramnames')
```
williamjameshandley, line 52: I do wonder whether the equivalent code further down for polychord might have been causing the matplotlib thread-unsafe elements we were experiencing (and if we just made these permanent files, whether we could reinstate parallel test execution).

lukashergt: Might be, some other test reading when the needed files have been temporarily renamed... Happy to have some more test files and try bringing back parallel testing.

lukashergt (approved):
Nice, this will reduce PR unrelated code coverage failures in the future.

Feel free to merge or address the file renaming issue.

PR #222, opened 2022-08-20
Reinstate parallel tests

williamjameshandley:
# Description

This PR attempts to reinstate parallel tests under the hypothesis from #221 that one piece of thread unsafe code we have in master is the renaming of files in `tests/test_reader.py`. 

# Checklist:


williamjameshandley:
Thus far this is looking good. I will continue to merge in updates to master to see if it fails again.

williamjameshandley:
Also removed the `matplotlib_agg`, which I believe was a workaround for Travis and/or CircleCI. GitHub actions allows very straightforward environment variable setting.

williamjameshandley:
There's a [failing test](https://github.com/williamjameshandley/anesthetic/runs/7940292823?check_suite_focus=true) which looks like this hasn't fully fixed it. It does seem much more stable than before, but I can't see any other obvious points if instability (e.g. file writes).

Issue #223, opened 2022-08-21
change `fig, axes` output to just `axes`?


lukashergt:
Should we be changing the our plotting output from `fig, axes` to only `axes`?

There are two different places to think about this:
1) the output from `make_1d_axes` and `make_2d_axes`.
2) the output from `samples.plot_1d` and `samples.plot_2d`.

lukashergt:
> 1) the output from `make_1d_axes` and `make_2d_axes`.

This is related to #203. Currently I will often write:
```python
fig = plt.figure(figsize=(10, 10))
fig, axes = make_2d_axes(['x0', 'x1'], fig=fig)
```

I would want to keep the `fig, axes` output for `make_1d_axes` and `make_2d_axes`, but ensure that they both accept all the kwargs you would want to pass to `plt.figure()`, such that there is no need for the first line in my example.


&nbsp;

> 2) the output from `samples.plot_1d` and `samples.plot_2d`.

In this case I don't have that clear an opinion. The main reason for that is that I typically don't bother with the output from these. I tend to use `make_1d_axes` and `make_2d_axes` to set up my figure and axes, and then I simply pass the axes on:
```python
fig, axes = make_2d_axes(['x0', 'x1'])
samples1.plot_2d(axes)
samples2.plot_2d(axes)
```

Returning only `axes` would make things slimmer, but on the other hand there is something to be said for returning the same as `make_1d_axes` and `make_2d_axes`.

williamjameshandley:
My initial view would be that
1. `make_nd_axes` should retun `fig, axes` (consistent with plt.subplots(), with no real equivalent in pandas)
2. `samples.plot...` should return `axes` (consistent with `df.plot`)

My usual use-case (where I don't need fine-grained `make_nd_axes` control) is:
```python
axes = ['x0', 'x1', 'x2']
fig, axes = samples1.plot_2d(axes)
fig, axes = samples2.plot_2d(axes)
```
which would now become:
```python
axes = ['x0', 'x1', 'x2']
axes = samples1.plot_2d(axes)
axes = samples2.plot_2d(axes)
```
The reason for collecting the output is that it's then very easy to rearrange the order of plotting without a `make_2d_axes` command.

The disadvantage of dropping the `fig` output is that it's a bit of a faff to collect `fig = ax.iloc[0,0].figure`, but this would also be true of `pandas`.

[...]

PR #224, opened 2022-08-21
Change figure and axes output handling

### LINKED ISSUE #203: add `figsize` as kwarg to `make_1d_axes` and `make_2d_axes`
When setting up `fig, axes` with `make_1d_axes` or `make_2d_axes` I often want to specify the figure size. Currently that requires doing e.g. 
```python
fig = plt.figure(figsize=(fw, fh))
fig, axes = make_2d_axes(['x0', 'x1'], fig=fig)
```

It would be nice to be able to drop the first line and directly do
```python 
fig, axes = make_2d_axes(['x0', 'x1'], figsize=(fw, fh))
```

lukashergt:
This PR addresses issues #203, #223, #229 by
* creating a `gridspec_kw` kwarg for `make_nd_plot` as done in `plt.subplots`
* forwarding remaining kwargs to `plt.figure`
* matching pandas behaviour for `samples.plot_nd` in only returning the axes
* shifting the bulk of the axes building into `AxesDataFrame` and thereby making `make_nd_axes` slimmer
* fixing minor tick locations and turning off minor ticks by default (yes, that's an "and", not a "by")
* changing the default tick labelrotation to 45
* implementing a convenience `tick_params` method that broadcasts across the whole `AxesSeries` or `AxesDataFrame`
* add `[lower, diagonal, upper]` kwargs to `AxesDataFrame.axlines` and `AxesDataFrame.axspans` to allow limiting them to specific triangle positions
* changing `axlines` and `axspans` to take a dictionary for the params input, which means we are not as closely mirroring the `plt.axvline` signature, but it is more in line with what we are faced with here, where we have multiple axes belonging to different parameters
* implementing a `AxesDataFrame.scatter` method that allows plotting points (or stars or crosses etc.) as truth values on the off-diagonal subplots (i.e. similar to `axlines`, but where `axlines` is better at delimiting some hard lines such as `Omega_K=0`, `scatter` is better when you have a truth value for every parameter and you really just want to highlight that one n-dimensional point) 

Fixes #203, #223, #229 

# Checklist:


lukashergt:
I've had a first go at this (sorry, progress is slow atm, I'm in Oslo at a workshop), see comment above.
Thoughts?

anesthetic/plot.py
```diff
@@ -59,6 +102,135 @@ def _constructor(self):
...
+                                                 sharex=sx, sharey=sy)
+
+                    if position[x][y] == 0:
+                        self[x][y].twin = self[x][y].twinx()
+                        self[x][y].twin.set_yticks([])
+                        self[x][y].twin.set_ylim(0, 1.1)
+                        make_diagonal(self[x][y])
+                        self[x][y].position = 'diagonal'
+                        self[x][y].twin.xaxis.set_major_locator(
+                            MaxNLocator(3, prune='both'))
```
williamjameshandley, line 163: I think we've resigned ourselves to having a special paired axis on the diagonal, but this is clunky way of doing it -- is there  way to have a `DiagonalAxes` class which is defined in the body of this file, and created directly, rather than adjusted after the fact?

anesthetic/plot.py
```diff
@@ -59,6 +102,135 @@ def _constructor(self):
...
+                                a.twin.tick_params('x', direction='inout',
+                                                   bottom=True,
+                                                   labelbottom=False)
+                elif ticks == 'outer':  # no inner ticks
+                    for a in ax_[:-1]:
+                        a.tick_params('x', bottom=False, labelbottom=False)
+                elif ticks is None:  # no ticks at all
+                    for a in ax_:
+                        a.tick_params('x', bottom=False, top=False,
+                                      labelbottom=False, labeltop=False)
```
williamjameshandley, line 230: Is there any way to do this in fewer lines/a single loop?

williamjameshandley (changes requested):
Looking good -- we could use this opportunity to clean up some of this function. Happy to help out here if anything is unclear.

lukashergt:
Ouf, I've fiddled around a bunch, and finally managed to get it to a place where I think tings are much neater, AND tests are actually passing (was fighting with the slicing such that sliced `AxesDataFrames` return `AxesDataFrames` or `AxesSeries`...). I had already written a GitHub capitulation message, when it finally thankfully clicked...

Anyhow, all the initialisation stuff is now happening in the `__init__()` method of both `AxesSeries` and `AxesDataFrame`. There are a few `@staticmethods` that help with that.

Still some stuff for me to do, e.g. addressing #229, but it would be good to get some feedback at this point from @williamjameshandley on the initialisation stuff.

anesthetic/plot.py
lukashergt: Probably. However, I don't think it will reduce much in terms of number of lines, and on the downside I think it will become less readable. I think it's cleaner to handle vertical and horizontal axes in two separate loops. They look like duplicates, but there are different kwargs such as `left` and `right` vs `top` and `bottom`, and more importantly there are some subtleties when it comes to the handling of the diagonal axes...

lukashergt: It is in its own `_tick_params` method now, though.

anesthetic/plot.py
lukashergt: I tried using `matplotlib.axes._subplots.SubplotBase` as a parent class, but that did not work. The parent class of `DiagonalAxes` is not a specific matplotlib base class, but somehow is connected to the specific GridSpec instance...

I'll turn `make_diagonal` into a `@staticmethod`, though.

anesthetic/plot.py
```diff
@@ -59,6 +102,135 @@ def _constructor(self):
...
+        for j, y in enumerate(self.index):
+            for i, x in enumerate(self.columns):
+                if all_params.index(x) < all_params.index(y):
+                    if lower:
+                        self[x][y] = -1
+                elif all_params.index(x) > all_params.index(y):
+                    if upper:
+                        self[x][y] = +1
+                elif diagonal:
+                    self[x][y] = 0
```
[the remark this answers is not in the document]
lukashergt: Not sure, whether I've done what you had in mind, but that loop is its own `_position_frame` (static) method, now.

anesthetic/plot.py
```diff
@@ -47,10 +52,48 @@ def _constructor(self):
...
+
+        for p, g in zip(self.index, gs):
+            self[p] = ax = fig.add_subplot(g)
+            ax.set_xlabel(labels[p])
+            ax.set_yticks([])
+
+        for x, ax in self.dropna().iteritems():
+            ax.xaxis.set_major_locator(MaxNLocator(2, integer=True))
+
+        return fig, self
```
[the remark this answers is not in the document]
lukashergt: See [this comment](https://github.com/williamjameshandley/anesthetic/pull/224#issuecomment-1244224198) for my new approach.

lukashergt:
Ok, I think I've taken care of everything I wanted to do. Open to feedback.

(A bunch of things came up when introducing anesthetic to some of the BeyondPlanck people, who have mostly been using the corner package, but got annoyed with how difficult it was to implement some things. They were quite intrigued by anesthetic, but did struggle a bit with the lack of thorough documentation...)

anesthetic/plot.py
```diff
@@ -39,6 +40,16 @@
...
 
+    def __init__(self, data=None, index=None, fig=None, ncol=None, labels=None,
+                 gridspec_kw=None, subplot_spec=None, *args, **kwargs):
+        if data is None and index is not None:
+            data = pandas.Series(np.full(np.shape(index), None), index=index)
+            data = self.axes_series(data=data, fig=fig, ncol=ncol,
+                                    gridspec_kw=gridspec_kw,
+                                    subplot_spec=subplot_spec)
+            self._set_xlabels(axes=data, labels=labels)
+        super().__init__(data=data, index=index, *args, **kwargs)
```
williamjameshandley, line 51: This is actually more in keeping with the other pandas extensions we have (`WeightedDataFrame` etc), in comparison with the `@classmethod` I had in mind.

anesthetic/plot.py
```diff
@@ -39,6 +40,16 @@
 class AxesSeries(pandas.Series):
     """Anesthetic's axes version of `pandas.Series`."""
 
+    def __init__(self, data=None, index=None, fig=None, ncol=None, labels=None,
+                 gridspec_kw=None, subplot_spec=None, *args, **kwargs):
+        if data is None and index is not None:
+            data = pandas.Series(np.full(np.shape(index), None), index=index)
```
williamjameshandley, line 46: This could be replaced with `super().__init__(np.full(np.shape(index), None), index)`

anesthetic/plot.py
```diff
@@ -39,6 +40,16 @@
 class AxesSeries(pandas.Series):
```
williamjameshandley, line 40: Could we `from pandas import Series` at the top and drop `pandas.Series` (this would be more consistent with `WeightedDataFrame`)

anesthetic/plot.py
```diff
@@ -59,314 +145,428 @@ def _constructor(self):
...
+            for i, x in enumerate(columns):
+                if all_params.index(x) < all_params.index(y):
+                    if lower:
+                        position[x][y] = -1
+                elif all_params.index(x) > all_params.index(y):
+                    if upper:
+                        position[x][y] = +1
+                elif diagonal:
+                    position[x][y] = 0
+        return position
```
williamjameshandley, line 164: Is it OK if I have a look at turning this into a function, rather than an object, and push to this branch?

anesthetic/plot.py
```diff
@@ -59,314 +145,428 @@ def _constructor(self):
...
+            for x, ax in rows.iteritems():
+                if ax is not None and ax.position in positions:
+                    if x in params:
+                        for vmin, vmax in np.atleast_2d(params[x]):
+                            ax.axvspan(vmin, vmax, **kwargs)
+                    if y in params and ax.position != 'diagonal':
+                        for vmin, vmax in np.atleast_2d(params[y]):
+                            ax.axhspan(vmin, vmax, **kwargs)
+
+    def scatter(self, params, lower=True, upper=True, **kwargs):
```
williamjameshandley, line 427: Something which came up as part of BeyondPlanck? When is this useful?

lukashergt: Yes. Vertical and horizontal lines (which is already implemented with `axlines`) are nice when there are some theoretical values that separate parameter space (e.g. `Omega_K=0` separates open and closed universes). However, when comparing contours to some truth value then it is actually just one n-dimensional point. Emphasis on point, i.e. this doesn't separate parameter space into different regions. So in those situations it would be nicer to highlight the truth with just a point/star/cross/...

anesthetic/plot.py
lukashergt: Done.

anesthetic/plot.py
```diff
@@ -47,10 +58,85 @@ def _constructor(self):
...
+        if subplot_spec is None:
+            gs = GridSpec(nrow, ncol, wspace=wspace, **gridspec_kw)
+        else:
+            gs = GridSpecFromSubplotSpec(nrow, ncol, wspace=wspace,
+                                         subplot_spec=subplot_spec,
+                                         **gridspec_kw)
+        for p, g in zip(axes.index, gs):
+            axes[p] = ax = fig.add_subplot(g)
+            ax.set_yticks([])
+        return axes
```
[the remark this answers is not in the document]
lukashergt: Into `AxesSeries._constructor`?  `axes_series` takes a bunch of args and kwargs which doesn't fit in a property such as `_constructor`, I think.

Or do you mean `AxesSeries.__init__`? It could go in that, I guess, but I tried not letting the `__init__` get too bulky...

anesthetic/plot.py
```diff
@@ -47,10 +58,85 @@ def _constructor(self):
...
+
+    @staticmethod
+    def _set_xlabels(axes, labels, **kwargs):
+        if labels is None:
+            labels = {}
+        labels = {p: labels[p] if p in labels else p for p in axes.index}
+        for p, ax in axes.iteritems():
+            ax.set_xlabel(labels[p], **kwargs)
```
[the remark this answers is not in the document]
lukashergt: I want the labels to be set already upon initialisation. For that I need a staticmethod, because there cannot be any slicing happening in the `__init__`, meaning any sort of `self[p]` or `self.iloc[0]` or `self.iteritems()` cannot be put into `__init__`. Otherwise this messes with `_constructor` and `_constructor_expanddim` and `_constructor_sliced`.

anesthetic/plot.py
williamjameshandley: Agreed.

williamjameshandley (approved):

PR #230, opened 2022-09-11
Fixes for dropping weights and labels

williamjameshandley:
# Description

In usage I have found a couple of bugs in the way that anesthetic currently drops weights and labels, namely that
- drop_labels returns a view rather than a copy in the case that no labels are present to be dropped (which is inconsistent with the expected behaviour)
- drop_weights returns nothing in the case that no weights are present.

This PR creates some failing tests, and then fixes them

# Checklist:


tests/test_weighted_pandas.py
```diff
@@ -926,3 +926,11 @@ def test_set_weights(mcmc_wdf):
     mcmc_wdf.set_weights(None, inplace=True)
     assert id(mcmc_wdf) == mcmc_id
     assert not mcmc_wdf.isweighted()
+
+
+def test_drop_weights(mcmc_wdf):
+    assert mcmc_wdf.isweighted()
+    noweights = mcmc_wdf.drop_weights()
+    assert not noweights.isweighted()
+    assert_array_equal(noweights.drop_weights(), mcmc_wdf)
```
lukashergt, line 935: Why this line? Shouldn't that be false?

Similar to the testing of the labelled dataframes, wouldn't it be better to work with both `mcmc_df` and `mcmc_wdf`?

tests/test_weighted_pandas.py
```diff
@@ -926,3 +926,11 @@ def test_set_weights(mcmc_wdf):
...
+    assert noweights.drop_weights is not noweights
```
lukashergt, line 936: `drop_weights` needs parentheses afterwards, right? Missing in this line.

tests/test_labelled_pandas.py
```diff
@@ -450,3 +450,12 @@ def test_reset_index():
...
     new.reset_index(inplace=True)
     assert_array_equal(new.index, ldf.reset_index().index)
+
+
+def test_drop_labels():
+    ldf = test_LabelledDataFrame_index()
+    assert ldf.islabelled()
+    nolabels = ldf.drop_labels()
+    assert not nolabels.islabelled()
+    assert_array_equal(nolabels.drop_labels(), nolabels)
```
lukashergt, line 460: Why this line? Isn't that the same by construction?

For addressing the problem in this PR's description, wouldn't it be better to have test_LabelledDataFrame_index() return both lframe and frame, and then do testing with both?

tests/test_weighted_pandas.py
williamjameshandley: Adjusted to now check with assert_frame_equal, and to check that assert_array_equal follows for noweights and mcmc_wdf

tests/test_weighted_pandas.py
williamjameshandley: Nice catch.

tests/test_labelled_pandas.py
williamjameshandley: It is, but before `nolabels.drop_labels()` returned it could return `None`. Update to `assert_frame_equal` which also tests indices.

lukashergt (approved):
Looks good now. Please squash and merge :)

PR #231, opened 2022-09-11
Fix for filenames which contain regular expressions

williamjameshandley:
# Description

Filenames which contain regex characters (like '+') currently cannot be read. 

PR introduces a failing test, which is fixed by changing one line to include an `re.escape(...)`. I have also removed the .ranges files in the examples, since these are no longer used as of #212.

# Checklist:


lukashergt (approved):
Looks good, please squash and merge.

PR #232, opened 2022-09-16
pass labels argument through when reading chains

Ormorod:
I noticed that `labels` isn't passed along in the same way as `columns` when reading chains.

### DIFF
--- a/anesthetic/read/polychord.py
+++ b/anesthetic/read/polychord.py
@@ -24,7 +24,7 @@ def read_polychord(root, *args, **kwargs):
     columns, labels = read_paramnames(root)
 
     columns = kwargs.pop('columns', columns)
-    labels = kwargs.pop("labels", labels)
+    labels = kwargs.pop('labels', labels)
     kwargs['label'] = kwargs.get('label', os.path.basename(root))
 
     return NestedSamples(data=data, columns=columns,
--- a/anesthetic/read/polychord.py
+++ b/anesthetic/read/polychord.py
@@ -21,9 +21,9 @@ def read_polychord(root, *args, **kwargs):
     except IOError:
         pass
     data, logL, logL_birth = np.split(data, [-2, -1], axis=1)
-    params, labels = read_paramnames(root)
+    columns, labels = read_paramnames(root)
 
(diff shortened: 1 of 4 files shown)


lukashergt (changes requested):
Hi @Ormorod,

thanks so much for stress testing and contributing to anesthetic! 

* Could we go for a consistent use of `'.'` rather than `"."` for the dictionary keys? This probably isn't perfectly consistent throughout anesthetic, but it's good to clean these things up when they come up...
* Could you do the changes also for `anesthetic/read/getdist.py`, please?

lukashergt:
> I can, guess I should get used to it (I've always preferred `"`, as I find myself much more often needing an apostrophe than quotation marks in a string)

There is no rule about this, PEP8 takes a mostly neutral stance, except saying to avoid escaping with backslashes. However, I think there is somewhat of a community consensus (take e.g. the votes on [this stackexchange answer](https://stackoverflow.com/a/56190/6135182)) which I try to follow where
* double quotes `"` are used for text, e.g. in print statements, warnings or error messages (which accommodates the use of apostrophes...),
* single quotes `'` are used for single characters, dictionary keywords or other identifiers such as `bins='auto'` in histograms.

So that is what I'd suggest to go with. While I am open to other suggestions, I am strongly in favour of some form of consistency, so seeing `'columns'` and `"labels"` in successive rows seems wrong to me...

anesthetic/read/polychord.py
```diff
@@ -21,9 +21,10 @@ def read_polychord(root, *args, **kwargs):
     except IOError:
         pass
     data, logL, logL_birth = np.split(data, [-2, -1], axis=1)
-    params, labels = read_paramnames(root)
+    columns, labels = read_paramnames(root)
 
-    columns = kwargs.pop('columns', params)
+    columns = kwargs.pop('columns', columns)
+    labels = kwargs.pop("labels", labels)
```
lukashergt, line 27: One more set of double quotes.

lukashergt (approved):
Thanks for this @Ormorod, please feel free to squash and merge.

PR #234, opened 2022-09-27
Fix warnings shown with pytest

Ormorod:
Worked my way through the warnings shown by pytest. 

Fixes:
- in `plot.hist_plot_1d()`, check that bins is a str before comparing to `['knuth', ...]`, as [str <-> numpy comparison is broken](https://stackoverflow.com/questions/40659212/futurewarning-elementwise-comparison-failed-returning-scalar-but-in-the-futur)
- in `utils.insertion_p_value()`, replace `indexes` with `np.array(indexes)`, to avoid the need to slice using `.iloc` if `indexes` is a `DataFrame`
- Add missing `method` parameter to `WeightedDataFrame` to match the same method of `DataFrame`

Pandas warnings:
- add `level=0` to `samples.drop('chain')` when reading Cobaya and GetDist chains. This could also be dealt with by sorting before dropping, something to consider?
- use `.loc[boolean_mask]` where appropriate

Pandas deprecations:
- replace iteritems() with items()
- replace `inplace=True` with `copy=False` for `DataFrame.set_axis()`

Changes to tests:
- `test_gui.test_gui()`: Skip references to button.observers if matplotlib>=3.6.0, as this no longer exists TODO: discuss what new test should be
- `test_labelled_pandas.py`: All the tests using the ABCDEFGHIK and vowel/consonant example cause Pandas to complain that the MultiIndex should be lexicographically sorted for performance (i.e. consonant before vowel). I've suppressed these with a warning filter, alternatively the index could be reordered BCDFGHJAEI.
- `test_samples.test_beta_with_logL_infinities()` sets the first ten elements of logL to `-np.inf`, which causes logL <= logL_birth and therefore the corresponding warning, which I have suppressed.
- More cases of `.loc[boolean_mask]`
- More cases of adding `level=0` to `samples.drop()`

### DIFF
--- a/anesthetic/read/cobaya.py
+++ b/anesthetic/read/cobaya.py
@@ -82,7 +82,7 @@ def read_cobaya(root, *args, **kwargs):
     samples.label = kwargs['label']
 
     if (samples.chain == samples.chain.iloc[0]).all():
-        samples.drop('chain', inplace=True, axis=1, level=0)
+        samples.drop(columns='chain', inplace=True, level=0)
     else:
         samples.set_label('chain', r'$n_\mathrm{chain}$')
 
--- a/anesthetic/read/cobaya.py
+++ b/anesthetic/read/cobaya.py
@@ -82,7 +82,7 @@ def read_cobaya(root, *args, **kwargs):
     samples.label = kwargs['label']
 
     if (samples.chain == samples.chain.iloc[0]).all():
-        samples.drop('chain', inplace=True, axis=1)
+        samples.drop('chain', inplace=True, axis=1, level=0)
     else:
         samples.set_label('chain', r'$n_\mathrm{chain}$')
 
--- a/anesthetic/read/getdist.py
+++ b/anesthetic/read/getdist.py
@@ -92,7 +92,7 @@ def read_getdist(root, *args, **kwargs):
 
     all_same_chain = (samples.chain == samples.chain.iloc[0]).all()
     if all_same_chain or samples.chain.isna().all():
-        samples.drop('chain', inplace=True, axis=1)
+        samples.drop('chain', inplace=True, axis=1, level=0)
     elif samples.islabelled():
         samples.set_label('chain', r'$n_\mathrm{chain}$')
 
--- a/anesthetic/read/getdist.py
+++ b/anesthetic/read/getdist.py
@@ -92,7 +92,7 @@ def read_getdist(root, *args, **kwargs):
 
     all_same_chain = (samples.chain == samples.chain.iloc[0]).all()
     if all_same_chain or samples.chain.isna().all():
-        samples.drop('chain', inplace=True, axis=1, level=0)
+        samples.drop(columns='chain', inplace=True, level=0)
     elif samples.islabelled():
         samples.set_label('chain', r'$n_\mathrm{chain}$')
 
--- a/anesthetic/samples.py
+++ b/anesthetic/samples.py
@@ -919,7 +919,8 @@ class NestedSamples(Samples):
             Importance re-weighted samples.
         """
         samples = super().importance_sample(logL_new, action=action)
-        samples = samples[(samples.logL > samples.logL_birth).to_numpy()].recompute()
+        mask = (samples.logL > samples.logL_birth).to_numpy()
+        samples = samples[mask].recompute()
         if inplace:
             self._update_inplace(samples)
         else:
--- a/anesthetic/samples.py
+++ b/anesthetic/samples.py
@@ -200,7 +200,7 @@ class Samples(WeightedLabelledDataFrame):
         kwargs['kind'] = kwargs.get('kind', 'kde_1d')
         kwargs['label'] = kwargs.get('label', self.label)
 
-        for x, ax in axes.iteritems():
+        for x, ax in axes.items():
             if x in self and kwargs['kind'] is not None:
                 xlabel = self.get_label(x)
                 self[x].plot(ax=ax, xlabel=xlabel,
@@ -304,7 +304,7 @@ class Samples(WeightedLabelledDataFrame):
             fig = axes.bfill().to_numpy().flatten()[0].figure
 
         for y, row in axes.iterrows():
-            for x, ax in row.iteritems():
+            for x, ax in row.items():
                 if ax is not None:
                     pos = ax.position
                     lkwargs = local_kwargs.get(pos, {})
--- a/anesthetic/samples.py
+++ b/anesthetic/samples.py
@@ -877,8 +877,8 @@ class NestedSamples(Samples):
                 logL = float(self.logL[logL])
             except KeyError:
                 pass
-        i = (self.logL >= logL) & (self.logL_birth < logL)
-        return Samples(self.loc[i]).set_weights(None)
+        i = ((self.logL >= logL) & (self.logL_birth < logL)).to_numpy()
+        return Samples(self[i]).set_weights(None)
 
     def posterior_points(self, beta=1):
         """Get equally weighted posterior points at temperature beta."""
@@ -919,7 +919,7 @@ class NestedSamples(Samples):
             Importance re-weighted samples.
         """
         samples = super().importance_sample(logL_new, action=action)
-        samples = samples.loc[samples.logL > samples.logL_birth].recompute()
+        samples = samples[(samples.logL > samples.logL_birth).to_numpy()].recompute()
         if inplace:
             self._update_inplace(samples)
         else:
@@ -964,7 +964,7 @@ class NestedSamples(Samples):
                 raise RuntimeError("Cannot recompute run without "
                                    "birth contours logL_birth.")
 
-            invalid = samples.logL <= samples.logL_birth
+            invalid = (samples.logL <= samples.logL_birth).to_numpy()
             n_bad = invalid.sum()
             n_equal = (samples.logL == samples.logL_birth).sum()
             if n_bad:
@@ -993,7 +993,7 @@ class NestedSamples(Samples):
                           " should investigate why your likelihood is throwing"
                           " NaNs. Dropping these samples at prior level",
                           RuntimeWarning)
-            samples = samples.loc[samples.logL.notna()].recompute()
+            samples = samples.loc[samples.logL.notna().to_numpy()].recompute()
 
         if inplace:
             self._update_inplace(samples)
--- a/anesthetic/samples.py
+++ b/anesthetic/samples.py
@@ -878,7 +878,7 @@ class NestedSamples(Samples):
             except KeyError:
                 pass
         i = (self.logL >= logL) & (self.logL_birth < logL)
-        return Samples(self[i]).set_weights(None)
+        return Samples(self.loc[i]).set_weights(None)
 
     def posterior_points(self, beta=1):
         """Get equally weighted posterior points at temperature beta."""
@@ -919,7 +919,7 @@ class NestedSamples(Samples):
             Importance re-weighted samples.
         """
         samples = super().importance_sample(logL_new, action=action)
-        samples = samples[samples.logL > samples.logL_birth].recompute()
+        samples = samples.loc[samples.logL > samples.logL_birth].recompute()
         if inplace:
             self._update_inplace(samples)
         else:
@@ -977,7 +977,7 @@ class NestedSamples(Samples):
                               "\nDropping the invalid samples." %
                               (n_bad, len(samples), n_equal),
                               RuntimeWarning)
-                samples = samples[~invalid].reset_index(drop=True)
+                samples = samples.loc[~invalid].reset_index(drop=True)
 
             samples.sort_values('logL', inplace=True)
             samples.reset_index(drop=True, inplace=True)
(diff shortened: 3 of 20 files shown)


Ormorod:
Personal preference, but I much prefer black as a formatting workflow...

tests/test_samples.py
```diff
@@ -828,8 +830,9 @@ def test_beta():
 def test_beta_with_logL_infinities():
     ns = read_chains("./tests/example_data/pc")
     ns.loc[:10, ('logL', r'$\ln\mathcal{L}$')] = -np.inf
-
-    ns.recompute(inplace=True)
+    with warnings.catch_warnings(record=True) as w:
+        ns.recompute(inplace=True)
+        assert "samples have logL <= logL_birth" in str(w[0].message)
```
lukashergt, line 835: For consistency, I'd go with pytest's own handling of catching warnings:
```python
with pytest.warns(UserWarning):
    ns.recompute(inplace=True)
```

tests/test_samples.py
```diff
@@ -713,12 +713,14 @@ def test_masking():
 
     for kind in kinds:
         fig, axes = make_1d_axes(['x0', 'x1', 'x2'])
-        pc[mask].plot_1d(axes=axes, kind=kind + '_1d')
+        pc.loc[mask].plot_1d(axes=axes, kind=kind + '_1d')
```
lukashergt, line 716: This is another case, where I'd prefer changing the mask to an array. When I see `.loc` I expect one of the labels for indexing...
```python
mask = pc['x0'].to_numpy() > 0
```

tests/test_samples.py
```diff
@@ -1067,7 +1070,7 @@ def test_logzero_mask_likelihood_level():
 
     ns1 = read_chains('./tests/example_data/pc')
     ns1.logL = np.where(mask, ns1.logL, -1e30)
-    ns1 = merge_nested_samples((ns1[ns1.logL > ns1.logL_birth],))
+    ns1 = merge_nested_samples((ns1.loc[ns1.logL > ns1.logL_birth],))
```
lukashergt, line 1073:
```python
mask = ns1.logL.to_numpy() > ns1.logL_birth.to_numpy()
ns1 = merge_nested_samples((ns1[mask],))
```

tests/test_samples.py
```diff
@@ -713,12 +713,14 @@ def test_masking():
...
+        pc.loc[mask].plot_1d(axes=axes, kind=kind + '_1d')
 
     for kind in kinds + ['scatter']:
         fig, axes = make_2d_axes(['x0', 'x1', 'x2'], upper=False)
-        pc[mask].plot_2d(axes=axes, kind=dict(lower=kind + '_2d',
-                                              diagonal='hist_1d'))
+        pc.loc[mask].plot_2d(
+            axes=axes,
+            kind=dict(lower=kind + '_2d', diagonal='hist_1d'),
+        )
```
lukashergt, line 723: With the mask changed into an array (see above), this can then be reverted.

anesthetic/read/cobaya.py
```diff
@@ -82,7 +82,7 @@ def read_cobaya(root, *args, **kwargs):
     samples.label = kwargs['label']
 
     if (samples.chain == samples.chain.iloc[0]).all():
-        samples.drop('chain', inplace=True, axis=1)
+        samples.drop('chain', inplace=True, axis=1, level=0)
```
lukashergt, line 85: I think this would also be more explicit and hence nicer as
```python
samples.drop(columns='chain', inplace=True, level=0)
```

anesthetic/read/getdist.py
```diff
@@ -92,7 +92,7 @@ def read_getdist(root, *args, **kwargs):
 
     all_same_chain = (samples.chain == samples.chain.iloc[0]).all()
     if all_same_chain or samples.chain.isna().all():
-        samples.drop('chain', inplace=True, axis=1)
+        samples.drop('chain', inplace=True, axis=1, level=0)
```
lukashergt, line 95: Again
```python
samples.drop(columns='chain', inplace=True, level=0)
```

tests/test_labelled_pandas.py
```diff
@@ -34,292 +35,329 @@ def assert_frame_equal_not_index(x, y):
...
 
 def test_LabelledSeries():
-    index = ['A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J']
-    labels = ['$%s$' % i for i in index]
-    data = np.random.rand(len(index))
+    with warnings.catch_warnings():
+        warnings.filterwarnings(
+            "ignore",
+            message=".*indexing past lexsort depth may impact performance.*",
+        )
```
lukashergt, line 42: I think a function decorator should work, here, and then we don't need the extra indent (will also make it much easier to review here on GitHub...).

```python
@pytest.mark.filterwarnings("ignore::PerformanceWarning")
def test_LabelledSeries():
    ...
```

Same for the other tests below.

lukashergt (changes requested):
Wow, thanks for looking into these warnings!

* If we are changing `inplace` to `copy`, then we need to change the `requirements.txt` to have `pandas>=1.5.0`.
* other comments inline

tests/test_samples.py
Ormorod: `mask` is already used in this test, so I'll give it a different name

anesthetic/weighted_pandas.py
```diff
@@ -76,7 +76,7 @@ def set_weights(self, weights, axis=0, inplace=False, level=None):
             names.insert(level, 'weights')
 
             index = MultiIndex.from_arrays(index, names=names)
-            result.set_axis(index, axis=axis, inplace=True)
+            result = result.set_axis(index, axis=axis, copy=False)
```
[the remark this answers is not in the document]
Ormorod: I've read in various places that inplace is going to disappear with pandas 2

tests/test_labelled_pandas.py
Ormorod: I get the following error:
```
ERROR: while parsing the following warning configuration:

  ignore::PerformanceWarning

This error occurred:

Traceback (most recent call last):
  File "/Users/adam/Programming/env_oct03/lib/python3.10/site-packages/_pytest/config/__init__.py", line 1639, in parse_warning_filter
    category: Type[Warning] = _resolve_warning_category(category_)
  File "/Users/adam/Programming/env_oct03/lib/python3.10/site-packages/_pytest/config/__init__.py", line 1678, in _resolve_warning_category
    cat = getattr(m, klass)
AttributeError: module 'builtins' has no attribute 'PerformanceWarning'
```

Any ideas how to solve?

tests/test_samples.py
Ormorod: Done (actually missed this one, github thought it would be helpful to hide three of these conversations)

btw is a `RuntimeWarning`, not `UserWarning`

anesthetic/weighted_pandas.py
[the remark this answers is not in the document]
lukashergt: Yes, sorry, I should have been clearer. This wasn't a request to change anything, mostly a comment to say thanks for catching this, and as a memo to my/our future selves to keep in mind that a simple change from `inplace=True` to `copy=False` is not enough. I think there is a bunch of `inplace` stuff throughout anesthetic, which might need updating at some point, when the `inplace` kwarg gets dropped.

tests/test_labelled_pandas.py
lukashergt: This is a pandas warning, so it might need importing at the head of the file.
```python
from pandas.errors import PerformanceWarning
```

anesthetic/weighted_pandas.py
```diff
@@ -350,14 +350,42 @@ def sem(self, axis=0, skipna=True):
...
+              fractional part of the index surrounded by `i` and `j`.
+            * lower: `i`.
+            * higher: `j`.
+            * nearest: `i` or `j` whichever is nearest.
+            * midpoint: (`i` + `j`) / 2.
+        method : {'single', 'table'}, default 'single'
+            Whether to compute quantiles per-column ('single') or over all
+            columns ('table'). When 'table', the only allowed interpolation
+            methods are 'nearest', 'lower', and 'higher'.
+        """
```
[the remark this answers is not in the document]
Ormorod: Looking back I'm honestly not sure why I felt the need to copy and paste ~an essay~ the entire docstring here, and if we wanted to add `(more info at...)` here then it should probably also feature in a lot of places

If we left the docstrings blank, then I think python >=3.5 will use inheritance to find the docstring?

tests/test_labelled_pandas.py
lukashergt: Try:
```python
@pytest.mark.filterwarnings("ignore::pandas.errors.PerformanceWarning")
```

tests/test_samples.py
lukashergt: Oh, you're right, since we raise it as a `RuntimeWarning` we should definitely also test for a `RuntimeWarning` and not for a `UserWarning`. 

As to which one would be more appropriate here... never 100% sure about this. I might actually be in favour of having an `AnestheticWarning` for anesthetic specific warnings, but that is definitely beyond this PR.

tests/test_labelled_pandas.py
Ormorod: perfect

anesthetic/weighted_pandas.py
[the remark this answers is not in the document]
lukashergt: > If we left the docstrings blank, then I think python >=3.5 will use inheritance to find the docstring?

Really?! That might be nice... potentially dangerous, but potentially nice... Have you tested that?

anesthetic/samples.py
```diff
@@ -977,7 +978,7 @@ def recompute(self, logL_birth=None, inplace=False):
                               "\nDropping the invalid samples." %
                               (n_bad, len(samples), n_equal),
                               RuntimeWarning)
-                samples = samples[~invalid].reset_index(drop=True)
+                samples = samples.loc[~invalid].reset_index(drop=True)
```
lukashergt, line 981: `.loc` can go...

anesthetic/samples.py
```diff
@@ -993,7 +994,7 @@ def recompute(self, logL_birth=None, inplace=False):
                           " should investigate why your likelihood is throwing"
                           " NaNs. Dropping these samples at prior level",
                           RuntimeWarning)
-            samples = samples[samples.logL.notna()].recompute()
+            samples = samples.loc[samples.logL.notna().to_numpy()].recompute()
```
lukashergt, line 997: `.loc` can go...

anesthetic/weighted_pandas.py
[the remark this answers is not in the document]
Ormorod: Yes, if I delete the docstring from `WeightedSeries.quantile`, then e.g. `help(ns.quantile)` or hovering over quantile in vscode shows the docstring from `DataFrame` (forgive the screenshot)

Screenshot 2022-10-13 at 18 35 30

tests/test_samples.py
Ormorod: mask2 is the limit of my creativity for now

anesthetic/weighted_pandas.py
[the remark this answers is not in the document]
lukashergt: Oooh, this could be neat. Had no idea that this was a thing. 

I don't think we should use this PR to remove docstrings all over the place, though, so I'd suggest reverting to the previous docstring for now.

Could you make an issue that this might be worth looking more into (screenshots appreciated)?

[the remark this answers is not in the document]
lukashergt: Does this risk a clash with `pydocstyle`?

tests/test_samples.py
lukashergt: I think overwriting `mask` would be fine, wouldn't it?

Ormorod: as you wish

anesthetic/weighted_pandas.py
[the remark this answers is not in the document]
Ormorod: I think deleting a load of docstrings would deserve its own PR

anesthetic/weighted_pandas.py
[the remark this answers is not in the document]
Ormorod: Yes there is, I actually whizzed through a load of them before realising `set_axis` is either one of, or the only method to get the new `copy`, and put them all back again.

anesthetic/samples.py
Ormorod: I think I've put back/`to_numpy()`ed all the ones I changed.

anesthetic/weighted_pandas.py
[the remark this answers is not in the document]
lukashergt: Ok, good to know, so maybe we are safe.

anesthetic/weighted_pandas.py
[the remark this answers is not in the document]
Ormorod: > Oooh, this could be neat. Had no idea that this was a thing.
> 
> I don't think we should use this PR to remove docstrings all over the place, though, so I'd suggest reverting to the previous docstring for now.
> 
> Could you make an issue that this might be worth looking more into (screenshots appreciated)?

Sorry, didn't see this reply. GH not playing ball with me this evening.

tests/test_labelled_pandas.py
lukashergt: Ah, so much nicer to review now, too :)

Ormorod:
First pass at changing `test_labelled_pandas.py` to use `pytest.fixtures`.

For now, I've just commented out the return statements which are never used.

lukashergt (approved):
Great, this nicely cleans up our test output. Thanks @Ormorod, please squash and merge.

Issue #233, opened 2022-09-25
Pandas 2 `inplace` deprecation


AdamOrmondroyd:
The Pandas team intend to depreciate the inplace parameter ready for Pandas 2, see the issue [16529](https://github.com/pandas-dev/pandas/issues/16529).

With respect to anesthetic, [Pandas 1.5.0](https://pandas.pydata.org/docs/whatsnew/v1.5.0.html) deprecated `inplace` from `DataFrame.set_axis()` and replaced it with `copy`:

```
df.set_axis(..., inplace=True)
```
should be replaced with
```
df = df.set_axis(..., copy=False)
```
There are others (e.g. `Series.set_axis()` and several methods of `Categorical`), and I would expect more to follow.

`DataFrame.iteritems()` has also been deprecated in favour of `DataFrame.items()`

Ormorod:
- [ ] I've successfully merged my changes to how pandas `HistPlot` handles weights, which I have copied into #235. This is a reminder to remove this once we move to pandas 2.

Ormorod:
Looks like [about two weeks](https://github.com/pandas-dev/pandas/releases/tag/v2.0.0rc0) until Pandas 2 is here...

williamjameshandley:
Resolved by #292

lukashergt:
Can this issue be closed?

AdamOrmondroyd:
I suppose so, the "more to follow" never arrived…

PR #235, opened 2022-09-28
Replace overridden `_make_plot()` with `_plot()`

Ormorod:
Since this morning, I've updated the other 1d plots to override `_plot()` instead of `_make_plot()`. For any readers, this gives us a lot of the plot argument handling, such as colo[u]?rs, for free, as it is inherited from Pandas' `HistPlot`

Given there will need to be changes for pandas 2, I am going to change this version to require pandas<2.0.0


Ormorod:
I went ahead and changed `hist_plot_1d()` to return all three `plt.hist()` outputs, as I feel it should given it's intended as a wrapper. Other than in hist.py, `hist_plot_1d()` is only used in tests, so I just updated `hist_plot_1d()` -> `hist_plot_1d()[2]`

Ormorod:
I expect tests.test_samples.test_contour_plot_2d_nan() to fail, due to a bug in Pandas. If the data contain NaNs, these are discarded in Pandas' HistPlot._make_plot(), however the corresponding weights are *not* discarded, so the data and weights have different lengths and the plot fails. However, both locally and on CSD3 this test is passed, though if I copy the test it fails.

Any ideas why this might be?

Ormorod:
`_make_plot()` is far from finished, but this setup works for at least the all 1D case.
I haven't fully worked out how to support weights if data is 2D, as if there is a single set of weights, different elements would need to be dropped for each y.
I also haven't worked out why each y in data would ever be 2D, and what should be done if that is the case.

Ormorod:
after a lot of head scratching, I've replaced `HistPlot_make_plot()` with that of my PR to Pandas.

The crux is that dropping weights corresponding to NaNs in the data and passing this to `super._make_plot()` doesn't work as if y and/or weights are 2D, weights would become non-square in general, and so can't be stored as an array. I had an attempt which used masking, but I think this may be overcomplicating things.

lukashergt:
Currently this PR is not set up to merge into our master branch, was that intentional? As a consequence CI is not working as supposed to, if it was not intentional, can we move this to a new PR that is based off of our master branch?

Ormorod:
I think I can just change this PR?

Ormorod:
I'm not happy with the way I've dealt with `weights` and `bw_method` for `FastKde1dPlot`, but I think dealing with it properly would require a big shake-up of the inheritance tree...

tests/test_samples.py
```diff
@@ -1179,7 +1179,7 @@ def test_samples_dot_plot():
         axes = samples.plot.fastkde_2d('x0', 'x1')
         assert len(axes.collections) == 5
         plt.close("all")
-        axes = samples.plot.fastkde_1d()
+        axes = samples.x0.plot.fastkde_1d()
```
Ormorod, line 1182: On the master branch, the original line plots the fastkde of `x0`; I don't see why this should happen. On this branch, the original line plots the fastkde for all columns, including logL, logL_birth and nlive. I think the most appropriate behaviour would be to plot `x0`-`x4`?

For the moment, as the other parts of `test_samples_dot_plot()` only do 1d plots for a single column, I have altered this line to match.

Ormorod:
My [pandas PR 48888](https://github.com/pandas-dev/pandas/pull/48888) has been merged to main, and as I understand it, will be included in pandas 2. I'll add a reminder to remove these lines to #233 when we inevitably have to change `inplace=True` to 'copy=False`...

Ormorod:
I had a go at this a while ago, and I couldn't reproduce it (I'm on macos).

Looking at `pandas.plotting._matplotlib.core.py`, I think the post plot logic only sets one of the y limits of an area plot to 0 if the axes aren't shared. I'll copy it across to `anesthetic` and always set a limit to 0 to see if it makes any difference

Ormorod:
@lukashergt I notice that the conda windows install includes pyqt, which is an optional matplotlib backend. Could this be affecting it?

Ormorod:
damn, I thought I had it

lukashergt:
> **Edit 2:** @lukashergt at what sort of rate were you seeing failures? the version above this comment has failed conda windows 3.10 2/2 times, below has succeeded 10/10.

To get the most persistent one to work (I don't remember which one exactly, but it was one of the windows ones) I had to restart it at least 5 times...

lukashergt:
> @lukashergt I notice that the conda windows install includes pyqt, which is an optional matplotlib backend. Could this be affecting it?

Possibly, although from #244 it seems like the culprit might be the boxplot...

tests/test_samples.py
williamjameshandley: I agree with testing it this way. Plotting all columns is the correct (albeit useless in this case) behaviour. The correct way to use this would be `samples[['x0','x1','x2','x3']].plot.fastkde_1d()`, which would only make sense for a nested sampling run if the parameters had the same dimensions (e.g. for locations of a galaxy, or angles in a protein, but not for cosmological parameters).

anesthetic/plotting/_matplotlib/hist.py
```diff
@@ -1,23 +1,31 @@
...
 )
-from pandas.plotting._matplotlib.core import MPLPlot, PlanePlot
+from pandas.io.formats.printing import pprint_thing
+from pandas.plotting._matplotlib.core import MPLPlot
+from pandas.plotting._matplotlib.groupby import (create_iter_data_given_by,
+                                                 reformat_hist_y_given_by)
+try:
+    from typing import Literal
+except ImportError:
+    from typing_extensions import Literal
```
williamjameshandley, line 18: This is no longer necessary as of #238 requiring Python 3.8

anesthetic/plotting/_matplotlib/hist.py
```diff
@@ -44,6 +52,77 @@ def _calculate_bins(self, data):
         )
         return bins
 
+    def _make_plot(self):
+        if version.parse(pd.__version__) >= version.parse("2.0.0"):
+            super._make_plot()
+            return
```
williamjameshandley, line 58: This can be ignored for coverage purposes by putting in: at each relevant clause
`def _make_plot(self):  # pragma: no cover`

since it's effectively covered by the pandas test suite.

anesthetic/plotting/_matplotlib/core.py
```diff
@@ -59,19 +79,40 @@ def __init__(self, data, x, y, s=None, c=None, **kwargs) -> None:
...
+
+        kwds = self.kwds.copy()
+        label = pprint_thing(self.label)
+        kwds["label"] = label
+
+        style, kwds = self._apply_style_colors(colors, kwds, 0, label)
+        if style is not None:
+            kwds["style"] = style
+
+        self._plot(ax, x, y, **kwds)
```
williamjameshandley, line 104: It looks like this code has been copied from `HistPlot`. Is there a reason (a) `HistPlot` is the right template to use (rather than `PlanePlot` or `LinePlot` and (b) why we can't just (diamond) inherit this functionality?

anesthetic/plotting/_matplotlib/hist.py
```diff
@@ -69,93 +148,122 @@ def _plot(
...
+        if not isinstance(bins, str):
+            base = np.zeros(len(bins) - 1)
+            bottom = bottom + cls._get_stacked_values(ax, stacking_id,
+                                                      base, kwds["label"])
+
+        # ignore style
+        n, bins, patches = hist_plot_1d(ax, y, bins=bins,
+                                        bottom=bottom, **kwds)
+        cls._update_stacker(ax, stacking_id, n)
+        return patches
```
williamjameshandley, line 233: What does this implement which `HistPlot._plot` does not?

anesthetic/plotting/_matplotlib/core.py
```diff
@@ -27,7 +31,23 @@ def __init__(self, data, *args, **kwargs):
...
+                name = self.data.drop_weights().index.name
+                if name is not None:
+                    name = pprint_thing(name)
+
+            # GH 45145, override the default axis label if one is provided.
+            index_name = self._get_custom_index_name()
+            if index_name is not None:
+                name = pprint_thing(index_name)
+
+            return name
```
williamjameshandley, line 50: This should be covered by a test for the use-case by which you discovered you needed it.

williamjameshandley (changes requested):
This is looking good -- the main thing we need to convince ourselves of is whether the extra functions are needed, or whether they can be diamond inherited (the latter is preferable as it will remain up-to-date with pandas, particularly when it comes to signatures). We also need to make sure all the code is covered. (this is highlighted in the "Files changed" tab.

Ormorod:
I remembered that these changes also allows the user to change the colormap.

e.g. `ns.plot.kde_2d('x0', 'x1', cmap='viridis')`:
before:
image
after:
image

lukashergt:
> @lukashergt I think you've added a test for this in #243 ?

I've added a test with a `cmap` kwarg to `test_samples.py`, yes. Note the `samples`. There were already plenty of colour tests in `test_plot.py`, which all work fine, the problem was piping the correct kwargs through to the plotting functions, where pandas uses it's own aliases and silently ignores some kwargs or errors on others, e.g. `cmap` vs `colormap`, or `color` vs `c`.

Ormorod:
One more thing I've messed up is the colour of kdes, this is a recreation of a plot nicked from the docs:


Screenshot 2023-03-10 at 02 15 25

and then this branch:

Screenshot 2023-03-10 at 02 14 41

Notice the bottom right panel is yellow instead of blue.

I worry there might be other small things like this which I haven't noticed, perhaps this should be discarded

Ormorod:
I think I've found the problem: pandas `_make_plot` picks a color from the cmap (not the 68% we want), and we ignore our cmap if a color is picked. Hopefully the duration of a coffee break will provide the answer

Ormorod:
Overriding `HistPlot._get_colors()` worked, though I have moved this up the inheritance tree to (our) `MPLPlot`. Is this too high @williamjameshandley @lukashergt ?

I think with that, this PR might finally be nearly over...

Ormorod:
I've just made doubly sure that my `HistPlot._make_plot()` matches the pandas 1.5.3 version exactly, apart from the specific changes I made for handling weights correctly. This area of the code has changed quite a lot for the release of pandas2.

williamjameshandley (approved):
OK, I think this should now be merged and tested in the wild. The only thing I worry about is the `_PlanePlot2d._make_plot` command, whose provenance is not entirely clear, in particular the `# another one of those hard-coded 0s` comment, but I think we should leave this in there for now to flag for the future if we find that this code misbehaves in some unforseen way.

Many thanks @Ormorod, this further unifies the approach with pandas, and is a substantial contribution!

PR #244, opened 2023-01-12
Conda CI AreaPlot issue

Ormorod:
# Description

The conda install CI finds an issue where one of the y limits of an `AreaPlot` is not zero for a `WeightedDataFrame`.

Solving this has proven difficult enough I think it deserves its own PR.

# Checklist:

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -81,7 +81,7 @@ jobs:
 
     - name: Test with pytest
       shell: bash -l {0}
-      run: python -m pytest --cov=anesthetic tests
+      run: python -m pytest --cov=anesthetic tests -s
 
     - name: Upload coverage to Codecov
       uses: codecov/codecov-action@v1
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -53,7 +53,7 @@ jobs:
       run: python -m pip install -r requirements_extras.txt 
 
     - name: Test with pytest
-      run: python -m pytest --cov=anesthetic tests
+      run: python -m pytest --cov=anesthetic tests -s
 
     - name: Upload coverage to Codecov
       uses: codecov/codecov-action@v1
--- a/anesthetic/plotting/_matplotlib/core.py
+++ b/anesthetic/plotting/_matplotlib/core.py
@@ -118,15 +118,4 @@ class BarhPlot(_BarhPlot, _WeightedMPLPlot):
 
 class AreaPlot(_AreaPlot, _WeightedMPLPlot):
     # noqa: disable=D101
-    def _post_plot_logic(self, ax, data):
-        LinePlot._post_plot_logic(self, ax, data)
-
-        is_shared_y = len(list(ax.get_shared_y_axes())) > 0
-        print(f"is_shared_y: {is_shared_y}")
-        print(f"ax.get_shared_y_axes(): {list(ax.get_shared_y_axes())}")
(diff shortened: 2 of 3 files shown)


Ormorod:
reverted to see what happens

Ormorod:
As I understand it, the issue is that the area plots sometimes think the axes are being shared, so they don't adjust the y limits. 

The most recent tests look like this, using the format (axes_df, axes_wdf) where appropriate
| OS | python | is_shared_y | ylim |
|- |- |-| - |
| ubuntu | 3.8 | False, True | 0, -0.134755 | 
| ubuntu | 3.10 | False, False | 0, 0 |
| MacOS | 3.8 | True, True | -0.134755, -0.134755 |
| MacOS | 3.10 | True, True | -0.134755, -0.134755 |
| Windows | 3.8 | | |
| Windows | 3.10 | | |

I have also noticed that the axes_wdf has an xlabel of "None", hinting at possibly another issue

Ormorod:
I've narrowed the problem down to two lines. In `test_BoxPlot()`, the plots created by grouping by `split` somehow sometimes survive, so that in `test_AreaPlot()` the y limit of 0 isn't applied. 

Can't say I know why this happens...

lukashergt (changes requested):
> Edit: perhaps it would be good to make more use of the pandas plot testing structure? Rather than plt.close("all") all over the shop

I fear we won't get around  using `plt.close('all')`, and I do think that that might have been the issue, here... Not easy to spot, but on the upside, at least it is easy to fix :)

As suggested inline, let's revert everything except the missing `plt.close('all')` line.

Ormorod:
> > Edit: perhaps it would be good to make more use of the pandas plot testing structure? Rather than plt.close("all") all over the shop
> 
> I fear we won't get around using `plt.close('all')`, and I do think that that might have been the issue, here... Not easy to spot, but on the upside, at least it is easy to fix :)
> 
> As suggested inline, let's revert everything except the missing `plt.close('all')` line.

I don't think this does fix it (but should be kept). The two lines I've commented out in `test_BoxPlot()` sometimes have side-effects on `test_AreaPlot()`, even with `plt.close("all")` added. Everything else is me exploring what is going on, and using the PR to test and record!

Ormorod:
Double-checked `plt.close("all")` isn't enough

lukashergt:
> Double-checked `plt.close("all")` isn't enough

That's a shame, but thanks for clarifying.

lukashergt:
> I don't think this does fix it (but should be kept). The two lines I've commented out in `test_BoxPlot()` sometimes have side-effects on `test_AreaPlot()`, even with `plt.close("all")` added. Everything else is me exploring what is going on, and using the PR to test and record!

Hmm, weird. Maybe it has to do with the `group_by` functionality? Does that create some hidden plot that doesn't get closed and then interferes with `test_AreaPlot`?

Ormorod:
I'll rerun these tests a few times, then remove the changes which are just to show the y limits of the `areaplot`s 

success 1/10

Ormorod:
I don't think this worked after all :(

williamjameshandley:
That is unbelievably frustrating...

Let's convene tomorrow morning in person to plan our strategy.

Ormorod:
^ Last attempt for today

> That is unbelievably frustrating...
> 
> Let's convene tomorrow morning in person to plan our strategy.

I can do 9-9:45am or ≥3pm (rowing preseason...)

Ormorod:
Dealt with in #248

PR #205, opened 2022-08-09
Perfect nested sampling functionality

williamjameshandley:
# Description

A set of scripts for generating perfect nested sampling runs which will prove useful in developing techniques such as importance sampling #132. 

Changes:
1. Implemented a spherical gaussian as found in John Skilling's original paper
2. Implemented a correlated gaussian, which can be used to simulate cosmological likelihoods
3. Moved the wedding cake class into a function in the same framework as the two above

# Checklist:

- [X] I have performed a self-review of my own code
- [X] My code is PEP8 compliant (`flake8 anesthetic tests`)
- [X] My code contains compliant docstrings (`pydocstyle --convention=numpy anesthetic`)
- [X] New and existing unit tests pass locally with my changes (`python -m pytest`)
- [X] I have added tests that prove my fix is effective or that my feature works

PR #246, opened 2023-01-16
API hints for existing users

### LINKED ISSUE #245: raise error for incorrect `'root'` usage
Raise error when `'root'` is incorrectly passed to `MCMCSamples` or `NestedSamples` instead of to `read_chains` in anesthetic 2.0 in order to warn old users of the new behaviour.

williamjameshandley:
# Description

This begins a PR for a general set of updates which aim to ease new users into the updated API. 

In the first instance it fixes #245 by ensuring that users trying the old setup of `NestedSamples(root='...')` are left with a `TypeError` that `root` is not an allowed keyword, rather than returning an empty array (which silently breaks a lot of existing scripts).

# Checklist:


tests/test_samples.py
```diff
@@ -1230,3 +1230,8 @@ def test_constructors():
     assert samples.T[0].islabelled()
 
     assert isinstance(samples['x0'].to_frame(), WeightedLabelledDataFrame)
+
+
+def test_old_gui():
+    with pytest.raises(TypeError):
+        MCMCSamples(root='./tests/example_data/pc')
```
lukashergt, line 1237: Use `Samples` instead of `MCMCSamples` here? That should cover both `MCMCSamples` and `NestedSamples`, too.

demo.py
```diff
@@ -62,21 +61,22 @@
...
 #| * unfilled contours  &  modifying individual axes:
 
-fig, axes = make_2d_axes(['omegabh2', 'omegach2', 'H0'], tex=mcmc.tex)
-mcmc.plot_2d(axes.iloc[0:1, :], types=dict(upper='kde', lower='kde', diagonal='kde'), fc=None);
-mcmc.plot_2d(axes.iloc[1:2, :], types=dict(upper='kde', lower='kde', diagonal='kde'), fc=None, cmap=plt.cm.Oranges, lw=3);
-mcmc.plot_2d(axes.iloc[2:3, :], types=dict(upper='kde', lower='kde', diagonal='kde'), fc='C2', ec='C3', c='C4', lw=2);
+fig, axes = make_2d_axes(['omegabh2', 'omegach2', 'H0'])
+mcmc.plot_2d(axes.iloc[0:1, :], kind=dict(upper='kde_2d', lower='kde_2d', diagonal='kde_1d'), fc=None);
+mcmc.plot_2d(axes.iloc[1:2, :], kind=dict(upper='kde_2d', lower='kde_2d', diagonal='kde_1d'), fc=None, cmap=plt.cm.Oranges, lw=3);
+mcmc.plot_2d(axes.iloc[2:3, :], kind=dict(upper='kde_2d', lower='kde_2d', diagonal='kde_1d'), fc='C2', ec='C3', c='C4', lw=2);
```
lukashergt, line 79: We'll want to raise a warning for the use of `'kde'` instead of `'kde_2d'`, since `'kde'` is the pandas version, whereas we would want to use anesthetic's own `'kde_1d'` and `'kde_2d'`.

tests/test_samples.py
williamjameshandley: Testing for all three now.

demo.py
williamjameshandley: This is done in 3faf577. I'm going to also put in hints for:
- tex
- ns_output
- fig, axes (if I can)
We should have a think if there are any other gotcha's, and add an issue to eventually remove these hints at another minor release.

anesthetic/samples.py
```diff
@@ -300,6 +308,14 @@ def plot_2d(self, axes, *args, **kwargs):
...
                     lkwargs = local_kwargs.get(pos, {})
                     lkwargs['kind'] = kind.get(pos, None)
+                    # TODO: remove this in version >= 2.1
+                    if lkwargs['kind'] == 'kde':
+                        warnings.warn(
+                            "You are using \'kde\' as a plot kind. "
+                            "\'kde_1d\' and \'kde_2d\' are the appropriate "
+                            "keywords for anesthetic. Your plots may look "
+                            "odd if you use this argument."
+                            )
```
lukashergt, line 318: I believe the same should go for `lkwargs['kind'] == 'hist'`. So probably good to add that here, too.

anesthetic/samples.py
```diff
@@ -196,6 +196,14 @@ def plot_1d(self, axes, *args, **kwargs):
         kwargs['kind'] = kwargs.get('kind', 'kde_1d')
         kwargs['label'] = kwargs.get('label', self.label)
 
+        # TODO: remove this in version >= 2.1
+        if kwargs['kind'] == 'kde':
+            warnings.warn(
+                "You are using \'kde\' as a plot kind. "
+                "\'kde_1d\' is the appropriate keyword for anesthetic. "
+                "Your plots may look odd if you use this argument."
+                )
```
lukashergt, line 205: I believe the same should go for `lkwargs['kind'] == 'hist'`. So probably good to add that here, too.

demo.py
lukashergt: > We should have a think if there are any other gotcha's, and add an issue to eventually remove these hints at another minor release.

Another few that come to mind:
* kwarg `kind` instead of the previous `types` (2d) or `plot_type` (1d)
* `D_KL` instead of `D`, as well as `d_G` instead of `d` (and hint that there is now also `logL_P`?)

Do we want to add those in this PR?

lukashergt (commented):
See some inline comments.

Since this PR overhauls the `demo.py`. Should we integrate that into our tests, i.e. create a test that checks that it runs without errors? For that we would have to either change the data used in `demo.py`, or create some mock data in our `example_data` folder that has the required parameters, i.e. `'omega_b'`, `'tau'` etc.

anesthetic/samples.py
williamjameshandley: done in 549979c

demo.py
williamjameshandley: Done in 1ef6cc7

anesthetic/samples.py
williamjameshandley: Done in 549979c

lukashergt (approved):
> Since this PR overhauls the `demo.py`. Should we integrate that into our tests, i.e. create a test that checks that it runs without errors? For that we would have to either change the data used in `demo.py`, or create some mock data in our `example_data` folder that has the required parameters, i.e. `'omega_b'`, `'tau'` etc.

We said we might do this in #205 instead, right? In that case this looks good now. Thanks, Will.

Ormorod:
This change has triggered warnings in `test_samples.py`, I think would be fixed by `"hist"` -> `"hist_2d"`?

https://github.com/williamjameshandley/anesthetic/blob/e308f9e1eacf03ff320c1840471dd940c45fb085/tests/test_samples.py#L189-L191

PR #254, opened 2023-01-20
Close figures on teardown

Ormorod:
# Description
First pass at adding fixtures to close figures at the end of each test

Personally, I think it would be more futureproof to make all the affected tests methods of a subclass of `pandas.tests.plotting.common.TestPlotBase`


williamjameshandley:
> Personally, I think it would be more futureproof to make all the affected tests methods of a subclass of `pandas.tests.plotting.common.TestPlotBase`

I'd rather keep things consistent, as (a) this is easier for new contributers to add tests and (b) we've only just got it working deterministically, and I'd rather keep things simple. Perhaps put this as an issue for now for discussion, and  we should just implement the fixtures as was started in #248 if we want to deal with too many open plots.

Ormorod:
flake8 is complaining that the fixture isn't used in the files where it is imported. Any ideas how to avoid this, apart from repeating the definition in each file it is used?

williamjameshandley:
No. I suggest just code repetition in this instance -- it's only for a test suite, and I think it's more explicit/Pythonic to put the fixture in there.

williamjameshandley (changes requested):
Looking good.

If you have a moment, is it possible to put in a CI test that checks that all tests including `matplotlib` have this fixture?

The bash one-liner is 

```bash
$ grep -L 'close_figures_on_teardown' $(grep -l 'matplotlib' tests/**/*.py)
tests/matplotlib_agg.py
tests/test_gui.py
``` 

The first of these will be removed in #205, but the second should probably be included!

Not sure how easy it is to include bash tests in the github actions, but this would probably be a good thing to have in general (e.g. testing version bumps). [PolyChordLite](https://github.com/PolyChord/PolyChordLite/blob/master/.github/workflows/check_version.yml) hase something that may be reusable.

Ormorod:
Why does `matplotlib_agg.py` needs the fixture?

Ormorod:
Workflow should now be working, aside from I haven't bothered to use is/are properly in the error description.

I'll still create an issue for moving to classes of tests, as I really do think that is a much cleaner way of adding teardown methods, this way requires us to remember to change the CI if an additional teardown is needed in the future, and as you can see above I find messing around with CI a massive faff. Only downside is the extra step of indentation makes complying with flake8 that much more difficult (which I think deserves its own issue...)

williamjameshandley (approved):
Excellent -- well done on the CI. Please squash and merge

PR #255, opened 2023-01-29
`.to_numpy()`

Ormorod:
Use `to_numpy()` on boolean mask in Gaussian example to prevent `UserWarnings`

I seriously think we need to have something which resists PRs introducing new warnings...

### DIFF
--- a/anesthetic/examples/perfect_ns.py
+++ b/anesthetic/examples/perfect_ns.py
@@ -53,8 +53,10 @@ def gaussian(nlive, ndims, sigma=0.1, R=1, logLmin=-1e-2):
         r = (points**2).sum(axis=-1, keepdims=True)**0.5
 
     samples = merge_nested_samples(samples)
-    logLend = samples[samples.nlive >= nlive].logL.max()
-    return samples[samples.logL_birth < logLend].recompute()
+    print("here0")
+    logLend = samples[(samples.nlive >= nlive).to_numpy()].logL.max()
+    print("here1")
+    return samples[(samples.logL_birth < logLend).to_numpy()].recompute()
 
 
 def correlated_gaussian(nlive, mean, cov, bounds=None):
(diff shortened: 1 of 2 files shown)


lukashergt:
> I seriously think we need to have something which resists PRs introducing new warnings...

There is the option of the `pytest -W error` flag, which will turn warnings into errors... However, I think this will get in the way, e.g. this could prevent quickly fixing a bug because of completely unrelated deprecation warnings. 

We could think about having a single test run with the `pytest -W error` flag switched on, which we then set as "allowed to fail". This can prompt both PR author and reviewer to check more closely for warnings and to try to prevent introducing new ones. 

Thoughts? @Ormorod @williamjameshandley

williamjameshandley:
I think that a `.loc` is better from a pandas perspective than a `.to_numpy()`, and I think that this removes the warnings as well. I've pushed this change, but feel free to revert.

I agree with @lukashergt that a requirement will introduce too much friction, but a check is entirely reasonable. Just add it as an extra block in the CI.yaml file, and I won't flag it as required. I'll approve this, and give you the option to implement the test to encourage users to fix warnings, or just to merge as-is.

williamjameshandley (approved):

Ormorod:
I thought Lukas didn't like `.loc`, which is why I used `.to_numpy()`

lukashergt:
>> I think that a `.loc` is better from a pandas perspective than a `.to_numpy()`, and I think that this removes the warnings as well. I've pushed this change, but feel free to revert.
> 
> I thought Lukas didn't like `.loc`, which is why I used `.to_numpy()`

Yes, watch out when trying boolean masks on pandas dataframes. When we think of boolean masks, we tend to have a picture in mind like what you get when applying `.to_numpy()`, so that would be the safer route than the current `.loc` syntax, where `samples.logL_birth < logLend` is a `WeightedDataFrame`. 

All that said, if we want a more pandas way of doing things, then the `.query()` method is the correct approach, e.g. instead of
```python
samples.loc[(samples.nlive >= nlive)].logL.max()
```
do
```python
samples.query("nlive >= @nlive").logL.max()
```

PR #243, opened 2022-12-23
Update documentation for 2.0 release

lukashergt:
I have started a more extensive sphinx documentation for the upcoming anesthetic 2.0 release. 

I am making use of `matplotlib.sphinxext.plot_directive` for auto-generating plots from code snippets within the documentation. This should help in keeping the documentation up to date with any changes, as breaking changes will flag up as errors when building the documentation. In that regard I thought it might be good to incorporate a check that the docs are building within our CI pipeline. Maybe at the end of the `lint` workflow? Ideas @williamjameshandley?

I've started moving some of our `demo.py` code into the documentation, will continue when I get to it. Any suggestions for further things that would profit from documenting?

Also, I've fixed the formatting of a bunch of docstrings to properly work with sphinx' autodoc feature, but I probably have missed a bunch, so if anyone spots something in the auto-generated API documentation, let me know.

By the way, for building the documentation locally, make sure to have `sphinx` and `numpydoc` installed, and then it should be a question of running `make html` in the `docs` folder, then opening one of the html files in the build folder with your favourite browser.

# To-Do:

- [x] Add documentation on DataFrame handling and the `Samples` class (compute mean/std/median/quantiles/etc.; whats special/different compared to pandas; creating new parameters, ...?)
- [x] Add documentation for `MCMCSamples` (chains, importance sampling, ...?)
- [x] Add documentation for `NestedSamples` (beta, prior samples, logZ, D_KL, d_G, logL_P, ...?)
- [x] Add some prose to the `quickstart.rst`

Fixes #62  

# Checklist:

- [x] I have performed a self-review of my own code
- [x] My code is PEP8 compliant (`flake8 anesthetic tests`)
- [x] My code contains compliant docstrings (`pydocstyle --convention=numpy anesthetic`)
- [x] New and existing unit tests pass locally with my changes (`python -m pytest`)
- [x] I have added tests that prove my fix is effective or that my feature works

PR #258, opened 2023-02-08
`MCMCSamples` methods for burn-in removal and Gelman--Rubin statistic

### LINKED ISSUE #219: remove_burn_in function for MCMCSamples
As [pointed out](https://github.com/williamjameshandley/anesthetic/pull/217#discussion_r946868921) by @lukashergt in #217, currently only chain readers are capable of burn-in. This (repeated) code should be moved into MCMCSamples, where it would correctly handle the chain column if it exists (perhaps using a reindex and a pandas groupby).

lukashergt:
This PR introduces new `MCMCSamples` methods that are particular to MCMC chains:
* `remove_burn_in`: discards the first few samples
* `Gelman_Rubin`: computes the Gelman--Rubin `Rminus1` convergence statistic

Fixes #219 

# Checklist:


lukashergt:
With the new `remove_burn_in` method, we could now ask ourselves whether we would like to make this the only way of getting rid of burn-in samples, i.e. remove the `burn_in` kwargs in the readers and remove the `read/utils.py` file. This would be a major change, so should be part of the 2.0.0 milestone.

I'd be happy to handle it that way. That makes it quite clear that always the full data is read in. And if you would like to get rid of burn-in samples, then that should be an active part of post-processing. It can even be done within the same line as the chain reading: `mcmc = read_chain("root").remove_burn_in(100)`.

Thoughts, @williamjameshandley?

lukashergt:
I'm running into new CI issues (unrelated to the changes of this PR...), making me think that it might be better to turn off parallel tests again...

anesthetic/samples.py
```diff
@@ -482,6 +482,102 @@ class MCMCSamples(Samples):
...
+                ('chain', '$n_\\mathrm{chain}$')
+        )
+        nchains = chains.ngroups
+
+        # Within chain variance ``W``
+        # (average variance within each chain):
+        covs = chains.cov()
+        W = covs.groupby(level=['params', 'labels']).mean().to_numpy()
+        # TODO: the above line should be a weighted mean
+        # --> need to fix groupby for WeightedDataFrames!
```
lukashergt, line 568: Thanks to our `WeightedDataFrames` the computation of the Gelman--Rubin statistic is really quite neat. There is this one issue that the data frames resulting from a `groupby` action should in general be weighted data frames themselves, which they currently are not.

A simple example is the mean of a parameter across the full weighted dataset should be the same as the (weighted!) mean of the means of all individual chains, i.e. we should have (in slightly simplified syntax for readability):
```python
assert samples.groupby('chain').mean().mean() == samples.mean()
```
Note the double `.mean()`, the first produces the within chain means that themselves should be weighted, and the mean of these chain means should match the overall mean.

williamjameshandley:
>I'm running into new CI issues (unrelated to the changes of this PR...), making me think that it might be better to turn off parallel tests again...

In the interests of sanity I have turned this off e9b0749

anesthetic/samples.py
```diff
@@ -482,6 +482,102 @@ class MCMCSamples(Samples):
...
+        could overwrite the ``'chain'`` column, splitting the samples into two
+        or more sets.
+
+        Parameters
+        ----------
+        params: list(str)
+            List of column names (i.e. parameters) to be included in the
+            convergence calculation.
+            Default: all parameters (except those parameters that contain
+            'prior', 'chi2', or 'logL' in their names)
```
williamjameshandley, line 542: This by default would include derived parameters -- is this how cobaya does it?

tests/test_samples.py
```diff
@@ -395,6 +395,46 @@ def test_hist_levels():
                levels=[0.95, 0.68], bins=20)
 
 
+def test_mcmc_stats():
+    mcmc = read_chains('./tests/example_data/cb')
+    chains = mcmc.groupby(('chain', '$n_\\mathrm{chain}$'), group_keys=False)
+    n0 = chains.count().iloc[0, 0]  # number samples in first chain
+    mcmc_head = chains.head(200).copy()
+    mcmc_asym = mcmc.remove_burn_in(burn_in=[2, 10])
```
williamjameshandley, line 403: Should we test asymmetric floats as well as integers?

williamjameshandley (changes requested):
>With the new remove_burn_in method, we could now ask ourselves whether we would like to make this the only way of getting rid of burn-in samples, i.e. remove the burn_in kwargs in the readers and remove the read/utils.py file. This would be a major change, so should be part of the 2.0.0 milestone.
I'd be happy to handle it that way. That makes it quite clear that always the full data is read in. And if you would like to get rid of burn-in samples, then that should be an active part of post-processing. It can even be done within the same line as the chain reading: mcmc = read_chain("root").remove_burn_in(100).

Providing this is clear in the examples then this is indeed conceptually neater. We would need to put in a temporary error which informs users in the initial release in how to update their code.

You should go ahead and remove the corresponding code for burn in at read-time

anesthetic/samples.py
lukashergt: I believe so, and I don't see a reason why they shouldn't be included.

anesthetic/samples.py
lukashergt: We could think about an optional `skipna` kwarg, but yes, by default I think it should be `skipna=True` which should result in dropping rows. For that particular case I'd be happy to accept a discrepancy between the full mean and the groupby mean.

tests/test_samples.py
lukashergt: I've heavily expanded on the tests to cover the various possible inputs (see https://github.com/williamjameshandley/anesthetic/pull/258/commits/95b90263f39ab038206cedf6bf93a838de580ea1). 

I've even included negative inputs (which effectively function as a "keep" rather than a "remove" --> also added to the docstring in https://github.com/williamjameshandley/anesthetic/pull/258/commits/a25117f66dc733f00ce62beca31d135379b2a3dc).

lukashergt:
> Excellent work! I presume this is functionality you've tested in the wild a little?

Yes, I played a bit with this for the BeyondPlanck stuff.



> Providing this is clear in the examples then this is indeed conceptually neater. We would need to put in a temporary error which informs users in the initial release in how to update their code.
> 
> You should go ahead and remove the corresponding code for burn in at read-time

Done in https://github.com/williamjameshandley/anesthetic/pull/258/commits/0c65b4469781366e552a299f2b259d0d2d36fad6.

anesthetic/samples.py
lukashergt: Should I turn this into an issue for now, to be addressed in a PR targeted at `groupby`? 

While the current situation isn't perfect, I think in practice for sufficiently long chains this shouldn't matter quite that much.

anesthetic/read/chain.py
```diff
@@ -31,6 +22,15 @@ def read_chains(root, *args, **kwargs):
...
 
     """
+    if 'burn_in' in kwargs:
+        raise KeyError(
+            "This is anesthetic 1.0 syntax. The `burn_in` keyword is no "
+            "longer supported in `read_chains`. You need to update, e.g.\n"
+            "read_chains(root, burn_in=0.5)         # anesthetic 1.0\n"
+            "read_chains(root).remove_burn_in(0.5)  # anesthetic 2.0\n"
+            "See also https://anesthetic.readthedocs.io/en/latest/"
+            "anesthetic.html#anesthetic.samples.MCMCSamples.remove_burn_in"
```
williamjameshandley, line 32: We should make a note in #243 to update [this page](https://anesthetic.readthedocs.io/en/latest/reading_writing.html).

williamjameshandley (approved):
Looks good -- please squash and merge!

anesthetic/samples.py
williamjameshandley: >Should I turn this into an issue for now, to be addressed in a PR targeted at groupby?

Please do

lukashergt: Done in #260.

anesthetic/read/chain.py
lukashergt: Done in https://github.com/williamjameshandley/anesthetic/pull/243/commits/c69793e549e1db1c85c130e2a61bda495b70cdad

PR #262, opened 2023-02-12
Improved packaging mark 2

williamjameshandley:
# Description

This follows direct from #259 as planned, correcting the 3.8/3.9/3.10 simultaneous build & uploads (which are not necessary for a pure python distribution), as well as now putting in an explicit version increment.

# Checklist:


williamjameshandley:
One thing which is not ideal is that this additional check will break the github automatic merging (e.g. when a new PR has been merged, and one would want to press the 'update branch' button on existing branches, the tests would fail). This annoyance is compounded in that at the moment a successful 'update branch' does not make reviews stale, but since this would require manual intervention it would need to be re-approved. 

Options:
1) Continue as planned: version numbering is important, should be confirmed by more than one person as to whether it is a patch or an update (minor/patch number).
2) Make version incrementing a warning rather than requirement (this means that the version number in the readme/tag does not correspond to a single patch).
3) Go back to a manual release model, with PR's more fine-grained than releases, but pypi/conda uploads tied to release creation rather than merges to master

I don't think we can automate this process with e.g. server-side hooks as github doesn't allow this.

Thoughts/ideas @lukashergt @Ormorod ?

Ormorod:
I've been pondering this, I think 1) might actually be good, as the failed test with the master merge serves as a notification for the PR to actually check what the changes are, and remember to git pull before they make any more commits.

williamjameshandley:
Here is the problem in action!

lukashergt:
> One thing which is not ideal is that this additional check will break the github automatic merging (e.g. when a new PR has been merged, and one would want to press the 'update branch' button on existing branches, the tests would fail).

Just to make sure I understand correctly, the scenario that you are describing is that in case of two PRs with e.g. the same version increment, where both have been reviewed and accepted, the merging of the first and updating of the second, will make the tests of the second fail, and any further commits will remove the previous approval. Is that the issue? 

Regarding the options:

> * Continue as planned: version numbering is important, should be confirmed by more than one person as to whether it is a patch or an update (minor/patch number).

This would work for me. If the only new commit is a version update, then any review should hopefully be fast and easy.

> * Make version incrementing a warning rather than requirement (this means that the version number in the readme/tag does not correspond to a single patch).

I don't like this. I'd much prefer a slight inconvenience a la option 1 over the risk of lacking potentially important version updates or even inconsistent version numbers. 

> * Go back to a manual release model, with PR's more fine-grained than releases, but pypi/conda uploads tied to release[/tag](https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/#publishing-the-distribution-to-pypi-and-testpypi) creation rather than merges to master

Does this still drop the version checks? Then I don't like it, same reason as option 2.

Overall, I think it can be really helpful to ensure consistent versioning. Especially for people that like to work with the latest version of the master branch (which is us...), what can happen is that we are producing paper code based on 3 commits after version 1.2.3, but unknown to that current self it will be another 5 commits before 1.2.4. Then during our zenodo upload we want to specify which version we used, we put in the latest version 1.2.3. A few (or many) version updates later a student tries to reproduce the results, and potentially can't rely on the advertised version but has to try out different commits. It would be nice to avoid that.

appetrosyan:
Consider 
https://github.com/marketplace/actions/github-action-for-committing-changes-to-a-repository
In that case, the process could look like this. 

1. Nobody changes the version in the code when creating a PR. 
2. A comment is used to determine if the version bump is major/minor/patch if not enough comments, fail the CI check. 
3. The PR gets merged, then a commit is created which contains just the version bump. 

## Benefits: 
- No need to review version bumps. Branch protection works as usual. 
- No need to manually bump versions or ask people to bump them. 
- Automatic PyPi/conda publishing. 
- Can automatically resolve voting conflicts. Editing comment re-triggers workflow. The highest number wins, so if one votes patch and other major, the major version is bumped. 

## Drawbacks: 
- Comment-based workflows are less reliable
- Unbumping version is almost impossible
- Requires scripting in CI. GitHub actions aren't the most usable in that mode.

williamjameshandley:
@appetrosyan's suggestions would be a good way to implement automation in option 3, though I can see it being as much of a headache as manual versioning given the unreliability of comment based workflows.

I propose we try option 1 for now, and see how inconvenient it is. Conveniently this just means @lukashergt or @Ormorod approving this.

Note I have just updated the `CI.yml` file to `CI.yaml`, and moved to a `pip install -e` model to (a) remove the requirement for a temporary requirements.txt file, and (b) allow the conda tests to pass since the removal of the requirements.txt intermediate file encodes the most up-to-date versions of packages (which might be available on pip but not on conda). For the sake of sanity I have just hard-coded `numpy pandas scipy matplotlib` without versions for the conda install tests.

.github/workflows/test_build.yaml
```diff
@@ -3,15 +3,12 @@ on: push
...
-      matrix:
-        python-version: ['3.8', '3.9', '3.10']
     steps:
     - uses: actions/checkout@master
-    - name: Set up Python ${{ matrix.python-version }}
+    - name: Set up Python 3.8
       uses: actions/setup-python@v2
       with:
-        python-version: ${{ matrix.python-version }}
+        python-version: 3.8
```
Ormorod, line 11: Why 3.8 and not 3.10? (or even 3.11)

.github/workflows/CI.yaml
```diff
@@ -62,15 +62,12 @@ jobs:
...
-        python -m piptools compile -o requirements.txt pyproject.toml
-        python -m pip install -r requirements.txt
+        python -m pip install -e ".[test]"
 
     - name: Install extra dependencies
       if: ${{ matrix.extras }} 
       run: |
-        python -m piptools compile --extra extras -o requirements_extras.txt pyproject.toml
-        python -m pip install -r requirements_extras.txt 
+        python -m pip install -e ".[extras]"
```
Ormorod, line 70: Does this need to be an editable install?

.github/workflows/test_build.yaml
lukashergt: I think 3.8 as our minimum requirement works for the build test.

lukashergt (approved):
Happy with this if you are.

pyproject.toml
```diff
@@ -21,7 +21,7 @@ authors = [
 description = "nested sampling post-processing"
 readme = "README.rst"
 license = {file = "LICENSE"}
-requires-python = ">=3.7"
+requires-python = ">=3.8"
```
appetrosyan, line 24: Just FYI, 3.8 is EOL in 2024.

PR #267, opened 2023-02-22
Improved packaging mark 5

williamjameshandley:
# Description

Steadily getting there (#259 #263 #265 #266)...

This one moves the testpypi build and publish to the CI file, which means that it is done alongside the other CI checks, importantly only on master/pull requests. The advantage of this is that the tests do not fail as they do (e.g. on ##243) when people are pushing to a fork (for which testpypi inevitably lacks the secret API token).

I've also removed the `.coveragerc`, which only tells codecov to ignore a file that has not been necessary since #65.

# Checklist:

- [X] I have performed a self-review of my own code
- [X] My code is PEP8 compliant (`flake8 anesthetic tests`)
- [X] My code contains compliant docstrings (`pydocstyle --convention=numpy anesthetic`)
- [X] New and existing unit tests pass locally with my changes (`python -m pytest`)
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] I have appropriately incremented the [semantic version number](https://semver.org/) in both README.rst and anesthetic/_version.py

PR #269, opened 2023-02-27
Improved packaging mark 6

williamjameshandley:
# Description

Unfortunately #267 did not fix the issue with secrets due to the fact that these are [not sharable](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/) to PRs. We can either remove this check (not wise, since the build step is very useful to check if anything has caused the build to fail before merging and incrementing the version number), or just allow the final step to fail silently when it is unable to upload (which is what this PR proposes).

I also reordered some of the build so that the automatic git tag reflects the version number in the corner case that the build fails for some reason in between PR and merge.

# Checklist:


lukashergt (approved):
The issue lies with untrusted forks not having access to the GitHub secrets, right?

Is there not an option of labelling certain forks as trusted? 

Otherwise, in [this CircleCI post](https://circleci.com/blog/triggering-trusted-ci-jobs-on-untrusted-forks/) they introduce an option for a reviewer with write privileges to manually launch an action (after a check of the changes) that would otherwise not even launch for a PR from an untrusted fork. Is there maybe something similar for GitHub actions?

williamjameshandley:
>The issue lies with untrusted forks not having access to the GitHub secrets, right?

Yeah, and I don't necessarily disagree with them here

>Is there not an option of labelling certain forks as trusted?

Not that I've found.

>Otherwise, in [this CircleCI post](https://circleci.com/blog/triggering-trusted-ci-jobs-on-untrusted-forks/) they introduce an option for a reviewer with write privileges to manually launch an action (after a check of the changes) that would otherwise not even launch for a PR from an untrusted fork. Is there maybe something similar for GitHub actions?

Not obviously -- I think github is quite security conscious (see our unfilled 'Security' box in 'community standards' tab in Insights).

PR #270, opened 2023-03-01
Changes for Pandas 2

AdamOrmondroyd:
# Description

I'm conscious that pandas 2 is on the horizon, with the release of 2.0.0rc0 onto pypi last week.

Some of these changes affect #235, so I have built on top of that here.

Changes:
- [x] implement `_PlanePlot2d._args_adjust()`
- [x] Remove `HistPlot._makeplot()`, copied from pandas #48888 (which is already outdated)

Still to fix:
- [ ] The Gelman-Rubin statistic is broken, as the `groupby()` seems to cause the pandas kwargs to `cov()`to be passed to our `cov()`, which raises the `NotImplementedError` - need to see how this fits in with #272
- [x] adjust_docstrings() is broken
- [x] update pandas requirement to ~= 2.0.0 once it actually releases

Fixes # (issue)

# Checklist:

- [ ] I have performed a self-review of my own code
- [x] My code is PEP8 compliant (`flake8 anesthetic tests`)
- [x] My code contains compliant docstrings (`pydocstyle --convention=numpy anesthetic`)
- [ ] New and existing unit tests pass locally with my changes (`python -m pytest`)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] I have appropriately incremented the [semantic version number](https://semver.org/) in both README.rst and anesthetic/_version.py - I think this is worthy of a minor increment

PR #272, opened 2023-03-10
WeightedGroupBy

### LINKED ISSUE #260: `WeightedDataFrame.groupby.mean` does not produce weighted means
## Describe the bug

There are two tiers to this bug:
1. Currently the means (and other statistics) after a pandas `groupby` method call from a `WeightedDataFrame` are not weighted means . The weights are being ignored. 
2. The groups themselves should be weighted objects (sum of weights within the group), such that the mean of the group mean matches the overall sample mean. So we need some sort of `WeightedGroup` object, such that weights are taken into account when performing between-group statistics.

## To Reproduce

```python
import numpy as np
from anesthetic.samples import Samples

params = ['a', 'b']
data = np.random.rand(4, 2)
weights = np.random.randint(1, 10, 4)
samples = Samples(data, weights=weights, columns=params)
samples['group'] = np.ones(4, dtype=int)
samples.loc[2:, 'group'] = 2

group1_means  = np.average(samples.loc[:1, params], axis=0)
group1_wmeans = np.average(samples.loc[:1, params], axis=0, weights=samples.get_weights()[:2])
group2_means  = np.average(samples.loc[2:, params], axis=0)
group2_wmeans = np.average(samples.loc[2:, params], axis=0, weights=samples.get_weights()[2:])
group_means  = np.vstack([group1_means,  group2_means])
group_wmeans = np.vstack([group1_wmeans, group2_wmeans])
group_weights = [samples.get_weights()[:2].sum(), samples.get_weights()[2:].sum()]
mean =  np.average(group_means,  axis=0)
wmean = np.average(group_wmeans, axis=0, weights=group_weights)
```

## Expected behavior

The mean values after a groupby call should be weighted means; however, currently the weights are being ignored.

```python
assert not np.any(samples.groupby('group').mean() == group_means)
assert np.all(samples.groupby('group').mean() == group_wmeans)
```

Furthermore, the mean of the group means should match the overall sample mean:

```python
assert not np.any(samples.groupby('group').mean().mean() == mean)
assert np.all(samples.groupby('group').mean().mean() == wmean)
```

## Additional context

This follows up from the discussion in https://github.com/williamjameshandley/anesthetic/pull/258#discussion_r1099617574, and as such is relevant for the accurat computation of the Gelman--Rubin statistic. In that comment we also discussed the potential issue of having missing values.

## Screenshots

image

AdamOrmondroyd:
# Description

Statistics of `WeightedDataFrame.groupby` such as `mean()` ignore weights. Since my ambitious attempt to fix this in pandas itself is looking very unlikely to get through, I have attempted to fix it here. However, they have suggested adding a `self._gb_cls` attribute to `Series` and `DataFrame`, which `.groupby()` returns. For now, I have copied the methods.

Fixes #260

# Checklist:

- [x] Get the `mean().mean()` case working (test currently commented out). This seems like it would require significantly more work...


Ormorod:
I'm aware coverage won't be complete as the test only considers the mean.

lukashergt:
>  Since my ambitious attempt to fix this in pandas itself is looking very unlikely to get through

The latest comment sounded in favour of the fix, so I'd still have hope...

What does the `transform` in `udf.groupby('group').transform('mean')` do? Is that something we could use/leverage?

Ormorod:
> What does the `transform` in `udf.groupby('group').transform('mean')` do? Is that something we could use/leverage?

`pandas.core.groupby.DataFrameGroupBy.transform(func)` takes func and applies it to each group. What we might be able to use/create is something like `.transform('weight') to calculate the total weight of each group, which might be useable for the mean of means...

Ormorod:
@lukashergt The failing tests are to do with building documentation, which I'll be honest I could do with some pointers of what you might like to see here.

lukashergt:
> @lukashergt The failing tests are to do with building documentation, which I'll be honest I could do with some pointers of what you might like to see here.

See https://github.com/williamjameshandley/anesthetic/pull/272/commits/d97b760f2a59a522ebb7c634ddf4c9361c586f8f, https://github.com/williamjameshandley/anesthetic/pull/272/commits/bfa2647e4c2082a14178e6402845f8b130fb8501, https://github.com/williamjameshandley/anesthetic/pull/272/commits/40cd004d8fcfb3edc55512069d374b89a53d3864, and https://github.com/williamjameshandley/anesthetic/pull/272/commits/0f2104e26ce78b6700de35c61e4576492c97ffd7 and commit messages therein for how I've tackled this.

anesthetic/weighted_pandas.py
```diff
@@ -405,15 +487,64 @@ def _constructor_sliced(self):
...
-    adjust_docstrings(cls, 'SeriesGroupBy',
-                           'pandas.core.groupby.SeriesGroupBy')
     adjust_docstrings(cls, 'pandas.core.window.Rolling.quantile',
                            'pandas.core.window.rolling.Rolling.quantile')
+    adjust_docstrings(cls, r'\bDataFrameGroupBy\b', 'WeightedDataFrameGroupBy')
+    adjust_docstrings(cls, r'\bSeriesGroupBy\b', 'WeightedSeriesGroupBy')
+    adjust_docstrings(cls, 'WeightedDataFrameGroupBy.sample',
+                           'pandas.core.groupby.DataFrameGroupBy.sample')
+    adjust_docstrings(cls, 'WeightedSeriesGroupBy.sample',
+                           'pandas.core.groupby.SeriesGroupBy.sample')
```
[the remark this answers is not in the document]
Ormorod: Do you mean adapting the methods themselves or their associated docstrings?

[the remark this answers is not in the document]
lukashergt: I meant the methods themselves. The problem on the doc level is that the pandas docs have sprinkled links to other methods throughout. Our `adjust_docstrings` reroutes anything pointing to `DataFrame` or `Series` to `WeightedDataFrame` or `WeightedSeries`. Specifically the line
```python
adjust_docstrings(cls, r'\bDataFrameGroupBy\b', 'WeightedDataFrameGroupBy')
```
causes the link from a pandas docstring to be changed from `DataFrameGroupBy.sample` to `WeightedDataFrameGroupBy.sample`. But as long as there is no `WeightedDataFrameGroupBy.sample` method, the link fails. 

Currently the line
```python
adjust_docstrings(cls, 'WeightedDataFrameGroupBy.sample', 'pandas.core.groupby.DataFrameGroupBy.sample')
```
routes the link back to `DataFrameGroupBy.sample`. Alternatively, if we were to implement the  `WeightedDataFrameGroupBy.sample` method, then the link would work.

williamjameshandley:
I'm looking into resolving the final weights issues today. 

Is there any reason we've opted to include `numeric_only` in the new `WeightedGroupBy` functions as a kwarg? I opted to not bother with them in the other `WeightedSeries` and `WeightedDataFrame` functions, (apart from bizarrely quantile, which I could just drop support for).

Would you be put out if I just dropped those for now? (it would also make the functions look a lot neater). Modern Pandas includes more arguments anyway, such as `engine`, which I don't imagine we'll support natively initially.

Ormorod:
I haven't yet had chance to play further with `WeightedGroupBy`, from Lukas' comment above it sounds like there might be quite a lot which I still need to implement. (e.g. I haven't got around to understanding what `WeightedGroupBy.sample` actually is...)

My approach with `numeric_only` was just to change as little as possible wrt pandas itself, as this has caused me grief with `cov` specifically (see #270) and I worry more of this will appear

anesthetic/weighted_pandas.py
```diff
@@ -1,14 +1,67 @@
...
+
+    def mean(self, *args, **kwargs):  # noqa: D102
+        result = self.agg(lambda df: self.obj._constructor(df).mean(
+            *args, **kwargs))
+        return result.__finalize__(self.obj, method="groupby")
+
+    def std(self, *args, **kwargs):  # noqa: D102
+        result = self.agg(lambda df: self.obj._constructor(df).std(
+            *args, **kwargs))
+        return result.__finalize__(self.obj, method="groupby")
```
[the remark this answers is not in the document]
williamjameshandley: I agree that it's far from ideal, but I don't see the alternative as being maintainable long-term as new arguments are added in future versions of pandas. In principle, if desired, we could implement a sensible ddof, but I'm only inclined to do that if there is a clear use case.

williamjameshandley:
OK, I've now implemented the mean-of-means version of the code, and added some tests which show this is working. I also implemented a weighted `_gotitem` (which just replaces the existing code with `WeightedSeriesGroupBy`. Without this, `samples.groupby('chain').x0` returned a `SeriesGroupBy` rather than a `WeightedSeriesGroupBy`.

Ormorod:
Getting performance warnings  `/Users/adam/programming/anesthetic/anesthetic/weighted_pandas.py:27: PerformanceWarning: dropping on a non-lexsorted multi-index without a level parameter may impact performance.`

anesthetic/weighted_pandas.py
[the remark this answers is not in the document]
lukashergt: I would much rather have the explicit args and kwargs, and _that_ would actually be the pythonic way, but we can't maintain a full copy of pandas and keep it up to date with all the little changes going on there. Hence it would have been much nicer to have GroupBy automatically point to the corresponding methods, without needing to create a GroupBy method for every WeightedDataFrame method that we want to use...

For any anesthetic-own functions I would strive to be explicit, but when it comes to inheritance from pandas, I would try to keep things as independent and robust to changes in pandas as possible.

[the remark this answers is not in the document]
lukashergt: Specifically `ddof` is probably something we could improve upon. `ddof` actually _does_ make sense for weighted data. For MCMC data (with integer weights, aka repeat type weights, aka frequency type weights) you can use it actually one to one compared to unweighted data. For nested samples (which I guess would fall into the category of reliability type weights) I am not sure what the best way of handling this would be, but just dividing by the sum of weights does not give you an unbiased estimator...

See the corresponding [Wikipedia article](https://en.wikipedia.org/wiki/Weighted_arithmetic_mean#Weighted_sample_variance) for an overview. But note that the section on reliability type weights might not be correct, see also this [Stackoverflow answer](https://stats.stackexchange.com/a/61641/209204).

lukashergt:
> I haven't yet had chance to play further with `WeightedGroupBy`, from Lukas' comment above it sounds like there might be quite a lot which I still need to implement. (e.g. I haven't got around to understanding what `WeightedGroupBy.sample` actually is...)

Maybe we don't need the `sample` method? The only reason `sample` came up is because it pops up as a reference in the docstrings of another method... But maybe it would be more manageable for us to only implement methods when we need them?

tests/test_samples.py
```diff
@@ -1339,3 +1339,38 @@ def test_old_gui():
         make_2d_axes(['x0', 'y0'], tex={'x0': '$x_0$', 'y0': '$y_0$'})
     with pytest.raises(NotImplementedError):
         make_1d_axes(['x0', 'y0'], tex={'x0': '$x_0$', 'y0': '$y_0$'})
+
+
+def test_groupby_stats():
+    mcmc = read_chains('./tests/example_data/cb')
+    chains = mcmc.groupby('chain')
```
lukashergt, line 1346: > Getting performance warnings /Users/adam/programming/anesthetic/anesthetic/weighted_pandas.py:27: PerformanceWarning: dropping on a non-lexsorted multi-index without a level parameter may impact performance.

This above line is the culprit, which got changed in https://github.com/williamjameshandley/anesthetic/pull/272/commits/c49536210ac8b3c74f99b548802d73737e04bb1a from:
```python
chains = mcmc.groupby(('chain', '$n_\\mathrm{chain}$'), group_keys=False)
```
Since we have a MultiIndex, we will get the warning when we only use the column name `'chain'` without its label. How do we prefer to handle this? It is good to know that things will work even when calling with only `'chain'`. But it's also nice to keep our test feedback clean. Shall we keep things as is and turn off the check for a PerformanceWarning for this line? Or revert to grouping with a joint call of column name and label?

tests/test_samples.py
```diff
@@ -1339,3 +1339,38 @@ def test_old_gui():
...
+    for chain in [1, 2]:
+        i = mcmc.chain == chain
+        assert_allclose(mcmc.loc[i].mean().drop('chain'),
+                        chains.mean().loc[chain, :])
+        assert_allclose(mcmc.loc[i].std().drop('chain'),
+                        chains.std().loc[chain, :])
+        assert_allclose(mcmc.loc[i].median().drop('chain'),
+                        chains.median().loc[chain, :])
+        assert_allclose(mcmc.loc[i].var().drop('chain'),
+                        chains.var().loc[chain, :])
```
lukashergt, line 1356: I think rather than using the `.loc[chain, :]` syntax it would be nicer to use the `chains.get_group(chain)` method, such that that gets tested as well.

tests/test_samples.py
```diff
@@ -1339,3 +1339,38 @@ def test_old_gui():
...
+
+    assert_allclose(mcmc.mean().drop('chain'), chains.mean().mean())
```
lukashergt, line 1358: Looks like this works, cool! :)

tests/test_samples.py
```diff
@@ -1339,3 +1339,38 @@ def test_old_gui():
...
+            for chain in [1, 2]:
+                i = mcmc.chain == chain
+                assert_allclose(mcmc.loc[i, col].mean(),
+                                chains[[col]].mean().loc[chain, :])
+                assert_allclose(mcmc.loc[i, col].std(),
+                                chains[[col]].std().loc[chain, :])
+                assert_allclose(mcmc.loc[i, col].median(),
+                                chains[[col]].median().loc[chain, :])
+                assert_allclose(mcmc.loc[i, col].var(),
+                                chains[[col]].var().loc[chain, :])
```
lukashergt, line 1371: As above, I think it would be nicer to use `chains.get_group(chain)`.

anesthetic/weighted_pandas.py
```diff
@@ -1,14 +1,112 @@
...
+    """Weighted version of ``pandas.core.groupby.DataFrameGroupBy``.
+
+    :meta private:
+    """
+
+    def get_weights(self):
+        """Return the weights of the grouped samples."""
+        return super().get_weights().min(axis=1-self.axis)
+
+    def _gotitem(self, key, ndim: int, subset=None): # pragma: no cover
```
lukashergt, line 73: Lint: Looks like there is a space missing before the inline comment here.

tests/test_samples.py
Ormorod: I'd prefer a joint call of column name and label, as warning suppression is quite messy

tests/test_samples.py
Ormorod: `chains.get_group(chain)` returns a `WeightedDataFrame`, so this change wouldn't actually be testing `WeightedDataFrameGroupBy`. Can additionally test this however

Ormorod: Additional tests added

lukashergt: hmm, I see what you are saying... but yes, we should have a test for `get_group` somewhere, too.

tests/test_samples.py
lukashergt: Can we get the same to work with `var`? (Might not be possible...)

lukashergt: Nope, I don't think that makes actually sense...

Ormorod: I *think* that what @williamjameshandley has done with setting weights in all of mean, var etc means that mean of means, variance of variances will be correctly calculated taking weights into consideration. 

I think you're right that the mean is somewhat special in that the mean of means is the same as the overall mean, but there might be similar tests we can construct for the others

tests/test_samples.py
williamjameshandley: resolved by @Ormorod in [5725ddf](https://github.com/williamjameshandley/anesthetic/pull/272/commits/5725ddfceb6d285c547cdbec8ed49f8596d2a9f8)

tests/test_samples.py
williamjameshandley: resolved by @Ormorod in [5725ddf](https://github.com/williamjameshandley/anesthetic/pull/272/commits/5725ddfceb6d285c547cdbec8ed49f8596d2a9f8)

tests/test_samples.py
williamjameshandley: Happy to go with @Ormorod's suggestion.

williamjameshandley:
>Maybe we don't need the sample method? The only reason sample came up is because it pops up as a reference in the docstrings of another method... But maybe it would be more manageable for us to only implement methods when we need them?

I actually implemented the sample method already in my last updates. This was quite simple, a `sample` already has a weights option, and all this does is hijack that. At the moment this draws one sample from each group in proportion to the weights within each group.

To match the other implementations in anesthetic, we're currently missing:

- [ ] `kurtosis`
- [ ] `kurt`
- [ ] `cov`
- [ ] `corr`
- [ ] `skew`
- [ ] `mad`
- [ ] `sem`
- [ ] `quantile`
- [ ] `corrwith`

tests/test_samples.py
williamjameshandley: The only one this works for (as far as I can see) is the mean of means. The mean of variances/variance of variances fundamentally fails because they correctly use the cluster means to compute their separate variances. There may be use-cases for this distinction in @Ormorod's clustering work.

Ormorod:
The remaining methods are a bit different, as they all require some kind of `other`.

Ormorod:
`corrwith()` doesn't exist for `Series` so I have left that off from the second set of tests

tests/test_samples.py
```diff
@@ -1339,3 +1339,126 @@ def test_old_gui():
...
+                                chains[[col]].kurtosis().loc[chain, :])
+                assert_allclose(mcmc.loc[i, col].skew(),
+                                chains[[col]].skew().loc[chain, :])
+                assert_allclose(mcmc.loc[i, col].mad(),
+                                chains[[col]].mad().loc[chain, :])
+                assert_allclose(mcmc.loc[i, col].sem(),
+                                chains[[col]].sem().loc[chain, :])
+                assert_allclose(mcmc.loc[i, col].cov(mcmc.loc[i, col]),
+                                chains[[col]].cov().loc[chain, :])
+                assert_allclose(mcmc.loc[i, col].corr(mcmc.loc[i, col]),
```
Ormorod, line 1448: This line is causing a warning

```shell
tests/test_samples.py::test_groupby_stats
  /Users/adam/programming/anesthetic/anesthetic/weighted_pandas.py:266: RuntimeWarning: invalid value encountered in scalar divide
    return self.cov(other, skipna=skipna)/norm
```
Any ideas?

lukashergt: Could this come from incorrect/missing dropping of the `'chain'` column?

tests/test_samples.py
```diff
@@ -1339,3 +1339,140 @@ def test_old_gui():
         make_2d_axes(['x0', 'y0'], tex={'x0': '$x_0$', 'y0': '$y_0$'})
     with pytest.raises(NotImplementedError):
         make_1d_axes(['x0', 'y0'], tex={'x0': '$x_0$', 'y0': '$y_0$'})
+
+
+def test_groupby_stats():
+    mcmc = read_chains('./tests/example_data/cb')
+    chains = mcmc.groupby(('chain', '$n_\\mathrm{chain}$'))
```
lukashergt, line 1346: I think it might be better to set things up as follows:
```python
mcmc = read_chains('/home/lthergt/Documents/Projects/AnestheticPrj/anesthetic/tests/example_data/cb')
params = ['x0', 'x1']
chains = mcmc[params + ['chain']].groupby(('chain', '$n_\\mathrm{chain}$'))
```
and then in the following to only use `params` instead of all columns, since some columns are actually constant which can cause nans and warnings.

For example instead of `mcmc.loc[mask]` we would write `mcmc.loc[mask, params]` and get around of a lot of `'chain'` dropping.

This will also be closer to the actual use-cases.

tests/test_samples.py
lukashergt: Or possibly from the constant 'minuslogprior` column?

tests/test_samples.py
```diff
@@ -1339,3 +1339,140 @@ def test_old_gui():
...
+    assert chains.median().isweighted() is True
+    assert chains.var().isweighted() is True
+    assert chains.kurt().isweighted() is True
+    assert chains.kurtosis().isweighted() is True
+    assert chains.skew().isweighted() is True
+    assert chains.mad().isweighted() is True
+    assert chains.sem().isweighted() is True
+    assert chains.corr().isweighted() is True
+    assert chains.cov().isweighted() is True
+    assert chains.corrwith(mcmc).isweighted() is True
```
lukashergt, line 1359: I've added these tests that ensure that our results from groupby methods are weighted. 

As you can see from https://github.com/williamjameshandley/anesthetic/pull/272/commits/142740cff80e44ca74590ccb6f474515dfe80583 the tests currently pass for `mean`, `std`, `median`, `var`, `kurt`, `kurtosis`, `skew`, and `sem`.

https://github.com/williamjameshandley/anesthetic/pull/272/commits/7b0a8e1aa233582d6d93679c378058154b04cba9 shows that `mad`, `corr`, `cov`, and `corrwith` are still failing, so I've unchecked them on our to-do list.

While the tests further down show that `cov` and Co. produce correct results within their respective group/chain, the tests don't show that the groups/chains are not correctly weighted with respect to one another.

tests/test_samples.py
```diff
@@ -1339,3 +1339,150 @@ def test_old_gui():
...
+
+    w1 = mcmc.loc[mcmc.chain == 1].get_weights().sum()
+    w2 = mcmc.loc[mcmc.chain == 2].get_weights().sum()
+    assert np.all(chains.mean().get_weights() == [w1, w2])
+    assert np.all(chains.std().get_weights() == [w1, w2])
+    assert np.all(chains.median().get_weights() == [w1, w2])
+    assert np.all(chains.var().get_weights() == [w1, w2])
+    assert np.all(chains.kurt().get_weights() == [w1, w2])
+    assert np.all(chains.kurtosis().get_weights() == [w1, w2])
+    assert np.all(chains.skew().get_weights() == [w1, w2])
```
lukashergt, line 1369: Some more tests that explicitly check that the groupby methods return the correct weights.

We will want to add tests for `cov` and `corr`, too, but this might be simplified by specifying parameters as suggested in [my comment above](https://github.com/williamjameshandley/anesthetic/pull/272/files#r1149789311).

tests/test_samples.py
williamjameshandley: Agreed, and done in 706d759

tests/test_samples.py
williamjameshandley: Resolved by 706d759

williamjameshandley:
OK, I think this resolves all of the above. @lukashergt could you have a look at the sphinx failure?

As all of us have contributed to this it is less clear-cut who should squash and merge, but given that @Ormorod submitted the original PR, I think he has a good claim. If @lukashergt could review & approve if he agrees, then I think we're done.

Ormorod:
e.g. the histograms:
```python
>>> mcmc = ac.read_chains("tests/example_data/cb")
>>> mcmc[['x0', 'x1', 'chain']].drop_labels().groupby('chain').hist()
chain  weights
1      5446.0     [[Axes(0.1...
2      5843.0     [[Axes(0.1...
dtype: object
>>> plt.show()
```
Screenshot 2023-03-30 at 12 15 27

```python
>>> mcmc[['x0', 'x1', 'chain']].drop_labels().groupby('chain').plot.hist()
hello from anesthetic HistPlot._make_plot
hello from anesthetic HistPlot._make_plot
chain
1    Axes(0.125...
2    Axes(0.125...
dtype: object
>>> plt.show()
```
Screenshot 2023-03-30 at 12 18 36

williamjameshandley:
>Aha! It turns out .groupby().plot.box() uses our anesthetic...BoxPlot, while .groupby().boxplot() uses the PANDAS boxplot. I've used print statments to test BoxPlot and HistPlot, and it would seem likely that the whole suite suffer from this

Nice spot! To fix this, you will need the GroupBy equivalent of this line:
```python
anesthetic.weighted_pandas._WeightedObject.plot =\
     CachedAccessor("plot", PlotAccessor)
```
in `anesthetic/samples.py`

Ormorod:
Looking at the pandas source, it points me to the similar `_make_wrapper()` area of `cov` etc, so the fix may be similar

Ormorod:
So I don't forget, the `_make_wrapper()` pattern works at the moment, but I think this will change with pandas2, as they start using `_op_via_apply()`

Ormorod:
I now believe that WeightedGroupBy is doing what we want it to, instead, the issue is that 
```python
mcmc.hist('x0')
```
Screenshot 2023-03-30 at 14 10 43

is different to 
```python
mcmc['x0'].plot.hist()
```
Screenshot 2023-03-30 at 14 12 11

essentially, skipping the area #235 was concerned with. However, I now think that the plots are, at the very least, using weights:
```python
fig, ax = plt.subplots()
mcmc.hist('x0', ax=ax, label='weighted')
mcmc.drop_weights().hist('x0', ax=ax, label='unweighted')
ax.legend()
```
Screenshot 2023-03-30 at 14 19 52

Ormorod:
Even in native pandas they have different formatting
```python
df = pd.DataFrame({'Animal': ['Falcon', 'Falcon',
                              'Parrot', 'Parrot'],
                   'Max Speed': [380., 370., 24., 26.]})
df.hist('Max Speed')
```
Screenshot 2023-03-30 at 14 29 49

```python
df['Max Speed'].plot.hist()
```
Screenshot 2023-03-30 at 14 34 38

Ormorod:
I guess I should probably test all the plot types...

anesthetic/weighted_pandas.py
```diff
@@ -1,12 +1,131 @@
...
+        return self._add_weights("median", *args, **kwargs)
+
+    def var(self, *args, **kwargs):  # noqa: D102
+        return self._add_weights("var", *args, **kwargs)
+
+    def kurt(self, *args, **kwargs):  # noqa: D102
+        return self._add_weights("kurt", *args, **kwargs)
+
+    def kurtosis(self, *args, **kwargs):  # noqa: D102
+        return self._add_weights("kurtosis", *args, **kwargs)
```
lukashergt, line 48: Why do we need to define `kurt` and `kurtosis` here, but don't need to define `skew`? Yet `skew` still passes the tests...? 

I first thought this had to do with pandas `GroupBy` not overriding `skew` as opposed to `mean`, `median`, `std` and `var`. However, the same would go for `kurt` and `kurtosis`, so now I wonder what is different/special about `skew`?

AdamOrmondroyd: fyi I've found that `skew` needs adding for pandas 2 (otherwise it fails the is weighted test), so could add it in here now

williamjameshandley: A little bit of an odd one -- pandas 1.5 doesn't actually implement groupby kurt and kurtosis (despite the fact I believe it could). @AdamOrmondroyd does it need adding for pandas 2? My instinct is to leave it in, as it doesn't do any harm.

williamjameshandley:
Is there anything else which needs to be done?

AdamOrmondroyd:
> Is there anything else which needs to be done?

I was working my way through testing all the plots, but so far I haven't found any issues with the plots themselves

AdamOrmondroyd:
Had a fruitless look into why it's failing on pip but not conda, isn't as simple as numpy 1.23 vs 1.24

lukashergt:
> Had a fruitless look into why it's failing on pip but not conda, isn't as simple as numpy 1.23 vs 1.24

I think it was a numerical/floating point thing where the covariance matrix turned out positive definite despite having a linear dependent parameter. Anyhow, more helped more (https://github.com/williamjameshandley/anesthetic/pull/272/commits/85fa6aefb57906f7226a523d3e078fb153a49ec1)...

AdamOrmondroyd:
@lukashergt if you're happy with `cov`/`GelmanRubin` then I reckon we're good to go

lukashergt (approved):
Ok, with the Gelman--Rubin statistic fixed thanks to the functioning groupby I am happy now. If you are happy too, feel free to squash and merge. 

Thanks, @AdamOrmondroyd, @williamjameshandley

PR #278, opened 2023-04-06
 Implemented Samples.credibility_interval 2.0 (#188 adapted for anesthetic 2.0)

Stefan-Heimersheim:
# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context.

Fixes # (issue)

# Checklist:

# Tests

Test correct result for Gaussian:
```
means = [[], []]
    stds = [[], []]
    for i in range(1000):
        samples = np.random.normal(loc=2, scale=0.1, size=10000)
        mean, std = credibility_interval(samples, level=0.68)
        means[0].append(mean[0])
        means[1].append(mean[1])
        stds[0].append(std[0])
        stds[1].append(std[1])
    plt.hist(means[0], bins=100, histtype='step', label='lower')
    plt.legend()
    plt.show()
    print("Mean = ", np.mean(means[0]), "+/-", np.std(means[0]))
    print("Std = ", np.mean(stds[0])
```

# Questions

I considered adding a test iff user passes a `NestedSamples` or `MCMCSamples` object then check that they passed the right weights. This would introduce the first anesthetic import into `utils.py` so wanted to double check.
```
# Check if samples is NestedSamples object
if type(samples) == NestedSamples:
    if weights is None:
        assert samples.get_weights() == np.ones(len(samples))
    else:
        assert np.allclose(weights, samples.get_weights())
```

# Notes

Will potentially make a separate PR to implement this method into existing `quantile()` function
```
def cdf(a, w=None, inverse=False, interpolation='linear'):
    """Compute the weighted (inverse) empirical cdf for a 1d array."""
    if w is None:
        w = np.ones_like(a)
    a = np.array(list(a))  # Necessary to convert pandas arrays
    w = np.array(list(w))  # Necessary to convert pandas arrays
    i = np.argsort(a)
    c = np.cumsum(w[i[1:]]+w[i[:-1]])
    c = c / c[-1]
    c = np.concatenate(([0.], c))
    if inverse:
        return interp1d(c, a[i], kind=interpolation)
    else:
        return interp1d(a[i], c, kind=interpolation)


def quantile(a, q, w=None, interpolation='linear'):
    """Compute the weighted quantile for a one dimensional array."""
    icdf = cdf(a, w, inverse=True, interpolation=interpolation)
    quant = icdf(q)
    if isinstance(q, float):
        quant = float(quant)
    return quant
```

```

def test_cdf():
    np.random.seed(3)
    data = np.random.randn(10000)
    weights = np.random.rand(10000)

    CDF = cdf(data, weights)
    ICDF = cdf(data, weights, inverse=True)

    linspace = np.linspace(0, 1, 1001)[1:-1]
    assert_array_almost_equal(linspace, CDF(ICDF(linspace)))

    sigs = [CDF(i)-CDF(-i) for i in [1, 2, 3]]
    assert_allclose(sigs, [0.6827, 0.9545, 0.9973], atol=0.01)
```

### DIFF
--- a/anesthetic/utils.py
+++ b/anesthetic/utils.py
@@ -72,31 +72,20 @@ def compress_weights(w, u=None, ncompress=True):
     extra = (u < fraction).astype(int)
     return (integer + extra).astype(int)
 
-def cdf(a, w=None, inverse=False, interpolation='linear'):
-    """Compute the weighted (inverse) empirical cdf for a 1d array."""
-    if w is None:
-        w = np.ones_like(a)
-    a = np.array(list(a))  # Necessary to convert pandas arrays
-    w = np.array(list(w))  # Necessary to convert pandas arrays
-    i = np.argsort(a)
-    c = np.cumsum(w[i[1:]]+w[i[:-1]])
-    c = c / c[-1]
-    c = np.concatenate(([0.], c))
-    if inverse:
-        return interp1d(c, a[i], kind=interpolation)
-    else:
-        return interp1d(a[i], c, kind=interpolation)
-
-def sample_cdf(samples, interpolation='linear'):
+def sample_cdf(samples, inverse=False, interpolation='linear'):
     """Sample the empirical cdf for a 1d array."""
     samples = np.array(samples)
     samples.sort()
     ngaps = len(samples)-1
     gaps = np.random.dirichlet(np.ones(ngaps))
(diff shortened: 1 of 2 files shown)


Stefan-Heimersheim:
Nevermind, encouraged by @williamjameshandley to embrace the git and have updated the old PR in #188 :)

PR #281, opened 2023-04-09
rename `[test]` optional dependency to `[tests]`

AdamOrmondroyd:
# Description

Quick one, I find it quite strange we have plural `extras` but singular `test` optional dependencies, so I've changed it to `tests`

Fixes # (issue)

# Checklist:

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -88,7 +88,7 @@ You can check that things are working by running the test suite:
 .. code:: bash
 
     export MPLBACKEND=Agg     # only necessary for OSX users
-    python -m pip install ".[test]"
+    python -m pip install ".[tests]"
     python -m pytest
     flake8 anesthetic tests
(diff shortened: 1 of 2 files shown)


williamjameshandley (changes requested):
Consistency is important -- I guess the question is whether this should be 'test, extra' or 'tests, extras'. Going by [pandas](https://github.com/pandas-dev/pandas/blob/e9e034badbd2e36379958b93b611bf666c7ab360/pyproject.toml#L59), we should use `test`, and perhaps `all` rather than `extras`

AdamOrmondroyd:
Ah, I'd misremembered that pandas used `[tests]`. Now swapped around to use `[all]`

williamjameshandley (approved):

williamjameshandley (approved):

PR #284, opened 2023-04-20
Extending channel_capacity() function

htjb:
# Description

Updated the `channel_capacity()` function to `effective_samples()` and added a `method` option which is defaulted to `channel' but allows users to calculate a Kish estimate of the effective number of samples given by

n_{eff} = \frac{(\sum_i w_i)^2}{\sum_i w_i^2}.

This was briefly mentioned in #280.

If anybody has a suggestion on how to test this bit of code that would be super helpful @williamjameshandley @AdamOrmondroyd @lukashergt?

# Checklist:


lukashergt:
Hi @htjb, 

thanks for initiating this. I see that you are pushing from your master branch, are you sure that you want to do that? Normally you would create a new branch dedicated to the change that you want to make, and only merge that into master once all the changes are fully addressed.


> If anybody has a suggestion on how to test this bit of code that would be super helpful @williamjameshandley @AdamOrmondroyd @lukashergt?

How about:

```python
import numpy as np
from pytest import approx
from anesthetic.utils import effective_samples

@pytest.mark.parametrize('method', ['channel', 'kish'])
def test_effective_samples(method):
    w = np.ones(20)
    assert effective_samples(w, method=method) == approx(20, rel=1e-6)
    w[10:] = 1e-3
    assert effective_samples(w, method=method) == approx(10, rel=1e-2)
```

Would we want a `skipna` kwarg for this?

htjb:
@lukashergt My bad, I was rushing to do this between meetings! I've moved my code onto another branch and will update the tests and make another PR from there. Thanks for the test suggestion!

PR #285, opened 2023-04-21
Effective samples

htjb:
# Description

Updating #284 to PR from branch rather than my master branch. Included test suggested by @lukashergt and updated failing tests where necessary. Copied the original PR description below.

Updated the `channel_capacity()` function to `effective_samples()` and added a `method` option which is defaulted to `channel' but allows users to calculate a Kish estimate of the effective number of samples given by

n_{eff} = \frac{(\sum_i w_i)^2}{\sum_i w_i^2}.

This was briefly mentioned in #280.

If anybody has a suggestion on how to test this bit of code that would be super helpful @williamjameshandley @AdamOrmondroyd @lukashergt?

# Checklist:


lukashergt:
How about going with this more general formulation dependent on a parameter `beta`?

```python
def effective_samples(w, beta=1):
    w = w / w.sum()
    if beta == 1:
        return np.exp(entropy(w))
    elif beta == np.inf:
        return 1 / np.max(w)
    else:
        return np.sum(w**beta)**(1/(1-beta))
```

For a nice overview, see: https://aakinshin.net/posts/huggins-roy-ess.

The channel capacity (or perplexity) method corresponds to `beta=1` and as such is a limiting case and hence singled out because of a division by zero error.

The Kish method corresponds to `beta=2`. 

`beta=np.inf` is singled out for the obvious reason that `np.inf` does not behave well and does not give the correct limit, here.

This formulation in terms of `beta` is nice in that it allows us to order the estimates: 
* For `beta1 < beta2` we have `Neff(beta1) > Neff(beta2)`. 
* `beta=1/2` gives a slightly larger `Neff` for when we want to be conservative. 
* `beta=np.inf` gives a smaller `Neff` for when we feel radical. 
* `beta=1` and `beta=2` are the good in-betweens.

htjb:
Sure I can have a look at this @williamjameshandley @lukashergt👍

htjb:
Hi @lukashergt @williamjameshandley, I updated the code to use the Hugins-Roy ESS as Lukas suggested. I changed beta to gamma to avoid confusion with thermodynamic beta (presuming they are different right?). It looked like one of the tests was failing locally too but I couldn't see how the changes I have made affected it.

AdamOrmondroyd:
>  It looked like one of the tests was failing locally too but I couldn't see how the changes I have made affected it.

Is it test_samples.py::test_mcmc_stats() which is failing to raise a `numpy.linalg.LinAlgError`? I am getting this locally too.

htjb:
@AdamOrmondroyd yep that's the one. Looks like it passed here though? Maybe a numpy version issue?

AdamOrmondroyd:
@htjb it also passes on CSD3, which is why I approved the PR which introduced this test (I was away from my laptop so CSD3 on my phone!). I've already had a play with numpy versions, and didn't come up with anything.

- it just occurred to me we're both mac users...

tests/test_utils.py
```diff
@@ -172,3 +174,11 @@ def test_p_values_from_sample():
...
     ks_results = insertion_p_value(ns.insertion[nlive:-nlive], nlive, batch=1)
     assert ks_results['p-value'] > 0.05
+
+
+@pytest.mark.parametrize('method', ['channel', 'kish'])
+def test_effective_samples(method):
+    w = np.ones(20)
+    assert effective_samples(w, method=method) == approx(20, rel=1e-6)
+    w[10:] = 1e-3
+    assert effective_samples(w, method=method) == approx(10, rel=1e-2)
```
lukashergt, line 184: Note, `gamma=0.5` leads to higher `n_eff` and will require either further reducing the `1e-3` and/or relaxing the relative error.

lukashergt, line 184: We should also append the test and check for what behaviour we expect in case there are zeros in `w`, e.g. follow up with the case:
```python
    w[15:] = 0
    ...
```
This would produce warnings and nans in `np.log`, but using scipy's `entropy` resolves that.

lukashergt (changes requested):
Thanks @htjb, some further suggestions inline.

Thanks @AdamOrmondroyd for #287 which will hopefully take care of the testing difficulties that mac users were experiencing.

anesthetic/utils.py
```diff
@@ -32,21 +32,66 @@ def logsumexp(a, axis=None, b=None, keepdims=False, return_sign=False):
...
+    if method == 'kish':
+        gamma = 2
+    elif method == 'channel':
+        gamma = 1
+
+    w = w / np.sum(w)
+    if gamma == 1:
+        with np.errstate(divide='ignore', invalid='ignore'):
+            H = np.nansum(np.log(w)*w)
+            return np.exp(-H)
```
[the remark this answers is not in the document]
htjb: When I use the `entropy` function the following test in `test_utils.py` fails

```python

def test_sample_compression_1d():
    np.random.seed(0)
    N = 10000
    x_ = np.random.rand(N)
    w_ = np.random.rand(N)
    n = 1000
    x, w = sample_compression_1d(x_, w_, n)
    assert len(x) == n
    assert len(w) == n
    assert np.isclose(w.sum(), w_.sum())

    x_ = np.random.rand(N)
    w_ = np.random.randn(N)
    x, w = sample_compression_1d(x_, w_, ncompress=True)
    assert len(w) <= N

```
Specifically, this bit

```python
x_ = np.random.rand(N)
w_ = np.random.randn(N)
x, w = sample_compression_1d(x_, w_, ncompress=True)
assert len(w) <= N
```
because the `entropy(w)` is evaluated as `-inf` giving 0.0 effective samples and throwing the following error 

```TypeError: expected a sequence of integers or a single integer, got '0.0'```

in the `np.choice` function in `sample_compression_1d()` in `utils.py`.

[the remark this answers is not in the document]
Comment (lukashergt) [reply anesthetic/utils.py:81]: Looks like we've got a bug in `test_sample_compression_1d`. 

The line
```python
    w_ = np.random.randn(N)
```
produces negative weights, but our weights (although not necessarily normalised) should be strictly positive...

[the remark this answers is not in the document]
htjb: How about replacing it with

`w_ = np.random.lognormal(1, 0.5, N)`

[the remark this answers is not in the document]
htjb: Hmm I get a similar error

`TypeError: expected a sequence of integers or a single integer, got '8828.14997708517'`

Should `ncompress` be an integer?

tests/test_utils.py
```diff
@@ -89,15 +91,15 @@ def test_sample_compression_1d():
     np.random.seed(0)
     N = 10000
     x_ = np.random.rand(N)
-    w_ = np.random.rand(N)
+    w_ = np.random.randn(N)
```
lukashergt, line 94: Why the normally distributed weights, here? This will produce negative weights...

anesthetic/utils.py
```diff
@@ -378,15 +414,15 @@ def sample_compression_1d(x, w=None, ncompress=True):
     if ncompress is False:
         return x, w
     elif ncompress is True:
-        ncompress = channel_capacity(w)
+        ncompress = effective_samples(w)
```
lukashergt, line 417: We should propagate the `gamma` kwarg to the various functions and methods that call `effective_samples`.

anesthetic/utils.py
```diff
@@ -378,15 +414,15 @@ def sample_compression_1d(x, w=None, ncompress=True):
...
     x = np.array(x)
     if w is None:
         w = np.ones_like(x)
     w = np.array(w)
 
     # Select inner samples for triangulation
     if len(x) > ncompress:
-        x_ = np.random.choice(x, size=ncompress, replace=False)
+        x_ = np.random.choice(x, size=int(ncompress), replace=False)
```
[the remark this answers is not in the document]
htjb: I'm not sure... the `np.random.choice` definitely needs an integer input...

tests/test_utils.py
htjb: I'll change it to the same log normal distribution used later on in the function.

htjb:
@lukashergt How is this looking?

tests/test_utils.py
lukashergt: Why change it at all? I'd rather keep tests as is, unless there is a bug...

anesthetic/utils.py
```diff
@@ -58,7 +94,7 @@ def compress_weights(w, u=None, ncompress=True):
         w = np.ones_like(u)
 
     if ncompress is True:
-        ncompress = channel_capacity(w)
+        ncompress = effective_samples(w, gamma=gamma)
     elif ncompress is False:
         return w
```
lukashergt, line 99: Hmm, do you reckon there could be a way of combining `ncompress` and `gamma` into one? 

E.g.
```python
if ncompress is True or isinstance(ncompress, str):
    ncompress = effective_samples(w, gamma=ncompress)
```

Then `effective_samples` would need to map `gamma=True` to `gamma=1`. Maybe we can expand the `gamma` string options to also work for `'1'`, `'2'` etc.?
```python
elif isinstance(gamma, str):
    gamma = float(gamma)
```

Or am I overcomplicating this? 

The thing is, `gamma` (or `ncompress` in its place) should be propagated all the way to `plot.py`. However, in `plot.py` a kwarg `gamma` will be very opaque, whereas for the already existing kwarg `ncompress` it is clearer what it is for and it essentially plays the same role as `gamma` in determining the number of samples after compression...

anesthetic/utils.py
```diff
@@ -58,7 +94,7 @@ def compress_weights(w, u=None, ncompress=True):
         w = np.ones_like(u)
 
     if ncompress is True:
-        ncompress = channel_capacity(w)
+        ncompress = effective_samples(w)
```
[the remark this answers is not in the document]
lukashergt: Also the functions that then call `compress_weights` all the way up. But see also my comment of possibly combining `ncompress` and `gamma`.

lukashergt:
> @lukashergt How is this looking?

Sorry, didn't realise this was ready for review. Feel free to press `Re-request review` when you are done with modifications and would like me to take a look. I'll try to be faster in my reaction.

anesthetic/utils.py
williamjameshandley: I think this could work, but the difference between `ncompress='2'` and `ncompress=2` is equally hard to parse. Perhaps calling it `strategy` or some synonym for compression strategy in this function, and let the real experts pass through `ncompress='0.5'` at the `plot.py` level.

I dislike `gamma`, and would prefer `beta` (since it is the same flavour as a thermodynamic beta).

williamjameshandley: In this instance I was suggesting changing `ncompress` to `strategy` in the `compress_weights` function, but leaving it as `ncompress` higher up.

Another convention would be using negative values for ncompress instead of strings.

This would however be a substantial breaking change, in that at the moment negative values indicate equal weighting (this could swap to ncompress=True, which currently codes for compressing to the channel capacity).

lukashergt: > In this instance I was suggesting changing `ncompress` to `strategy` in the `compress_weights` function, but leaving it as `ncompress` higher up.

Ah ok, so `strategy` instead of `ncompress` to be more descriptive. That works for me, I just want to avoid two potentially conflicting kwargs.

> Another convention would be using negative values for ncompress instead of strings.

What I like about the string option is that it fits to the two string options which we already have, `beta='entropy'` and `beta='kish'`.

> This would however be a substantial breaking change, in that at the moment negative values indicate equal weighting (this could swap to ncompress=True, which currently codes for compressing to the channel capacity).

`samples.compress(ncompress='entropy')` will also result in equally weighted samples, right? Only for `ncompress=-1` it is ensured that each sample point will be unique, right?

tests/test_utils.py
htjb: I changed it to `w_ = np.random.lognormal(1, 0.5, N)` since `np.random.randn` will give you negative weights and these are unphysical, right? Would you ever draw a sample with a negative weight?

anesthetic/utils.py
htjb: I'm a bit unclear on the outcome of this discussion @williamjameshandley @lukashergt? 

You want `ncompress` changed to `strategy` in `compress_weights` function but `ncompress` left as `ncompress` everywhere else. But then you also want to replace `gamma` with `ncompress` and for `ncompress` to take on numerical positive values and also 'kish', 'entropy' and things like '0.5' etc?

anesthetic/utils.py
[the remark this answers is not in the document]
htjb: Same as above.

tests/test_utils.py
lukashergt: But why change it to `randn` in the first place? Why not leave it as `rand`?

anesthetic/utils.py
lukashergt: Yes, apologies, that was probably a bit difficult to follow. I have pushed the changes that I had in mind in commits https://github.com/handley-lab/anesthetic/pull/285/commits/1c71f772fea0e2325626c359d19549ea71f3c49f through https://github.com/handley-lab/anesthetic/pull/285/commits/c6fc942c544511cf215484804434b919b1047f86. Would be good for someone to have a look through that to make sure things make sense.

I've left the `ncompress` kwarg in `compress_weights` as is for now. I don't actually expect `compress_weights` to be used much on its own. It is mostly there to be called from `WeightedDataFrame.compress` where the docstring for `ncompress` works well, I think. But feel free to change to `strategy` or `degree` if you still think that would be better.

lukashergt:
With this new `effective_samples` functionality, it might be worth thinking over some of our `ncompress` defaults.

Currently we use the following defaults:
* `ncompress=False` in `kde_plot_1d`: this is fine, since `nplot_1d` ensures that should be speedy enough.
* `ncompress=1000` in `kde_plot_2d`: I'm thinking that we could maybe change this to `ncompress='inf'`, which is the minimal number of effective samples according to the Huggins & Roy formula. Thus it is less arbitrary than 1000, has a better chance of not missing any relevant weights, while still trying to keep the number low for things to be speedy enough...
* `ncompress=True` in `scatter_plot_2d`: this is the same as `ncompress='entropy'='1'`. For the scatter plot this is probably more than we usually want, so here a change to `ncompress='inf'` might be good, too...

Thoughts?

williamjameshandley (changes requested):
> With this new `effective_samples` functionality, it might be worth thinking over some of our `ncompress` defaults.
> 
> Currently we use the following defaults:
> 
> * `ncompress=False` in `kde_plot_1d`: this is fine, since `nplot_1d` ensures that should be speedy enough.
> * `ncompress=1000` in `kde_plot_2d`: I'm thinking that we could maybe change this to `ncompress='inf'`, which is the minimal number of effective samples according to the Huggins & Roy formula. Thus it is less arbitrary than 1000, has a better chance of not missing any relevant weights, while still trying to keep the number low for things to be speedy enough...
> * `ncompress=True` in `scatter_plot_2d`: this is the same as `ncompress='entropy'='1'`. For the scatter plot this is probably more than we usually want, so here a change to `ncompress='inf'` might be good, too...
> 
> Thoughts?

I agree this is the time to address this, and those defaults seem very reasonable

`ncompress='inf'` is now equivalent to `ncompress=-1`. Could we deprecate the latter, which is a rather unintuitive interface?

anesthetic/utils.py
```diff
@@ -32,21 +32,61 @@ def logsumexp(a, axis=None, b=None, keepdims=False, return_sign=False):
...
+        .. math::
+
+            N = \frac{(\sum_i w_i)^2}{\sum_i w_i^2}
     """
-    with np.errstate(divide='ignore', invalid='ignore'):
-        W = np.array(w)/sum(w)
-        H = np.nansum(np.log(W)*W)
-        return np.exp(-H)
+    w = w / np.sum(w)
+    if beta == 1 or beta == '1' or beta == '1.0' or beta == 'entropy':
```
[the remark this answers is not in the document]
lukashergt: Changed to 
```python
    if beta == 'entropy' or beta != 'kish' and str(float(beta)) == '1.0':
```
in https://github.com/handley-lab/anesthetic/pull/285/commits/8b5ae3322e4e628a0d9b8e40d18270c9bbbded91. Thanks to lazy evaluation this actually works.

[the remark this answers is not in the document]
williamjameshandley: Nice

williamjameshandley:
OK, I've made a couple of changes. (a) `effective_samples`->`neff`. This is more consistent with the rest of the code, and a more informative name (effective_samples is actually incorrect, since it is the number of effective samples).
(b) I've changed `ncompress=-1` to be `ncompress='inf'` or `ncompress='equal'`.

With these changes I'm happy for this to be merged.

williamjameshandley (approved):

williamjameshandley (approved):

tests/test_utils.py
htjb: Hmm I can't remember why I changed it... I have rolled it back to the `rand` function.

lukashergt (approved):
> (b) I've changed `ncompress=-1` to be `ncompress='inf'` or `ncompress='equal'`.

It's really neat how this is now part of `neff`. 

I've implemented the default changes discussed above and simplified/cleaned up the docstring for `neff` a bit.

Happy for this to be squashed and merged now. Many thanks @htjb!

htjb:
Great, thanks @lukashergt and @williamjameshandley for the help with this one! 

@lukashergt It looks like some of the tests are failing? Maybe pandas related?

htjb:
@lukashergt I had a flick through the output for the failed conda test and compared it to the master branch. I couldn't see many differences but on the master branch conda installed pandas=1.5.3 whereas here it installed 2.0.2? Is this branch ahead/behind the master? Maybe the matplotlib version requirement is being overwritten by the requirement of a different package like pandas?

AdamOrmondroyd:
anesthetic now requires pandas>=2.0 (see #292), good spot @htjb. It is worrying that the master branch test used 1.5.3...

htjb:
> anesthetic now requires pandas>=2.0 (see #292), good spot @htjb. It is worrying that the master branch test used 1.5.3...

@AdamOrmondroyd its probably worth somebody double-checking I have read the output of the test properly!

AdamOrmondroyd:
I've just spotted that in the conda CI it installs pandas twice, once with scipy, numpy and matplotlib, then once again specifying >=2.0.0

htjb:
@AdamOrmondroyd Ahh yes I missed that! Perhaps pandas isn't the culprit here hahaa.

lukashergt:
>> @lukashergt I had a flick through the output for the failed conda test and compared it to the master branch. I couldn't see many differences but on the master branch conda installed pandas=1.5.3 whereas here it installed 2.0.2? Is this branch ahead/behind the master? Maybe the matplotlib version requirement is being overwritten by the requirement of a different package like pandas?
> 
> I've just spotted that in the conda CI it installs pandas twice, once with scipy, numpy and matplotlib, then once again specifying >=2.0.0

Yes, the last pandas install is the one that matters, which is forced to be >=2 for both master and PR tests: 

```yaml
      - name: Install dependencies
        shell: bash -l {0}
        run: |
          conda config --append channels conda-forge
          conda install pytest pytest-cov
          conda install scipy numpy pandas matplotlib
          conda install 'pandas>=2.0.0'
```

The curious difference between master and PR tests is that the line
```yaml
          conda install scipy numpy pandas matplotlib
```
installs different versions between master and PR tests. As far as I can tell the too low matplotlib version in the PR tests is the problem...

AdamOrmondroyd:
I'm checking what happens if pandas is removed from the first conda install in #295, hopefully it's as simple as that

appetrosyan:
@AdamOrmondroyd 

I have yet to find people that like conda, even among people who are actively working on it. In all fairness, the whole Python ecosystem has no better answer and that is an inherent limitation.

lukashergt:
> @lukashergt, with the changes to the default number of samples there is [one line](https://app.codecov.io/gh/handley-lab/anesthetic/pull/285/indirect-changes#7ebf6efb63f35c9c91ea4bda421bae01-R398) which is no-longer covered. We'll need an extra test with large compression to catch this.

Done in https://github.com/handley-lab/anesthetic/pull/285/commits/d790153fff4cef6baf6caecec8edff9fd7cf6811.

williamjameshandley (approved):

lukashergt (approved):
@htjb, looks like this is finally ready. Please feel free to squash and merge.

htjb:
Nice, thanks for the help! 😄

PR #287, opened 2023-05-02
Use the scipy version of `linalg.cholesky` which consistently raises `LinAngError` with linearly dependent columns

AdamOrmondroyd:
# Description

On MacOS at least, `Samples.Gelman_Rubin()` does not consistently raise `np.LinAlgError` when the samples have linearly dependent columns, as was pointed out by @htjb in #285 Using the scipy version appears to fix this

# Checklist:


lukashergt:
Sorry, this PR slipped me by.

> On MacOS at least, `Samples.Gelman_Rubin()` does not consistently raise `np.LinAlgError` when the samples have linearly dependent columns, as was pointed out by @htjb in #285 Using the scipy version appears to fix this

Yeah, the LinAlgError was always a bit finicky, which is why I ended up repeating the same line three times when one time should have been enough... Do the tests still pass when we reduce it to a single repeated line?

AdamOrmondroyd:
I remember having the opposite issues with the PolyChord macOS CI, where I could run things locally fine but not pass the same tests on github.

I've just tested pythons 3.8-11 locally (M1 Pro macbook pro, macOS 13.3) and this test fails on all of them, and all four are fixed with this numpy->scipy changed.

> @lukashergt It doesn't look like there is a macos test with python 3.8? Although I get the error locally with python3.10 on my mac and it passes on github.

This was a deliberate change a few months ago in #249. Tests on windows and macOS take much longer as they are running on virtual machines, so I think it was decided a range of python versions on linux and a single macOS and windows test would be sufficient.

lukashergt:
> Do the tests still pass when we reduce it to a single repeated line?

Well, that didn't work (see the failed tests after https://github.com/handley-lab/anesthetic/pull/287/commits/da96aea26c2c4a32a4423e131184467a289cd6bc)... That's annoying, now I'm less confident that this is truly resolved. If anybody has an idea how to reliably test for LinAlgErrors, let me know. For now, let's just leave it as is. If a similar issue pops up again, then maybe it's best to get rid of the entire try-except block...

I put the try-except block in there for confused people (aka my past self) that tried this with MCMC samples from a Gibbs chain, where some parameters didn't get updated as often as others, and I initially didn't question the samples but the G-R implementation...

lukashergt (approved):
Thanks @AdamOrmondroyd, please squash and merge.

AdamOrmondroyd:
> > Do the tests still pass when we reduce it to a single repeated line?
> 
> Well, that didn't work... It worked for me locally but as you can see the CI failed after [da96aea](https://github.com/handley-lab/anesthetic/commit/da96aea26c2c4a32a4423e131184467a289cd6bc)... That's annoying, now I'm less confident that this is truly resolved. If anybody has an idea how to reliably test for LinAlgErrors, let me know. For now, let's just leave it as is. If a similar issue pops up again, then maybe it's best to get rid of the entire try-except block...
> 
> I put the try-except block in there for confused people (aka my past self) that tried this with MCMC samples from a Gibbs chain, where some parameters didn't get updated as often as others, and I initially didn't question the samples but the G-R implementation...

Just checking we're on the same page: the test is working correctly, just `np.linalg.cholesky` successfully computes something and doesn't throw an error. I couldn't get to the bottom of why numpy does this

lukashergt:
Yes, numpy (and scipy) don't reliably produce a LinAlgError where they should. It appears that OS, python version, and platform all play a role in this. This should not affect the user, but is annoying when checking for code coverage, because it is difficult to reliably get the LinAlgError.

AdamOrmondroyd:
@lukashergt still happy for me to merge?

lukashergt:
> @lukashergt still happy for me to merge?

Seems to resolve things for now. If that's the end of it, that's great. My lengthy tangent was more for future reference.

So yes, please feel free to squash and merge. Many thanks.

PR #289, opened 2023-05-06
Make name of Colormaps returned by basic_cmap a string 

### LINKED ISSUE #288: `reversed` method broken for the `Colormap` returned by `basic_cmap`
**Summary**
If `plot.basic_cmap` is called with an input colour given as a tuple e.g. `(1.0, 1.0, 0.0)` the returned `LinearSegementedColormap` object's `reversed` method fails with a `TypeError`.

**To reproduce:**
Install anesthetic at version 2.0.0b28.

Run,
```
from anesthetic.plot import basic_cmap
cmap = basic_cmap((1.0, 1.0, 0.0, 1.0))
reversed_cmap = cmap.reversed()
```
Which raises on my system,
> Traceback (most recent call last):
>   File "minimal_example.py", line 22, in <module>
>     main()
>   File "minimal_example.py", line 14, in main
>     reversed_cmap = cmap.reversed()
>   File "/home/thomas/.virtualenvs/SuperconductingCosmicStringConstraints/lib/python3.8/site-packages/matplotlib/colors.py", line 1096, in reversed
>     name = self.name + "_r"
> TypeError: can only concatenate tuple (not "str") to tuple

**Expected behaviour**
Returns a `LinearSegmentedColormap` with the colors reversed with respect to `basic_cmap`.

**Additional context**
This error seems to stem from the setting of the default name of the new reversed `Colormap`
```
name = self.name + "_r"
```
which is derived from the name of the existing `Colormap`. Consulting the `matplotlib` documentation, the `Colormap` class expects the name to be a `str`. However, `basic_cmap` sets the `Colormap`'s name to the input `color`, and so if this is given in `tuple` format, as in the example, the `Colormap`'s name will be a `tuple`. This in turn causes a type error when the `reversed` method is called and it attempts to append `"_r"` to a `tuple`.

A workaround for this bug exists by manually setting the reversed `Colormap`'s name
```
reversed_cmap = cmap.reversed(name='reversed_cmap')
```

ThomasGesseyJones:
# Description

Sets the name of the `Colormap` created by `basic_cmap` to `str(color)` rather than the input `color`, so that the name is of the expected type by the `Colormap` class. This, fixes TypeErrors in methods that assume the `Colormap`'s name is a `str` such as `.reversed()`, when `color` is given as a `tuple`.

Fixes #288 

# Checklist:


tests/test_plot.py
```diff
@@ -726,3 +726,9 @@ def test_quantile_plot_interval_tuple(q1, q2):
     _q1, _q2 = quantile_plot_interval(q=(q1, q2))
     assert _q1 == q1
     assert _q2 == q2
+
+
+def test_basic_cmap_reversible():
+    # Check that the basic cmap is now reversible for tuple input
+    cmap = basic_cmap((1.0, 1.0, 0.0, 1.0))
+    cmap.reversed()
```
lukashergt, line 734: Since we don't actually have a dedicated test of `basic_cmap`, I suggest naming this more generally `test_basic_cmap` and testing more colour options. How about:
```python
@pytest.mark.parametrize('c', ['C0', 'k', 'gold', '#00FFFF', (1.0, 1.0, 0.0, 1.0)])
def test_basic_cmap_reversible(c):
    cmap = basic_cmap(c)

    # Check that the basic cmap is reversible
    cmap.reversed()
```

lukashergt (changes requested):
Hi Thomas, thanks for this PR. Looks all good to me. I've got one suggestion to expand the test to more colour options (see inline).

ThomasGesseyJones:
Hi @lukashergt, thanks for the suggestion, I have implemented it in the latest commit. Let me know if you think any more additions would be useful

lukashergt (dismissed):
Looks good to me, thanks @ThomasGesseyJones, please squash and merge.

ThomasGesseyJones:
I am afraid since I don't have write access to the repository I cannot squash and merge the changes

lukashergt:
> I am afraid since I don't have write access to the repository I cannot squash and merge the changes

Ah, and I am afraid I don't have the rights to grant you write access... 
@williamjameshandley could you add @ThomasGesseyJones, please?

williamjameshandley:
@ThomasGesseyJones has been sent an invite for write access, and I've now given @lukashergt admin access to be able to do this in the future!

lukashergt (approved):

PR #292, opened 2023-06-01
pandas2.0 take III: Revenge of the Sith

AdamOrmondroyd:
# Description

For the third, and hopefully final, time, here are the changes required for compatibility with pandas 2.

Edit: looks like conda still hasn't caught up :cry:

# Changes
- Remove redundant `HistPlot._make_plot()` as fixed on the pandas end
- Add `_PlanePlot2d._args_adjust()` as this is now a abstract method
- Remove `squeeze` from groupby constructors
- Add weighted version of groupby `_op_via_apply`, this has started being used in preference to `agg`
- Add `cov` to groupby
- Remove `mad` from groupby tests as `mad` been removed from pandas
- Pandas 2.0 requires matplotlib >= 3.6.1
- Add missing weighted groupby `skew`

Fixes # (issue)

# Checklist:


anesthetic/weighted_pandas.py
```diff
@@ -72,13 +75,25 @@ def wrapper(*args, **kwargs):
         wrapper.__name__ = name
         return wrapper
```
williamjameshandley, line 76: I believe that with `_op_via_apply` we no longer need _make_wrapper?` (it's not [covered](https://app.codecov.io/gh/handley-lab/anesthetic/commit/b68bb6f33a61450f25558300ca1e2095efd16cb2/blob/anesthetic/weighted_pandas.py#L64) at the moment)

williamjameshandley (changes requested):
Many thanks for putting this all together @AdamOrmondroyd.

williamjameshandley (changes requested):
Many thanks for putting this all together @AdamOrmondroyd.

pyproject.toml
```diff
@@ -25,8 +25,8 @@ requires-python = ">=3.8"
 dependencies = [
     "scipy",
     "numpy",
-    "pandas~=1.5.0",
-    "matplotlib>=3.5.0",
+    "pandas~=2.0.0",
+    "matplotlib>=3.6.1",
```
williamjameshandley, line 29: In the interests of completeness, I think 3.6.1 is required by pandas 2.0.0, so you can drop this hard-coded number.

williamjameshandley (changes requested):
I'm happy for this to be merged, although it's currently such a cutting edge requirement that even [arch linux](https://archlinux.org/packages/extra/x86_64/python-pandas/) hasn't yet updated to pandas 2.0.0. It's quite neat that we'll be releasing anesthetic 2.0.0 to time with the pandas major version upgrade.

@lukashergt are you happy to update to pandas 2.0.0?

pyproject.toml
AdamOrmondroyd: Without it, it breaks the minimum requirements workflow.

AdamOrmondroyd: Unless you mean remove matplotlib completely?

williamjameshandley: Ah -- you're right. pandas doesn't require matplotlib, but if it's present requires it at this level, so we do indeed need to specify.

lukashergt (commented):
It would have been nice to be compatible with lower pandas versions, but considering how much we are building on pandas, it does indeed seem fitting to have our major version upgrade more or less coincide with the pandas one. 

Thanks @AdamOrmondroyd, just one question regarding the pandas version requirement. Otherwise happy for this to be merged.

pyproject.toml
```diff
@@ -25,8 +25,8 @@ requires-python = ">=3.8"
 dependencies = [
     "scipy",
     "numpy",
-    "pandas~=1.5.0",
-    "matplotlib>=3.5.0",
+    "pandas~=2.0.0",
```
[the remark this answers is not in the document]
AdamOrmondroyd: I used ~= to force us to actually check that e.g. pandas 2.1 doesn't break anything when it comes along. I can change if you would prefer >=?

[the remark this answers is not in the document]
lukashergt: I think `>=` would be better. Should pandas 2.1 indeed break something, it will pop up in our daily scheduled CI checks of the master branch with `>=`. When the master branch fails from one day to the next it is typically down to a new matplotlib or pandas version. With `~=`, on the other hand, we won't notice it in the daily checks.

lukashergt (approved):
Many thanks @AdamOrmondroyd, apologies for the back and forth.

Ready to squash and merge for me, but since @williamjameshandley requested changes, he will have to approve, too, I believe.

PR #293, opened 2023-06-09
Fix to anesthetic gui

williamjameshandley:
# Description

This accomplishes two things

a) Very quick fix to anesthetic gui. It's not currently working when called as a script, since [sliders need to be stored in a variable](https://stackoverflow.com/a/49737989). This PR fixes this. It's quite hard to write a test for it, since this bug only shows up when working interactively.

b) More robust slider behaviour. At the moment, the temperature slider ranges from T=1e-1 to T=1e5, logarithmically between the two, and the evolution slider by iteration. These can result in surprising or unhelpful behaviour. With this upgrade, we move to 'beta' (inverse temperature, consistent with the actual interface), with beta now placed on a non linear scale which moves the peak of the typical set/higson plot linearly along the logX axis. The evolution slider now moves with logX, rather than iteration, which is better behaved when the number of live points is varying over the run.

# Checklist:


anesthetic/gui/plot.py
```diff
@@ -90,23 +90,27 @@ def set_text(self, logL, n):
...
     """Logarithmic slider controlling temperature of the posterior points."""
 
-    def __init__(self, fig, gridspec, action):
-        super().__init__(fig, gridspec, action, r'$kT$', -1, 5, 0, 'vertical')
+    def __init__(self, beta, D_KL, fig, gridspec, action):
+        self._beta = beta
+        self._D_KL = D_KL
+        D_KL0 = np.interp(1, beta, D_KL)
+        super().__init__(fig, gridspec, action, r'$\beta$',
+                         D_KL[0], D_KL[-1], D_KL0, 'vertical')
```
lukashergt, line 98: Out of curiosity, how does D_KL factor into the beta range/scaling? 
Or is the better question why/how does D_KL govern the peak position on the Higson plot?

anesthetic/gui/plot.py
```diff
@@ -245,8 +252,8 @@ def points(self, label):
     def update(self, _):
         """Update all the plots upon slider changes."""
         logX = np.log(self.samples.nlive / (self.samples.nlive+1)).cumsum()
-        kT = self.temperature()
-        LX = self.samples.logL/kT + logX
+        beta = self.temperature()
+        LX = self.samples.logL*beta + logX
```
lukashergt, line 256: This relates to my comment above, `beta = temperature` is a bit confusing, so I think we should rename the temperature slider to a beta slider.

lukashergt (changes requested):
> a) Very quick fix to anesthetic gui. It's not currently working when called as a script, since [sliders need to be stored in a variable](https://stackoverflow.com/a/49737989). This PR fixes this. It's quite hard to write a test for it, since this bug only shows up when working interactively.

Hmm, what do you mean by working interactively? When I call the (master) anesthetic gui script from the command line, it works fine for me... maybe this is platform dependent... Anyhow, things still work fine with these changes.

With the change from a temperature slider to a beta slider (I like the change), we should also rename the `Temperature` class to a `Beta` class and the `RunPlotter.temperature` attribute to a `RunPlotter.beta` attribute, and update docstrings accordingly.

anesthetic/gui/plot.py
williamjameshandley: To quote Ed Jaynes via Skilling, "this is well known to those who know such things". The D_KL is the information, i.e. the size of the typical set, so most of the posterior bulk/non-zero weights will appear once nested sampling has compressed by about this much (and dissappear as we move onward ard upward). [Sklling](https://projecteuclid.org/journals/bayesian-analysis/volume-1/issue-4/Nested-sampling-for-general-Bayesian-computation/10.1214/06-BA127.full) covers this throughout the initial nested sampling paper, but it first appears in section 4.

anesthetic/gui/plot.py
williamjameshandley: done

lukashergt (dismissed):
Looks good, please squash and merge.

lukashergt (approved):

lukashergt (approved):

PR #306, opened 2023-06-28
dlogX -> logdX

williamjameshandley:
# Description

This has come up in the past, but explicitly today with [margret](https://www.mpa-garching.mpg.de/members/122289), the notation `dlogX` is unfortunate, in that it is quite definitely wrong. In the spirit of `logL`, it should really be `logdX` since it is the log of small volume shell dX. This makes expressions make a little more sense.

This is a breaking change (albeit a minor one -- very few users will be using this underlying code), so it should be merged before 2.0.0 release. I have added a helpful error to the old notation.

# Checklist:


tests/test_samples.py
```diff
@@ -1347,6 +1347,8 @@ def test_old_gui():
     with pytest.raises(NotImplementedError):
         make_1d_axes(['x0', 'y0'], tex={'x0': '$x_0$', 'y0': '$y_0$'})
 
+    with pytest.raises(NotImplementedError):
+        samples.dlogX(1000)
```
lukashergt, line 1351: A bit funny, that the removal of `dlogX` is what introduces its first test...

Should we add a test for `logdX` to at least cover its existence in the tests?

lukashergt (commented):
This definitely clears up the notation, very nice. Should we add a check for `logdX`'s existence?

lukashergt (approved):
Great, please squash and merge.

Issue #110, opened 2020-07-01
change all for-loops in unit tests to parametrised tests


lukashergt:
Instead of using for-loops in unit tests it would be better to use parametrised tests using [@pytest.mark.parametrize](https://docs.pytest.org/en/stable/parametrize.html#pytest-mark-parametrize-parametrizing-test-functions). These decorators can also be stacked on top of each other in which case all combinations are performed (the equivalent of nested for-loops). 

Using parametrised tests has the huge advantage that a failed test will tell which parametrisation exactly caused the issue, which would be lost in a for-loop.

Some of the unit tests have already been converted to parametrised tests. It would be good to extend that to all other tests that involve for-loops.

williamjameshandley:
Partially implemented by @lukashergt in #109 and #99.

williamjameshandley:
We should also move to more skips of e.g. fgivenx and astropy rather than checks #314

PR #314, opened 2023-07-17
minor bug fixes

lukashergt:
* The `neff` method of `WeighedDataFrame` was missing its optional `beta` kwarg. Failing test in https://github.com/handley-lab/anesthetic/pull/314/commits/1f479b7aaef8708fba00ace9d5c2905b2005627c highlights the issue.
* When running `samples.plot_1d()` pandas annoyingly introduces the ylabel "Frequency" (see for example the [2.0.0 version of readthedocs](https://anesthetic.readthedocs.io/en/stable/quickstart.html#plot-example-1-marginalised-1d-posteriors)), which this PR removes. 

# Checklist:


williamjameshandley (dismissed):

williamjameshandley (dismissed):

anesthetic/plot.py
```diff
@@ -697,6 +697,7 @@ def make_2d_axes(params, labels=None, lower=True, diagonal=True, upper=True,
                          ticks=ticks,
                          gridspec_kw=gridspec_kw,
                          subplot_spec=subplot_spec)
+    fig.align_labels()
     return fig, axes
```
lukashergt, line 701: I snuck this little line in. I can revert it if so wished, but I find this significantly improves the appearance of triangle plots by aligning the labels independent of the length of the tick labels. Here a comparison:

* without alignment:
  image
* with alignment:
  image

williamjameshandley:
To deal with the irritating frequency labelling, it's better to oblate the offending [Frequency code](https://github.com/pandas-dev/pandas/blob/3c01ce2af5099a8a830195b2050d3a4036d14ff3/pandas/plotting/_matplotlib/hist.py#L182-L188). This therefore covers the case when `samples.x0.plot.hist_1d` is called on its own, rather than `hist_1d`. This has been fixed in 9925b07. I've also added a test for this on b02eb54 which fails on master.

I've got another bug I'd like to squash associated with hist_1d and its plot range (which is why I'm a bit more into this than one might expect)

anesthetic/plotting/_matplotlib/hist.py
```diff
@@ -138,6 +142,20 @@ class Hist1dPlot(HistPlot):
...
+            weights = self.kwds.get("weights", None)
+            xmin = quantile(data, q[0], weights)
+            xmax = quantile(data, q[-1], weights)
+            self.kwds["range"] = (xmin, xmax)
+            result = super()._calculate_bins(data)
+            self.kwds.pop("range")
+        else:
+            result = super()._calculate_bins(data)
+        return result
+
```
williamjameshandley, line 158: In order to get 1d histogram binning and ranges to work, this code had to be copied from `anesthetic.plot.hist_plot_1d`. Without it, bins are calculated on the full range of the data.

This kind of thing would further justify moving our current `anesthetic.plot` code into `anesthetic.plotting._matplotlib`, but that would likely be a 3.0.0 move. Flagged in #318.

tests/test_samples.py
```diff
@@ -1380,6 +1366,23 @@ def test_samples_plot_labels():
         assert samples.get_label(col) == ax.get_xlabel()
 
 
+@pytest.mark.parametrize('kind', ['kde', 'hist', 'fastkde']
+                         if 'fastkde' in sys.modules else
+                         ['kde', 'hist'])
```
williamjameshandley, line 1371: Is this the best way to deal with fastkde, or should it be skipped/xfailed?

tests/test_samples.py
```diff
@@ -1693,3 +1696,83 @@ def test_groupby_plots():
...
+    for ax in axes:
+        assert ax.get_ylabel() != 'Frequency'
+
+    fig, ax = plt.subplots()
+    ax = pc['x0'].plot(kind='hist_1d', ax=ax)
+    assert ax.get_ylabel() != 'Frequency'
+
+    fig, ax = plt.subplots()
+    ax = pc.x0.plot.hist_1d(ax=ax)
+    assert ax.get_ylabel() != 'Frequency'
```
williamjameshandley, line 1718: Added this to explicitly check Frequency

tests/test_samples.py
```diff
@@ -1693,3 +1696,83 @@ def test_groupby_plots():
...
+
+    pc.x0 -= 6
+    pc.x1 += 6
+    pc.plot_2d(axes, kind=f'{kind}_2d', **kwargs)
+    xmin, xmax = axes['x0']['x1'].get_xlim()
+    ymin, ymax = axes['x0']['x1'].get_ylim()
+    assert -3.9 < xmin < -3
+    assert 3 < xmax < 3.9
+    assert -3.9 < ymin < -3
+    assert 3 < ymax < 3.9
```
williamjameshandley, line 1778: This code explictly checks that upon replotting, the axes expand to include all plots. Note that infuriatingly this does not work for non-levelled hist_2d plots, since [matplotlib.axes.Axes.hist2d](https://github.com/matplotlib/matplotlib/blob/d5210e4a87665581205f5147703dbd229b44cf66/lib/matplotlib/axes/_axes.py#L7119-L7120) has the behaviour that "Currently hist2d calculates its own axis limits, and any limits previously set are ignored." in contrast to other matplotlib functions (and the desired functionality for anesthetic).

williamjameshandley (commented):
You should check my additions, and if you think they look good, we should merge this before the PR becomes too unwieldy.

williamjameshandley (dismissed):

williamjameshandley (dismissed):

tests/test_samples.py
lukashergt: This achieves the same as skipping. Skipping would appear in the tests as a little `s` rather than in the form of fewer tests. Would we prefer that?

tests/test_samples.py
lukashergt: Why not check for `ax.get_ylabel() == ''` rather than `ax.get_ylabel() != 'Frequency'`?

tests/test_samples.py
lukashergt: > Note that infuriatingly this does not work for non-levelled hist_2d plots, since [matplotlib.axes.Axes.hist2d](https://github.com/matplotlib/matplotlib/blob/d5210e4a87665581205f5147703dbd229b44cf66/lib/matplotlib/axes/_axes.py#L7119-L7120) has the behaviour that "Currently hist2d calculates its own axis limits, and any limits previously set are ignored." in contrast to other matplotlib functions (and the desired functionality for anesthetic).

It probably isn't that big a deal for the typical anesthetic user, since it's probably rare to expand on the plotting range of a hist2d plot. You'd want a 2d histogram to cover the entire range of interest in the first place...

anesthetic/plotting/_matplotlib/hist.py
lukashergt: Is there a test for this (that would fail on the master branch)? 
Since this is indeed in `hist_plot_1d` I can't quite see, why we need this...

tests/test_samples.py
williamjameshandley: The changes are specifically to stop anesthetic placing its own (hard-coded) 'Frequency' label, so this test explicitly checks this. There is a possible future where we want to put our own label on this, which a check against a blank string would fail.

tests/test_samples.py
williamjameshandley: Fewer lines in the code is good. Being explicit that tests are skipped when not everything is installed is better. Mentioned in #110.

anesthetic/plotting/_matplotlib/hist.py
williamjameshandley: >Is there a test for this (that would fail on the master branch)?

Yes! [The new test](https://github.com/lukashergt/anesthetic/blob/50601e4169a2c5280b943cbd7e30a03bb43e9250/tests/test_samples.py#L1748-L1782) in the 'hist' branch will fail on master, since without this code levelled histograms adjust the ranges incorrectly.

williamjameshandley:
Feel free to squash and merge @lukashergt if you agree with the changes

williamjameshandley (approved):

PR #319, opened 2023-07-25
Test suite tidy

### LINKED ISSUE #110: change all for-loops in unit tests to parametrised tests

williamjameshandley:
# Description

This PR begins a long road to upgrading the test suite with more systematic parametrisation. First step is to move to skipping tests rather than using 'if' statements, which the first commit does.

Fixes #110

# Checklist:


tests/test_reader.py
```diff
@@ -189,9 +179,7 @@ def test_read_multinest():
     ns.plot_1d(['x0', 'x1', 'x2', 'x3'])
 
 
-@pytest.mark.xfail('h5py' not in sys.modules,
-                   raises=NameError,
-                   reason="ultranest reading requires `h5py` package")
+@h5py_mark_xfail
```
lukashergt, line 182: Should this test be simply skipped rather than xfailed?

tests/test_reader.py
```diff
@@ -269,8 +257,7 @@ def test_regex_escape():
 
 
 @pytest.mark.parametrize('root', ['pc', 'gd'])
-@pytest.mark.skipif(pytables_imported is False,
-                    reason="requires tables package")
+@pytables_mark_xfail
```
lukashergt, line 260: Should this test be simply skipped rather than xfailed?

tests/test_samples.py
```diff
@@ -366,42 +361,35 @@ def test_plot_1d_colours():
...
-        assert len(set(pc_colors)) == 1
+    assert len(set(gd_colors)) == 1
+    assert len(set(mn_colors)) == 1
+    assert len(set(pc_colors)) == 1
 
 
-@pytest.mark.xfail('astropy' not in sys.modules,
-                   raises=ImportError,
-                   reason="requires astropy package")
+@astropy_mark_xfail
```
lukashergt, line 392: Should this test be simply skipped rather than xfailed?

lukashergt (commented):
See some comments inline.

tests/test_convert.py
```diff
@@ -1,20 +1,19 @@
 from anesthetic import read_chains
 from anesthetic.convert import to_getdist
 from numpy.testing import assert_array_equal
+from utils import getdist_mark_xfail
 
 
+@getdist_mark_xfail
```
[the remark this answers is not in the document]
williamjameshandley: If we don't run it, then we miss 100% coverage -- one could write specific tests which check for failure, but I think this is neater.

anesthetic/read/ultranest.py
```diff
@@ -23,6 +19,7 @@ def read_ultranest(root, *args, **kwargs):
     num_params = len(labels)
 
     filepath = os.path.join(root, 'results', 'points.hdf5')
+    import h5py
```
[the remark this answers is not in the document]
williamjameshandley: done in 3d4af45f

tests/utils.py
```diff
@@ -0,0 +1,76 @@
...
+
+condition = 'astropy' not in sys.modules
+reason = "requires astropy package"
+raises = ImportError
+astropy_mark_skip = pytest.mark.skipif(condition, reason=reason)
+astropy_mark_xfail = pytest.mark.xfail(condition, raises=raises, reason=reason)
+
+
+def astropy_skipif(param):
+    return pytest.param(param, marks=astropy_mark_skip)
```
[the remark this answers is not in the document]
williamjameshandley: Agreed. Changed in 870e0a2 (pleasingly no new line breaks)

tests/test_plot.py
```diff
@@ -337,110 +337,102 @@ def test_2d_axes_scatter(axesparams, params, upper):
...
+    # Check xlim, Uniform (i.e. data and limits span entire prior boundary)
+    fig, ax = plt.subplots()
+    data = np.random.uniform(size=1000)
+    plot_1d(ax, data, q=0)
+    xmin, xmax = ax.get_xlim()
+    assert xmin <= 0
+    assert xmax >= 1
 
 
+@fastkde_mark_xfail
```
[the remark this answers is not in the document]
williamjameshandley: If we don't run it, then we miss 100% coverage -- one could write specific tests which check for failure, but I think this is neater.

tests/test_plot.py
```diff
@@ -485,19 +477,16 @@ def test_hist_plot_1d():
     assert polygon.get_ec() == ColorConverter.to_rgba('r')
 
 
+@astropy_mark_xfail
```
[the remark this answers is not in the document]
williamjameshandley: If we don't run it, then we miss 100% coverage -- one could write specific tests which check for failure, but I think this is neater.

tests/test_reader.py
williamjameshandley: If we don't run it, then we miss 100% coverage -- one could write specific tests which check for failure, but I think this is neater.

tests/test_reader.py
williamjameshandley: If we don't run it, then we miss 100% coverage -- one could write specific tests which check for failure, but I think this is neater.

tests/test_samples.py
williamjameshandley: If we don't run it, then we miss 100% coverage -- one could write specific tests which check for failure, but I think this is neater.

anesthetic/samples.py
```diff
@@ -413,8 +413,8 @@ def plot_2d(self, axes=None, *args, **kwargs):
         'kde': {'diagonal': 'kde_1d', 'lower': 'kde_2d'},
         'kde_1d': {'diagonal': 'kde_1d'},
         'kde_2d': {'lower': 'kde_2d'},
-        'fastkde': {'diagonal': 'fastkde_1d', 'lower': 'fastkde_2d'},
         'hist': {'diagonal': 'hist_1d', 'lower': 'hist_2d'},
+        'fastkde': {'diagonal': 'fastkde_1d', 'lower': 'fastkde_2d'},
```
lukashergt, line 417: I prefer the old ordering: kdes, fastkde, hists.
I think this comes from the already merged #314.

tests/test_samples.py
lukashergt: Do we really miss 100% coverage? Normally our CI should be set up such that `extras: true` runs all tests without any skipping which gives us the 100% coverage. `extras: false` tests the minimal anesthetic version and is allowed to have less than 100% coverage. 

That is why the coverage tests often initially show up as failed, because the faster `extras: false` CI jobs finish first with insufficient coverage and we have to wait for the first `extras: true` to run through to get 100% coverage.

lukashergt (commented):
More comments inline.

anesthetic/samples.py
williamjameshandley: mismerge -- fixed it cbab943

tests/test_samples.py
williamjameshandley: I've changed xfails to skips so you can see the problem at [this permalink](https://app.codecov.io/gh/handley-lab/anesthetic/commit/9e159dfcdcd1cf7d37c1e87c5d9d003f31e8d1c4/blob/anesthetic/convert.py)

lukashergt: Ah, ok, thanks. It is the try-except blocks in those tests, where we provide informative error messages if a package is not installed, e.g.:
```python
    try:
        import getdist
    except ModuleNotFoundError:
        raise ImportError("You need to install getdist to use to_getdist")
```
So in this getdist example the xfail with condition runs the actual test if getdist is installed (where codecov misses the except block) and checks for this ImportError if it is not (where codecov hits the except block).

lukashergt (approved):
Thanks @williamjameshandley, please squash and merge.

williamjameshandley:
Many thanks @lukashergt

PR #320, opened 2023-07-25
anesthetic 2 deprecation cleanup

williamjameshandley:
# Description

This PR can be merged when we are happy to deprecate the guardrails for users still transferring from anesthetic 1.

# Checklist:


lukashergt:
In light of the rather fast incrementing version number, I'd suggest giving it some more time... 
Maybe ~o(3months) from 2.0.0 release?

williamjameshandley:
> In light of the rather fast incrementing version number, I'd suggest giving it some more time... Maybe ~o(3months) from 2.0.0 release?

Agreed * sets reminder *

anesthetic/samples.py
```diff
@@ -370,21 +330,6 @@ def plot_2d(self, axes=None, *args, **kwargs):
...
-                            "keywords for anesthetic. Your plots may look "
-                            "odd if you use this argument."
-                            )
-                    elif lkwargs['kind'] == 'hist':
-                        warnings.warn(
-                            "You are using \'hist\' as a plot kind. "
-                            "\'hist_1d\' and \'hist_2d\' are the appropriate "
-                            "keywords for anesthetic. Your plots may look "
-                            "odd if you use this argument."
-                            )
```
AdamOrmondroyd, line 387: I don't understand why these warnings should be removed?

lukashergt: Agreed, these will actually still be useful for new users of anesthetic. This is different from warnings that are about changes from anesthetic 1 to anesthetic 2.

So, yes, I think we should keep them, but then they should become part of the tests. Let's address this in #342.

lukashergt (changes requested):

PR #323, opened 2023-07-31
Flake8 update fix

williamjameshandley:
# Description

The CI is failing due to a recent upgrade in flake8. This replaces the appropriate `==` comparison with `is` for testing types

# Checklist:


anesthetic/plot.py
```diff
@@ -969,7 +969,7 @@ def hist_plot_1d(ax, data, *args, **kwargs):
     xmax = quantile(data, q[-1], weights)
     range = kwargs.pop('range', (xmin, xmax))
 
-    if type(bins) == str and bins in ['knuth', 'freedman', 'blocks']:
+    if type(bins) is str and bins in ['knuth', 'freedman', 'blocks']:
```
lukashergt, line 972: I believe the more canonical way is actually to use `isintance(bins, str)`.

tests/test_reader.py
```diff
@@ -281,14 +281,14 @@ def test_hdf5(tmp_path, root):
 
     with HDFStore(filename) as store:
         assert_frame_equal(samples, store[key])
-        assert type(store[key]) == type(samples)
+        assert type(store[key]) is type(samples)
```
lukashergt, line 284: Possibly `isinstance(store[key], type(samples))` or `isinstance(store[key], Samples)` would be more appropriate. `type(.) is type(.)` is stricter. Not sure which would be better here and in the rest of this test...

lukashergt (commented):
See comments inline.

anesthetic/plot.py
williamjameshandley: agreed. Updated in c6d173e

tests/test_reader.py
williamjameshandley: stricter is better here -- we definitely want the type to not change in between writing and reading.

lukashergt (approved):

Issue #96, opened 2020-06-08
RuntimeError in kde plot when data includes NaNs


lukashergt:
When the data contains NaNs `kde_contour_plot_2d` raises a RuntimeError. This should not happen for cases where the corresponding `logL` is `-inf` and `weight` is `0`. In such cases these entries should be simply ignored.

NaNs can appear for derived parameters which never got calculated when the sampling parameters fell into an excluded region (`logL=-inf` and `weight=0`) of parameter space.

### MWE

data: [unilog_r.log](https://github.com/williamjameshandley/anesthetic/files/4746944/unilog_r.log)

```python
data = pandas.read_csv("unilog_r.log", sep='\t')
ns = NestedSamples(data=data[['n_s', 'r', 'logr']], logL=data.logL, logL_birth=data.logL_birth, w=data.weight)
ns.n_s[:10] = np.nan
assert np.all(ns.weight[:10] == 0)
fig = plt.figure(figsize=(5, 5))
fig, axes = make_2d_axes(['n_s', 'r', 'logr'], fig=fig)
ns.plot_2d(axes)
axes.iloc[-1, 0].legend(bbox_to_anchor=(len(axes), len(axes)), loc='upper left');
```

The error message:
```python
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-160-14d6415b9f86> in <module>
      5 fig = plt.figure(figsize=(5, 5))
      6 fig, axes = make_2d_axes(['n_s', 'r', 'logr'], fig=fig)
----> 7 ns.plot_2d(axes)
      8 axes.iloc[-1, 0].legend(bbox_to_anchor=(len(axes), len(axes)), loc='upper left');

~/.virtualenvs/py38env/lib/python3.8/site-packages/anesthetic/samples.py in plot_2d(self, axes, *args, **kwargs)
    352                     plot_type = types.get(pos, None)
    353                     lkwargs = local_kwargs.get(pos, {})
--> 354                     self.plot(ax_, x, y, plot_type=plot_type, *args, **lkwargs)
    355 
    356         return fig, axes

~/.virtualenvs/py38env/lib/python3.8/site-packages/anesthetic/samples.py in plot(self, ax, paramname_x, paramname_y, *args, **kwargs)
    192                     x = self[paramname_x]
    193                     y = self[paramname_y]
--> 194                     return kde_contour_plot_2d(ax, x, y, weights=self.weight,
[...]

PR #324, opened 2023-08-01
Fix for logL_birth

### LINKED ISSUE #310: Plots with logL birth
As discussed with @lukashergt, whilst passing samples from [nessai](https://github.com/mj-will/nessai) and producing plots, I ran into something that I found slightly confusing. 

Here's some pseudocode for what I was doing:

```python
nested_samples =   ... # np.ndarray
names = ... # List[str]
logL = ... # np.ndarray
logL_birth = ... # np.ndarray (contains -inf)

samples = NestedSamples(
    nested_samples,
    columns=names,
    logL=logL,
    logL_birth=logL_birth,
)
samples.plot_1d()
```

This produces a `ValueError` because of the -infs in `logL_birth`. 

I found this counterintuitive because I assumed calling `samples.plot_1d()` would work if I had passed the correct inputs to `NestedSamples`. Given `logL_birth` will typically contain -inf (or very negative numbers), do you think there should be a special case for this column so that it doesn't fail?

williamjameshandley:
# Description

This PR fixes #310 by adding an explicit guard for logL_birth, dropping nans and infs before plotting

It's not the neastest solution, so any suggestions for improvements are welcome.

# Checklist:


anesthetic/samples.py
```diff
@@ -244,8 +244,12 @@ def plot_1d(self, axes=None, *args, **kwargs):
...
             if x in self and kwargs['kind'] is not None:
                 xlabel = self.get_label(x)
-                self[x].plot(ax=ax, xlabel=xlabel,
-                             *args, **kwargs)
+                if x == 'logL_birth':
+                    self[x].replace(-np.inf, np.nan
+                                    ).dropna().plot(ax=ax, xlabel=xlabel,
+                                                    *args, **kwargs)
+                else:
+                    self[x].plot(ax=ax, xlabel=xlabel, *args, **kwargs)
```
lukashergt, line 252: * The `.dropna()` shouldn't be necessary.
* Shall we separate inf replacement and plotting to reduce code duplication? We could have 
  ```python
  if x == 'logL_birth':
      selfx = self[x].replace(...)
  else:
      selfx = self[x]
  selfx.plot(...)
  ```
* I feel like changing infs to nans shouldn't happen silently. Shall we at least have a warning that we are setting `-inf` to `nan` for plotting purposes?

anesthetic/samples.py
```diff
@@ -389,13 +393,27 @@ def plot_2d(self, axes=None, *args, **kwargs):
...
-                                      ylabel=ylabel, *args, **lkwargs)
+                            if x == 'logL_birth' or y == 'logL_birth':
+                                self.replace(-np.inf, np.nan
+                                             ).dropna().plot(x, y, ax=ax,
+                                                             xlabel=xlabel,
+                                                             ylabel=ylabel,
+                                                             *args, **lkwargs)
+                            else:
+                                self.plot(x, y, ax=ax, xlabel=xlabel,
+                                          ylabel=ylabel, *args, **lkwargs)
```
lukashergt, line 416: See `plot_1d` comment.

tests/test_samples.py
```diff
@@ -440,6 +440,10 @@ def test_plot_1d_no_axes():
     assert axes.iloc[0].get_xlabel() == 'x0'
     assert axes.iloc[1].get_xlabel() == 'x1'
     assert axes.iloc[2].get_xlabel() == 'x2'
+    axes = ns.plot_1d()
+    assert axes['logL_birth'].get_xlim()[0] > -1000
```
lukashergt, line 444: Is this `xlim[0] > -1000` check future proof? 
Is it needed in the first place? 
Would a check that `ns.plot_1d()` runs without error be enough?
If we add a warning, catching a warning here might actually be nice.

lukashergt (commented):
See comments inline.

anesthetic/samples.py
williamjameshandley: dropna is necessary for the 2D KDE, but not otherwiseo so I've removed those.
I've reduced code repetition as suggested.
With regard to infs to nans, I'm not that keen to have a warning for default behaviour (i.e. `samples.plot_1d()`)

tests/test_samples.py
williamjameshandley: I've removed the check.

anesthetic/samples.py
lukashergt: > With regard to infs to nans, I'm not that keen to have a warning for default behaviour (i.e. `samples.plot_1d()`)

Well, I'd argue here that `samples.plot_1d()` or `samples.plot_2d()` is not default behaviour for nested sampling data (nor for anesthetic dataframes in general, which typically come with very many columns), otherwise this would have come up earlier, too. By default the user should specify which columns are to be plotted. 

So I think a warning message could help the new user (who is most likely to call the plotting functions without specifying columns) to learn faster how to use the plotting command. This is also how this PR came to be: new user simply trying `samples.plot_1d()` to see if it works...

I agree that too many warning messages are annoying and get in the way, but how often do we intentionally plot the posterior distribution of `logL_birth`? Also, for people that indeed want to intentionally plot `logL_birth`, they can easily avoid the warning by dropping infs first.

williamjameshandley: I am persuaded by this, so think we should have a warning message.

If we're going down a warning route, should we restrict this to logL_birth ? The code would be much neater if we just moved the relevant inf guards into `kde_contour_plot_2d` and `kde_1d` and `kde_2d`. This would then be more consistent with the other functions behaviour, e.g. `samples.plot()`, which just ignores nans and infs.

README.rst
```diff
@@ -136,15 +136,15 @@ If you use ``anesthetic`` to generate plots for a publication, please cite
...
 
 or using the BibTeX:
 
 .. code:: bibtex
 
    @article{anesthetic,
-       doi = {10.21105/joss.01414},
-       url = {http://dx.doi.org/10.21105/joss.01414},
+       doi = {10.2.1.6/joss.01414},
+       url = {http://dx.doi.org/10.2.1.6/joss.01414},
```
lukashergt, line 147: Accidental DOI change?

williamjameshandley:
Note that this has come up in the past in #96, which this PR obseletes, so that test is now removed.

anesthetic/samples.py
yallup: My input from stumbling across this discussion via #332 

Perhaps the default behavior is sensible (include nlive logl and loglbirth in the plot if nothing is specified), but it strikes me that the easiest thing to make the various kde errors less impactful is add a kind=scatter and make that the default for the specific case that the columns to plot are empty?

lukashergt:
The current question is whether we should change the default behaviour to ignoring the columns `logL`, `logL_birth`, and `nlive` by default?

I would vote against that default change. `plot_1d` and `plot_2d` are more general `Samples` methods which I don't think should drop any columns by default. 

Those columns are more specific `NestedSamples` parameters. We could create `NestedSamples.plot_1d` and `NestedSamples.plot_2d` methods which would handle the treatment of those parameters (and then forward to `Samples.plot_1d` etc.), be it a warning or a change to default behaviour. I'd prefer a simple warning.

> Perhaps the default behavior is sensible (include nlive logl and loglbirth in the plot if nothing is specified), but it strikes me that the easiest thing to make the various kde errors less impactful is add a kind=scatter and make that the default for the specific case that the columns to plot are empty?

I'd keep the kde default, but am open to be convinced otherwise.
Either way, it might be nice to have a simpler shortcut for scatter plots, so maybe we should create a `kind='scatter'` shortcut, which uses `'scatter_2d'` in the lower triangle (and `'hist_1d'` on the diagonal?)?

williamjameshandley:
> Either way, it might be nice to have a simpler shortcut for scatter plots, so maybe we should create a `kind='scatter'` shortcut, which uses `'scatter_2d'` in the lower triangle (and `'hist_1d'` on the diagonal?)?

Added `scatter` and `scatter_2d` to default kinds in b51b3ef

README.rst
williamjameshandley: from `bin/bump_version` sed command. Fixed in df622ef

williamjameshandley:
I would also vote to keep things in `Samples` and avoid a `NestedSamples` override.

anesthetic/samples.py
```diff
@@ -388,14 +393,22 @@ def plot_2d(self, axes=None, *args, **kwargs):
                     if x in self and y in self and lkwargs['kind'] is not None:
                         xlabel = self.get_label(x)
                         ylabel = self.get_label(y)
+                        if ~np.isfinite(self[x]).all():
+                            warnings.warn(f"column {x} has inf values.")
```
lukashergt, line 397: This will currently warn that there are inf values when possibly there are no inf values, just some nan values, which from the pandas perspective are fine to be ignored... 

Might be better to work with `np.isinf(self[x]).any()`, here.

anesthetic/samples.py
```diff
@@ -388,14 +393,22 @@ def plot_2d(self, axes=None, *args, **kwargs):
...
+                            selfx = self[x].replace([-np.inf, np.inf], np.nan)
+                            selfx.plot(ax=ax.twin, xlabel=xlabel,
+                                       *args, **lkwargs)
                             ax.set_xlabel(xlabel)
                             ax.set_ylabel(ylabel)
                         else:
-                            self.plot(x, y, ax=ax, xlabel=xlabel,
-                                      ylabel=ylabel, *args, **lkwargs)
+                            if ~np.isfinite(self[y]).all():
+                                warnings.warn(f"column {y} has inf values.")
```
lukashergt, line 406: see other comment

lukashergt (changes requested):

README.rst
lukashergt: fixed in https://github.com/handley-lab/anesthetic/pull/324/commits/247c10c2b905b39e6f6c3749e332ad34a6c54ae2

tests/test_samples.py
```diff
@@ -429,6 +432,10 @@ def test_plot_1d_no_axes():
     assert axes.iloc[1].get_xlabel() == 'x1'
     assert axes.iloc[2].get_xlabel() == 'x2'
 
+    axes = ns.plot_1d()
+    axes = ns[['x0', 'logL_birth']].plot_1d()
+    axes = ns.drop_labels()[['x0', 'logL_birth']].plot_1d()
```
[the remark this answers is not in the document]
lukashergt: done in https://github.com/handley-lab/anesthetic/pull/324/commits/d2915fb7b1c2f5e0d1c2d2917c019dcd4b983922

tests/test_samples.py
```diff
@@ -416,6 +416,9 @@ def test_plot_2d_no_axes():
     assert axes.iloc[-1, 1].get_xlabel() == 'x1'
     assert axes.iloc[-1, 2].get_xlabel() == 'x2'
 
+    axes = ns[['x0', 'logL_birth']].plot_2d()
+    axes = ns.drop_labels()[['x0', 'logL_birth']].plot_2d()
```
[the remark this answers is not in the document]
lukashergt: done in https://github.com/handley-lab/anesthetic/pull/324/commits/d2915fb7b1c2f5e0d1c2d2917c019dcd4b983922

anesthetic/samples.py
```diff
@@ -244,8 +244,10 @@ def plot_1d(self, axes=None, *args, **kwargs):
         for x, ax in axes.items():
             if x in self and kwargs['kind'] is not None:
                 xlabel = self.get_label(x)
-                self[x].plot(ax=ax, xlabel=xlabel,
-                             *args, **kwargs)
+                if ~np.isfinite(self[x]).all():
+                    warnings.warn(f"column {x} has inf values.")
```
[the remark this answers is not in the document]
lukashergt: done in https://github.com/handley-lab/anesthetic/pull/324/commits/a498a0b1304b26b627b3ce2732200c7258b80a6c

anesthetic/samples.py
lukashergt: done in https://github.com/handley-lab/anesthetic/pull/324/commits/a498a0b1304b26b627b3ce2732200c7258b80a6c

anesthetic/samples.py
lukashergt: done in https://github.com/handley-lab/anesthetic/pull/324/commits/a498a0b1304b26b627b3ce2732200c7258b80a6c

lukashergt (dismissed):
@williamjameshandley, if you are happy with my tweaks, feel free to squash and merge.

lukashergt (approved):

lukashergt (approved):

lukashergt (approved):

lukashergt (approved):

PR #328, opened 2023-08-09
implement axes logscales

lukashergt:
This PR implements logarithmic scaling of select plotting axes in triangle/rectangle plots. Effectively this means that a quantity `log(x)` is plotted while the axes labels show `x` with logarithmically spaced tick labels. 

NOTE: This has to be set up at axes creation. A simple re-scaling of x- or y-axis (e.g. `ax.set_xscale('log')`) won't work, but lead to distorted plots. To correctly compute histogram bins or KDEs, the computation has to be performed in logspace from the start.

The way I set this up here is through a `logx` kwarg for `make_1d_axes` or `plot_1d`, and through a `logxy` kwarg for `make_2d_axes` or `plot_2d`. This relates to the pandas kwargs `logx` and `logy`, e.g. in [pandas.DataFrame.plot](https://pandas.pydata.org/pandas-docs/version/0.22/generated/pandas.DataFrame.plot.html#pandas-dataframe-plot). 

Example usage:
```python
samples = read_chains("./tests/example_data/pc")
params = ['x0', 'x1', 'x2', 'x3', 'x4']
samples.plot_2d(params, logxy=['x2'])
```

To see the added documentation: 
https://anesthetic.readthedocs.io/en/logscale/plotting.html#log-scale

This PR was motivated by a discussion with @williamjameshandley and might be useful for some upcoming Gambit runs.

# Checklist:


anesthetic/plot.py
```diff
@@ -1263,17 +1340,13 @@ def hist_plot_2d(ax, data_x, data_y, *args, **kwargs):
...
-        cmax = kwargs.pop('cmax', None)
-        pdf, x, y = np.histogram2d(data_x, data_y, bins, rge,
-                                   density, weights)
+    bins = kwargs.pop('bins', 10)
+    density = kwargs.pop('density', False)
+    cmin = kwargs.pop('cmin', None)
+    cmax = kwargs.pop('cmax', None)
+    pdf, x, y = np.histogram2d(data_x, data_y, bins, rge,
+                               density, weights)
+    if levels is not None:
```
lukashergt, line 1349: Note, how I have dropped the use of `ax.hist2d`, here. 
Do we remember why we wanted this distinction? 
Tests are passing, so maybe this is an unnecessary complication?

anesthetic/plot.py
```diff
@@ -948,6 +1000,8 @@ def hist_plot_1d(ax, data, *args, **kwargs):
     kwargs = normalize_kwargs(kwargs)
     weights = kwargs.pop('weights', None)
     bins = kwargs.pop('bins', 10)
+    if ax.get_xaxis().get_scale() == 'log' and isinstance(bins, int):
+        bins = np.logspace(np.log10(data.min()), np.log10(data.max()), bins+1)
```
lukashergt, line 1004: Note that these two lines are only needed when directly calling `hist_plot_1d`. When it gets called via `plot_1d` or `plot_2d`, then this is handled in [anesthetic/plotting/_matplotlib/hist.py](https://github.com/handley-lab/anesthetic/pull/328/files#diff-10c686dc66f7a34f6647a52acd441230844a8e49d2e24a6adef1f57722a908de). 

This is similar to the double range handling from https://github.com/handley-lab/anesthetic/pull/314#discussion_r1272296536.

docs/source/plotting.rst
```diff
@@ -251,10 +251,29 @@ method from there, directing it to the correct position with the ``loc`` and
...
+.. plot:: :context: close-figs
+
+    fig, axes = make_1d_axes(['x0', 'x1', 'x2'], logx=['x2'])
+    samples.plot_1d(axes, label="'x2' on log-scale")
+
+.. plot:: :context: close-figs
+
+    fig, axes = make_2d_axes(['x0', 'x1', 'x2'], logxy=['x2'])
+    samples.plot_2d(axes, label="'x2' on log-scale")
+
```
lukashergt, line 272: New docs can be found here:
https://anesthetic.readthedocs.io/en/logscale/plotting.html#log-scale

anesthetic/plot.py
williamjameshandley: It's been there since the beginning in #46. I suspect it was due to wanting to use existing functionality where possible, but then having to implement something more specific for the level-based plotting.

Now that we're mirroring pandas more closely, the `ax.hist2d` is now accomplished (in weighted form!) by using `samples.plot.hist`, so we can indeed safely remove this. It's much neater now!

anesthetic/plot.py
```diff
@@ -43,6 +43,8 @@ class AxesSeries(Series):
     labels : dict(str:str), optional
         Dictionary mapping params to plot labels.
         Default: params
+    logx : list(str), optional
+        List of parameters to be plotted on a log scale.
```
williamjameshandley, line 47: Is this the best API? Ideally we'd want to allow for (eventual, probably not this PR) functionality for [symlog or logit](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.xscale.html) scales as well.

How about `xscale` as a keyword argument, which can function as a defaultdict between parameters and keywords (with the default being `'linear'`.

anesthetic/plot.py
```diff
@@ -148,6 +164,8 @@ class AxesDataFrame(DataFrame):
         If 'outer', plot ticks only on the very left and very bottom.
         If 'inner', plot ticks also in inner subplots.
         If None, plot no ticks at all.
+    logxy : list(str), optional
+        List of parameters to be plotted on a log scale.
```
williamjameshandley, line 168: It would be a very unusual set-up but we could in principle allow for some parameters to be log on the x axis and linear on the y axis (in e.g. a rectangular plot).

anesthetic/plot.py
```diff
@@ -768,6 +814,8 @@ def fastkde_plot_1d(ax, data, *args, **kwargs):
     i = ((x > quantile(x, q[0], p)) & (x < quantile(x, q[-1], p)))
 
     area = np.trapz(x=x[i], y=p[i]) if density else 1
+    if ax.get_xaxis().get_scale() == 'log':
+        x = 10**x
```
williamjameshandley, line 818: We could avoid this hard-coding by using the [existing matplotlib functionality](https://matplotlib.org/stable/api/scale_api.html#matplotlib.scale.LogScale.get_transform)

anesthetic/plotting/_matplotlib/hist.py
```diff
@@ -28,6 +28,8 @@
 class HistPlot(_WeightedMPLPlot, _HistPlot):
     # noqa: disable=D101
     def _calculate_bins(self, data):
+        if self.logx:
+            data = np.log10(data)
```
williamjameshandley, line 32: ideally we'd make sure these lines were covered by a test (to ensure they're doing what we think they're doing).

docs/source/plotting.rst
williamjameshandley: Nice

williamjameshandley (commented):
Outcome is excellent. I have some inline comments on the API/implementation that are related.

anesthetic/plot.py
lukashergt: `logx` is a [standard pandas kwarg](https://pandas.pydata.org/pandas-docs/version/0.22/generated/pandas.DataFrame.plot.html#pandas-dataframe-plot), so that's why I picked that... 

I did indeed start out with an `xscale` kwarg, but then stumbled over pandas handling of this and switched... What do we prefer?

anesthetic/plot.py
lukashergt: Hui, that would mess with our distinction of `DiagonalAxes` (which are for 1d plots) and `OffDiagonalAxes` (which are for 2d plots)...

anesthetic/plot.py
williamjameshandley: > `logx` is a [standard pandas kwarg](https://pandas.pydata.org/pandas-docs/version/0.22/generated/pandas.DataFrame.plot.html#pandas-dataframe-plot), so that's why I picked that...

I had not clocked this, which makes this much more convincing. Stick with logx until pandas also supports symlog/logit.

We can always add an xscale kwarg if necessary to override, but for now we should go with  the general principle of closer pandas mirroring (#59).

anesthetic/plot.py
williamjameshandley: By #59, we should probably go with `loglog` rather than `logxy`.

If somebody needs the fine-grained control I describe, we can e.g. add tuples to this list as an option at a laterdatae.

anesthetic/plot.py
williamjameshandley: This is probably overkill until/if we implement symlog/logit, so leave it as-is for now.

anesthetic/plotting/_matplotlib/hist.py
williamjameshandley: This does however need covering in order to pass the checks!

anesthetic/plot.py
lukashergt: `loglog` is never really true. `loglog` would be a kwarg used for two different parameters x and y that are both supposed to be on a log axes. Here, we are applying a log axis to a range of plots for a given parameter z and dependent on whether parameter z appears on x or y axis we are going with the `logx` or `logy` kwarg...

anesthetic/plotting/_matplotlib/hist.py
lukashergt: This is work in progress, #330 required slight re-writes which I am not done with.

lukashergt: Coverage back up to 100%.

lukashergt:
> Outcome is excellent. I have some inline comments on the API/implementation that are related.

@williamjameshandley, anything else that needs addressing here?

AdamOrmondroyd:
Just had chance to read this properly, stunning work @lukashergt

AdamOrmondroyd:
There's something odd going on with fastkde, look at the figures produced by `tests/test_samples.py::test_plot_logscale`:

`"kde"`
Screenshot 2023-10-03 at 17 58 02

`"fastkde"`
Screenshot 2023-10-03 at 17 57 17

The `"fastkde"` case is responsible for the remaining new warnings, which is why I spotted this. This is why I'm allergic to warnings!

AdamOrmondroyd:
Could this be because `kde_contour_plot_2d()` uses `cut_and_normalise_gaussian()` to give the hard cutoff at x=0, which `fastkde_contour_plot_2d()` does not?

AdamOrmondroyd:
Now I'm wondering if the x and y limits should actually be the same if only one of them is logarithmic

lukashergt:
> There's something odd going on with fastkde, look at the figures produced by `tests/test_samples.py::test_plot_logscale`:

Thanks for catching this. This was a subtle thing to do with `set_xlim` which we emit to both horizontally and vertically shared plots, which we shouldn't do if only one is logarithmic.

> Now I'm wondering if the x and y limits should actually be the same if only one of them is logarithmic

I have decoupled them now.

AdamOrmondroyd (approved):

AdamOrmondroyd:
lgtm. Great work!

williamjameshandley (approved):
I agree this looks great. I will test it out in the wild for my GAMBIT runs in the near future and get back to you if I think anything needs adjusting.

lukashergt:
Sorry, missed the approval. Thanks for the reviews.

PR #330, opened 2023-08-15
Automatic bin computation for weighted data

lukashergt:
So far we were using `bins=10` as our default bin number, matching matplotlib's behaviour. However, when plotting multiple 1d plots (as for example in triangle plots), then the same number of bins will rarely look good for different parameters. It would be more desirable to have something automatic and adaptable to the data. For unweighted data, numpy has a bunch of string options for `bins` in [`np.histogram_bin_edges`](https://numpy.org/doc/stable/reference/generated/numpy.histogram_bin_edges.html#numpy-histogram-bin-edges). This PR implements some of those string options (namely `['fd', 'scott', 'sqrt']`) for weighted data.

Many thanks go to Deborah Odunuyi who explored this in a little summer project.

# Checklist:


anesthetic/utils.py
```diff
@@ -226,6 +226,72 @@ def histogram(a, **kwargs):
     return xpath, ypath
 
 
+def histogram_bin_edges(samples, weights, bins='fd', range=None, beta='equal'):
```
williamjameshandley, line 229: Why `'equal'` as the default beta? Would be [more consistent](https://github.com/handley-lab/anesthetic/blob/337b07aa5197e2d8b95c15d7668ffdc024e06c9b/anesthetic/utils.py#L35) to go with `beta=1` unless there is a good reason here to go with equally weighted samples (also in docstring), or more consistent with e.g. Scott's rule to go with `beta=2`.

williamjameshandley (commented):
Nice. Only one comment on defaults to consider.

anesthetic/utils.py
lukashergt: This is mostly the result from playing around with different `beta` values. Note, these differences are not dramatic, they are small. However, as a tendency it looks to be on the high side in terms of number of bins, so to keep the number of bins low a higher `beta` helps.

If we want to rationalise it, these rules were created for unweighted data, so maybe it's not surprising that the effective number of samples that would compress to equally weighted samples looks the best...?

williamjameshandley (dismissed):

williamjameshandley (approved):

Issue #325, opened 2023-08-03
function for computing tension probability


lukashergt:
@williamjameshandley, do you have the code to compute things like suspiciousness and its corresponding tension probability? Could we add that somewhere? Maybe as a function taking multiple `Samples` instances in `samples.py`? 

Have you tried computing this from MCMC data?

AdamOrmondroyd:
I've put some things together in my [suspiciousness package](https://github.com/adamormondroyd/suspiciousness), might be some use here

PR #333, opened 2023-08-24
Implementing tension statistics

DilyOng:
# Description

 This is a work in progress pull request aiming to address #325 and as a learning exercise on how to do pull request.

# Checklist:

### DIFF
--- a/tension_calculator.py
+++ b/tension_calculator.py
@@ -1,0 +1,48 @@
+class tension_calculator(dict):
+     """Class for computing tension metrics
+
+     Parameters
+     ----------
+     samples : dict
+         Dictionary of NestedSamples
+     nsamples : int
+         Number of samples to use for computing statistics
+     """
+     def __init__(self, samples, nsamples=1000):
+         for k, v in tqdm.tqdm(samples.items()):
+             self[k] = v.stats(nsamples)
+
+     def union(self, A, B):
+         A_ = set(A.split('_'))
+         B_ = set(B.split('_'))
+         return  '_'.join(sorted(list(A_.union(B_))))
+
+     def intersection(self, A, B):
+         A_ = set(A.split('_'))
+         B_ = set(B.split('_'))
+         return  '_'.join(sorted(list(A_.intersection(B_))))
+
+     def stat(self, A, B, stat):
+         union = self.union(A, B)
+         intersection = self.intersection(A, B)
+         return self[union][stat] - self[A][stat] - self[B][stat] + self[intersection][stat]
+
+     def logR(self, A, B):
+         return self.stat(A, B, 'logZ')
+
+     def logS(self, A, B):
+         return self.stat(A, B, 'logL_P')
+
+     def d(self, A, B):
+         return self.stat(A, B, 'd_G')
+
+     def D_KL(self, A, B):
+         return self.stat(A, B, 'D_KL')
+
+     def p(self, A, B, d=None):
+         logS = self.logS(A, B)
+         if d is None:
+             d = self.d(A, B)
+         else:
+             d = np.ones_like(logS)*d
+         return scipy.stats.chi2.sf(d[d>0]-2*logS[d>0], d[d>0])


williamjameshandley:
I think after that it would also be good to implement a function in addition to (or possibly in place of!) for producing a `Samples` object containing columns of `logR`, `D_KL`, `d`, `S`, `p`, as an analogue to the output of `NestedSamples.stats`.

tension_calculator.py
```diff
@@ -0,0 +1,48 @@
...
+         return self[union][stat] - self[A][stat] - self[B][stat] + self[intersection][stat]
+
+     def logR(self, A, B):
+         return self.stat(A, B, 'logZ')
+
+     def logS(self, A, B):
+         return self.stat(A, B, 'logL_P')
+
+     def d(self, A, B):
+         return self.stat(A, B, 'd_G')
```
AdamOrmondroyd, line 37: I used `d` in `suspiciousness` for conciseness as I'd never actually intended it to be used by others, perhaps it would be better to stick to `d_G` to be consistent with the rest of `anesthetic`

AdamOrmondroyd (changes requested):

AdamOrmondroyd:
Please remember to remove (`git remove`) the .DSstore files you've added (they are to do with macOS file management, so not relevant to the repo)

williamjameshandley:
It would be good to get this finalised and merged now that #348 is complete -- any thoughts @DilyOng

anesthetic/tension_pvalue.py
```diff
@@ -0,0 +1,27 @@
...
+
+    logS = logR-logI
+    samples['logS'] = logS
+
+    d_G = statsA.d_G + statsB.d_G - statsAB.d_G
+    samples['d_G'] = d_G
+
+    p = chi2.sf(d_G-2*logS, d_G)
+    samples['p'] = p
+    return samples
```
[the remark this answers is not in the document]
DilyOng: Thanks for pointing out that! It is indeed a duplication. I'll delete it.

williamjameshandley (changes requested):
Hi @DilyOng, let's try to get this merged soon.

I've reorganised some of the cosmetics of the tests, but there is still some work which will make this mirror the existing statistical functionality, which I've suggested inline.

williamjameshandley:
@lukashergt @AdamOrmondroyd @DilyOng: at the moment we have a .stats function for NestedSamples. The tension_stats function assumes NestedSamples implicitly by calling this, but in practice one can compute logL_p, logS and d for (MCMC)Samples objects too, and should probably be allowed. Should we make a `.stats` function for (MCMC)Samples objects? In principle one could compute other criteria too (e.g. BIC, AIC).

anesthetic/tension.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+def tension_stats(A, B, AB, nsamples, beta):
+    # Compute Nested Sampling statistics
+    statsA = A.stats(nsamples, beta)
+    statsB = B.stats(nsamples, beta)
+    statsAB = AB.stats(nsamples, beta)
+
+    samples = Samples(index=statsA.index)
+    print(samples)
```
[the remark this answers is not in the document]
AdamOrmondroyd: it also occurs to me that a more instructive name than `samples` might be helpful (though I appreciate the `.stats` function also uses `samples`

tests/test_tension.py
```diff
@@ -0,0 +1,99 @@
+from anesthetic.examples.perfect_ns import correlated_gaussian
+import numpy as np
+from scipy.linalg import inv, solve, slogdet
+from anesthetic.tension import tension_stats
+
+
+def test_tension_stats_compatiable_gaussian():
+    d = 3
+    V = 1.
```
AdamOrmondroyd, line 9: If we really need V to be a float, then this should probably be written as 1.0. However, I don't see any reason 1 wouldn't be fine

tests/test_tension.py
```diff
@@ -0,0 +1,99 @@
...
+    logI_std = samples.logI.std()
+    logI_mean = samples.logI.mean()
+    logI_exact = -d/2 - 1/2 * slogdet(2*np.pi*(covA+covB))[1] + np.log(V)
+    assert logI_mean - 3 * logI_std < logI_exact < logI_mean + 3 * logI_std
+
+
+def test_tension_stats_incompatiable_gaussian():
+    np.random.rand(0)
+    d = 3
+    V = 1.
```
AdamOrmondroyd, line 57: ditto

AdamOrmondroyd:
> @lukashergt @AdamOrmondroyd @DilyOng: at the moment we have a .stats function for NestedSamples. The tension_stats function assumes NestedSamples implicitly by calling this, but in practice one can compute logL_p, logS and d for (MCMC)Samples objects too, and should probably be allowed. Should we make a `.stats` function for (MCMC)Samples objects? In principle one could compute other criteria too (e.g. BIC, AIC).

My ears perked up at BIC since I need to get my head around it again for xmeans clustering.

Potentially worth making an issue and listing all the things which we think we can make for each of plain old `Samples` , `MCMCSamples` and `NestedSamples`? Then we can hopefully make something without unnecessary duplication between the three.

lukashergt:
> @lukashergt @AdamOrmondroyd @DilyOng: at the moment we have a .stats function for NestedSamples. The tension_stats function assumes NestedSamples implicitly by calling this, but in practice one can compute logL_p, logS and d for (MCMC)Samples objects too, and should probably be allowed. Should we make a `.stats` function for (MCMC)Samples objects? In principle one could compute other criteria too (e.g. BIC, AIC).

Hmm, should we have a `Samples.stats` method for those stats that work for both MCMC and nested samples, and then we call `super()` in `NestedSamples.stats` and add in `logZ`, `D_KL`, and `R`?

anesthetic/tension.py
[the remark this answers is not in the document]
DilyOng: I have changed samples to samples_stats.

anesthetic/tension.py
```diff
@@ -0,0 +1,27 @@
+from anesthetic.samples import Samples
+from scipy.stats import chi2
+
+
+def tension_stats(A, B, AB, nsamples, beta):
```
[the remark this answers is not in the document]
DilyOng: @williamjameshandley Done! Please check if the docstring I wrote is correct. Thanks!

tests/test_tension.py
DilyOng: I think for python, 1. and 1.0 are identical?

AdamOrmondroyd: Indeed, in all but their legibility. While 1. doesn't violate pep, the black formatter would change it to 1.0

anesthetic/tension.py
```diff
@@ -0,0 +1,90 @@
...
+    - ``p``: p-value for the tension between two samples
+
+        .. math::
+            p = \int_{d_\mathrm{G} - 2 \log S}^{\infty} \chi^2 (x)dx
+
+    Parameters
+    ----------
+    A : :class:`Samples`/:class:`NestedSamples`
+        Dataset A
+        List or array-like of one or more nested sampling runs
```
AdamOrmondroyd, line 41: List or array-like? I'll eat my hat if these have .stats methods

anesthetic/tension.py
```diff
@@ -0,0 +1,90 @@
...
+
+    B : :class:`Samples`/:class:`NestedSamples`
+        Dataset B
+        List or array-like of one or more nested sampling runs
```
AdamOrmondroyd, line 45: Ditto

anesthetic/tension.py
```diff
@@ -0,0 +1,90 @@
...
+
+    AB : :class:`Samples`/:class:`NestedSamples`
+        Dataset AB
+        List or array-like of one or more nested sampling runs
```
AdamOrmondroyd, line 49: Ditto

tests/example_data/.DS_Store
AdamOrmondroyd, line 1: Another .DS_store

tests/test.ipynb
AdamOrmondroyd, line 1: Reminder to remove this empty file

AdamOrmondroyd (commented):
See inline

AdamOrmondroyd (commented):
See inline

docs/source/anesthetic.rst
```diff
@@ -83,6 +84,15 @@ anesthetic.scripts module
    :show-inheritance:
 
 
+anesthetic.tension module
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. automodule:: anesthetic.tension
+   :members:
+   :undoc-members:
+   :show-inheritance:
```
lukashergt, line 93: I've added the tension module to the docs. Can be found here: https://anesthetic.readthedocs.io/en/tension/anesthetic.html#module-anesthetic.tension

anesthetic/tension.py
```diff
@@ -0,0 +1,90 @@
...
+    - ``d_G``: Gaussian model dimensionality
+      (or posterior variance of the log-likelihood)
+
+      .. math::
+          d_\\mathrm{G}/2 = \\mathrm{var}(\\log L)_P
+
+    - ``p``: p-value for the tension between two samples
+
+      .. math::
+          p = \\int_{d_\\mathrm{G} - 2 \\log S}^{\\infty} \\chi^2 (x)dx
```
lukashergt, line 35: This was probably copied from `NestedSamples.stats`, let's make this more relevant for the tension module by focusing on `logR`, `logI`, `logS`.

anesthetic/tension.py
```diff
@@ -0,0 +1,90 @@
...
+
+    Parameters
+    ----------
+    A : :class:`anesthetic.samples.Samples` or \
+        :class:`anesthetic.samples.NestedSamples`
```
lukashergt, line 40: Currently this only works for `NestedSamples` and not for `Samples`. 
If we want this to work for general `Samples`, then we should add the necessary modifications to `samples.py`. 
Otherwise the type hinting should only list `NestedSamples`.

(Side note: outside `samples.py` you have to use ``:class:`anesthetic.samples.Samples` `` rather than  ``:class:`Samples` `` for the automatic linking to work.)

anesthetic/tension.py
```diff
@@ -0,0 +1,90 @@
...
+          values inferred by nested sampling
+
+    beta : float, array-like, optional
+        inverse temperature(s) beta=1/kT. Default 1
+
+    Returns
+    -------
+    samples_stats : :class:`anesthetic.samples.Samples`
+        DataFrame containing the following tension statistics:
+        logZ, D_KL, logL_P, d_G, p
```
lukashergt, line 63: These are not the tension statistics...

anesthetic/tension.py
```diff
@@ -0,0 +1,90 @@
...
+
+    logR = statsAB.logZ-statsA.logZ-statsB.logZ
+    samples_stats['logR'] = logR
+    samples_stats.set_label('logR', r'$\log{R}$')
+
+    logI = statsA.D_KL + statsB.D_KL - statsAB.D_KL
+    samples_stats['logI'] = logI
+    samples_stats.set_label('logI', r'$\log{I}$')
+
+    logS = logR-logI
```
lukashergt, line 79: Should we rather go with the calculation via the likelihood, which would work for both `MCMCSamples` and `NestedSamples`? 

```python
logS = statsAB.logL_P - statsA.logL_P - statsB.logL_P
``` 

We can then use `logS = logR - logI` in the tests.

lukashergt (commented):
A few comments inline.

anesthetic/tension.py
DilyOng: Thanks for your comment. This is updated in the newest commit.

anesthetic/tension.py
DilyOng: Updated! Thanks.

tests/test.ipynb
lukashergt: To remove it, do:
```bash
git rm tests/test.ipynb
git commit -m 'remove empty notebook file'
```

anesthetic/tension.py
lukashergt: You updated that specific line, but I still think this would be good to use in the tests, to make sure things work the way we expect them to.

anesthetic/tension.py
lukashergt: I think it's "Logarithm" rather than "Logarithmic", right?

anesthetic/tension.py
williamjameshandley: Let's leave this to a later date.

I've updated the docstrings to include \\mathcals, and in addition I have renamed tension_stats to just stats, since `anesthetic.tension.tension_stats` is a bit repetitive.

williamjameshandley (dismissed):
Hi @DilyOng,

I'm happy for this to be merged now. The tests are 'failing' due to the numpy< 2.0 flag, which won't be resolved until we merge #388, which needs a fastkde upgrade/deprecation

Please press 'squash and merge'. Congrats on your first PR.

williamjameshandley:
@AdamOrmondroyd needs to approve the changes in order for this to be merged

tests/test_tension.py
```diff
@@ -0,0 +1,109 @@
...
+    covB = np.array([[.01, -0.009, 0.01],
+                     [-0.009, .01, -0.001],
+                     [0.01, -0.001, 0.1]])*0.01
+    logLmaxB = 0
+    samplesB = correlated_gaussian(nlive, meanB, covB, bounds, logLmaxB)
+
+    covAB = inv(inv(covA) + inv(covB))
+    meanAB = covAB@(solve(covA, meanA)+solve(covB, meanB))
+    dmeanAB = np.array(meanA)-np.array(meanB)
+    logLmaxAB = -1/2 * dmeanAB@solve(covA+covB, dmeanAB) + logLmaxA + logLmaxB
```
AdamOrmondroyd, line 30: Missing spaces around `@`

tests/test_tension.py
```diff
@@ -0,0 +1,109 @@
...
+
+    logR_std = samples_stats.logR.std()
+    logR_mean = samples_stats.logR.mean()
+    logR_exact = (-1/2*dmeanAB@solve(covA+covB, dmeanAB) -
+                  1/2*slogdet(2*np.pi*(covA+covB))[1] + np.log(V))
+    assert logR_mean - 3 * logR_std < logR_exact < logR_mean + 3 * logR_std
+
+    logS_std = samples_stats.logS.std()
+    logS_mean = samples_stats.logS.mean()
+    logS_exact = d/2 - 1/2*dmeanAB@solve(covA+covB, dmeanAB)
```
AdamOrmondroyd, line 45: Ditto

tests/test_tension.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+def test_tension_stats_compatiable_gaussian():
+    d = 3
+    V = 6.0
+    nlive = 1000
+    bounds = [[-1, 1], [0, 3], [0, 1]]
+
+    meanA = [0.1, 0.3, 0.5]
+    covA = np.array([[.01, 0.009, 0],
```
AdamOrmondroyd, line 14: Missing 0 in 0.01 (there are a few of these, I'm technically on holiday and using my phone so you can find the rest)

AdamOrmondroyd (changes requested):
Comments inline

tests/test_tension.py
lukashergt: done

tests/test_tension.py
lukashergt: done

tests/test_tension.py
lukashergt: done

anesthetic/tension.py
```diff
@@ -0,0 +1,104 @@
...
+        raise ValueError("Shapes of stats_A, stats_B, and stats_AB do not "
+                         "match. Make sure to pass consistent `nsamples`.")
+
+    samples = Samples(index=statsA.index)
+
+    samples['logR'] = statsAB['logZ'] - statsA['logZ'] - statsB['logZ']
+    samples.set_label('logR', r'$\ln\mathcal{R}$')
+
+    samples['logI'] = statsA['D_KL'] + statsB['D_KL'] - statsAB['D_KL']
+    samples.set_label('logI', r'$\ln\mathcal{I}$')
```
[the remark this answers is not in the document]
williamjameshandley: I agree we should go with `I` rather than `lnI` to be consistent with the shannon information (in contrast with the paper).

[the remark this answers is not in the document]
williamjameshandley: @DilyOng could you update the PR accordingly, and I think we'll then be ready to merge.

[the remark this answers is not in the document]
DilyOng: @williamjameshandley In the second-latest push, I have changed all log I to I, and updated all the related lines in anesthetic/tension.py and tests/test_tension.py. All tests have been passed on my local computer. For some reason, the automated tests on Github are not starting. I have tried pushing an empty commit (the last push), but the tests are still not being initialised. Please advise. Thank you.

williamjameshandley:
OK @DilyOng they seem to be running for me -- try correcting the version number in the README and __init__.py back to 2.10.0 to see if it's working now.

DilyOng:
> OK @DilyOng they seem to be running for me -- try correcting the version number in the README and **init**.py back to 2.10.0 to see if it's working now.

@williamjameshandley It's all done - All checks have passed!

williamjameshandley (dismissed):
OK, I think this is ready for merging -- any further suggestions can be presented as additional pull requests.

Many thanks @DilyOng. Please press 'squash and merge'.

williamjameshandley:
(Actually I think @AdamOrmondroyd also has to confirm the changes before merging)

lukashergt (changes requested):
Hi @williamjameshandley and @DilyOng,

I have already been testing this branch out a bit. In that context I realised that the current state is not quite flexible enough. Currently, we consider two datasets A and B separately, and the joint dataset AB. However, there are places where we will want to look at more than two datasets at the same time, e.g. A, B, and C. So we might want a more flexible... Thoughts?

I already have some modifications that could address this locally. Would it be ok for me to highjack this PR with these changes?

lukashergt:
> OK, I think this is ready for merging -- any further suggestions can be presented as additional pull requests.

Our comments crossed. The suggestions in my previous comment would be major change (in the semantic versioning lingo) to the user interface of these tension functions...

anesthetic/tension.py
```diff
@@ -0,0 +1,105 @@
...
+
+    - ``logR``: R statistic for dataset consistency
+
+      .. math::
+        \log R = \log Z_{AB} - \log Z_{A} - \log Z_{B}
+
+    - ``I``: information ratio
+
+      .. math::
+        I = exp(D_{KL}^{A} + D_{KL}^{B} - D_{KL}^{AB})
```
lukashergt, line 20: This is not quite what I meant. I was suggesting the following re-definition of equation (9) in the [Quantifying tensions paper](https://journals.aps.org/prd/pdf/10.1103/PhysRevD.100.043504) (note the lack of `exp`):
```
I = D_A + D_B - D_AB
```
such that equation (10) becomes:
```
logS = logR - I
```

tests/test_tension.py
```diff
@@ -46,14 +46,14 @@ def test_tension_stats_compatible_gaussian():
     logS_exact = d / 2 - dmu_cov_dmu_AB / 2
     assert s.logS.mean() == approx(logS_exact, abs=3*s.logS.std())
 
-    logI_exact = logV - d / 2 - slogdet(2*np.pi*(covA+covB))[1] / 2
-    assert s.logI.mean() == approx(logI_exact, abs=3*s.logI.std())
+    I_exact = np.exp(logV - d / 2 - slogdet(2*np.pi*(covA+covB))[1] / 2)
+    assert s.I.mean() == approx(I_exact, abs=3*s.I.std())
 
-    assert s.logS.mean() == approx(s.logR.mean() - s.logI.mean(),
+    assert s.logS.mean() == approx(s.logR.mean() - np.log(s.I).mean(),
```
lukashergt, line 52: And accordingly this should not have the `np.log`.

anesthetic/tension.py
```diff
@@ -0,0 +1,105 @@
+"""Tension statistics between two datasets."""
+from anesthetic.samples import Samples
+from scipy.stats import chi2
+import numpy as np
+
+
+def stats(A, B, AB, nsamples=None, beta=None):  # noqa: D301
```
AdamOrmondroyd, line 7: @lukashergt @DilyOng for multiple datasets, I think unpacking will neatly handle arbitrary numbers of datasets, something like

```python
def stats(h0, *h1, nsamples=None, beta=None):
    ```h0 = null hypothesis = AB, h1stats = alternative hypothesis = A, B etc```
    ...
    samples['logR'] = h0stats['logZ'] - sum(_h1stats['logZ'] for _h1stats in h1stats)
    ...
```

which can be called `tension.stats(abcde, a, b, c, d, e)`

anesthetic/tension.py
```diff
@@ -0,0 +1,105 @@
...
+
+    - ``p``: p-value for the tension between two samples
+
+      .. math::
+        p = \int_{d-2\log{S}}^{\infty} \chi^2_d(x) dx
+
+    Parameters
+    ----------
+    A : :class:`anesthetic.samples.Samples`
+        :class:`anesthetic.samples.NestedSamples` object from a sampling run
```
AdamOrmondroyd, line 40: This is contradictory: does A have to be `Samples` or `NestedSamples`?

anesthetic/tension.py
AdamOrmondroyd: @lukashergt if we're doing arbitrary numbers of datasets, then we'll need to tweak these equations too, something like

```math
  I = \sum_i {\mathcal{D}_\mathrm{KL}}_i - {\mathcal{D}_\mathrm{KL}}_{H_0}
```
?

AdamOrmondroyd (changes requested):

anesthetic/tension.py
lukashergt: Sidenote: oooh, neat, I didn't know that Markdown can by now handle math input :)

That said, for docstrings I would go for maximal readability even without rendering, so I'd say a simple math example is enough. Leave the rest to papers, or if really necessary, write a dedicated documentation page...?

anesthetic/tension.py
lukashergt: Yes, that's about what I have. I call them `joint` and `separate`, which I find a bit more descriptive...

anesthetic/tension.py
DilyOng: @lukashergt Thank you very much for clarifying. I have changed the relevant lines.

tests/test_tension.py
DilyOng: @lukashergt Updated.

anesthetic/tension.py
DilyOng: @AdamOrmondroyd A can be either `anesthetic.samples.Samples` or `anesthetic.samples.NestedSamples`. If `NestedSamples` are given, the function gets `Samples` using `Samples.stats()`. 

From line 71 to 82
```
    if set(columns).issubset(A.drop_labels().columns):
        statsA = A
    else:
        statsA = A.stats(nsamples=nsamples, beta=beta)
    if set(columns).issubset(B.drop_labels().columns):
        statsB = B
    else:
        statsB = B.stats(nsamples=nsamples, beta=beta)
    if set(columns).issubset(AB.drop_labels().columns):
        statsAB = AB
    else:
        statsAB = AB.stats(nsamples=nsamples, beta=beta)
```

DilyOng:
> Hi @williamjameshandley and @DilyOng,
> 
> I have already been testing this branch out a bit. In that context I realised that the current state is not quite flexible enough. Currently, we consider two datasets A and B separately, and the joint dataset AB. However, there are places where we will want to look at more than two datasets at the same time, e.g. A, B, and C. So we might want a more flexible... Thoughts?
> 
> I already have some modifications that could address this locally. Would it be ok for me to highjack this PR with these changes?

@lukashergt @williamjameshandley  My apologies for the delayed reply. This is indeed a very good suggestions and we should definitely extend the function to include more than 2 datasets comparison simultaneously. However, this will be a major change and likely invite more discussion. Shall we merge the current branch first, so anesthetic's users can start using the tension statistics calculator, while we upgrade it using another pull request? 

I am going to upgrade the tension statistics calculator to introduce the correction factor for polychord's nprior and ndiscard values. I have already written the code for this modification locally, but as this is also a major change, it may be better to do another PL. 

@lukashergt @williamjameshandley  What do you think?

lukashergt (approved):
Hi all,

a few changes:
* `tension_stats` now can take more than 2 datasets (with argument packaging [as suggested](https://github.com/handley-lab/anesthetic/pull/333#discussion_r1942533762) by @AdamOrmondroyd).
* Inputs to `tension_stats` are expected to be `Samples` as returned by `NestedSamples.stats`, i.e. with columns `['logZ', 'D_KL', 'logL_P', 'd_G']`.
  * Hence, I dropped the option of passing `NestedSamples` directly. If you want, I can bring this back, but I would always prefer calculating `stats` and `tension_stats` in separate steps, so I don't really see the need for the alternative option. Recomputing `stats` every time is quite time consuming, and when investigating more than a single dataset combination, then you typically don't want to do that.
* Introduced `tension` column, which expresses the tension in terms of numbers of sigma.

With this, I'd be ready to merge, or we could additionally incorporate the `nprior` correction @DilyOng [mentioned](https://github.com/handley-lab/anesthetic/pull/333#issuecomment-2818918081).

The more general adaption of a `Samples.stats` method I'd leave for a future PR.

williamjameshandley (changes requested):
Could we remove the joint_f functionality for now -- I think this is a more fundamental adjustment to nested samples and the logZ calculation, which we should do as a seperate PR rather than specific to the tension calculation.

lukashergt:
Hi all, I'm writing on a section myself, and I prefer a slightly different notation to what has been used before, making things more consistent/symmetric. See the slightly altered equation (12) below, which changes the sign of $I$ in equation (14), making it match the Occam penalty equation (9)... Thoughts? Especially @williamjameshandley since you might dislike that this does not match your previous papers...?

image

README.rst
```diff
@@ -2,7 +2,7 @@
 anesthetic: nested sampling post-processing
 ===========================================
 :Authors: Will Handley and Lukas Hergt
-:Version: 2.10.3
+:Version: 2.10.4
```
lukashergt, line 5: This should be a minor (not micro) version change.

lukashergt (changes requested):

README.rst
lukashergt: done

lukashergt (approved):
Happy with the current state, but see [my previous comment](https://github.com/handley-lab/anesthetic/pull/333#issuecomment-3343858342). If everyone agrees I can quickly implement the suggested change there.

The CI still has some hick-ups, if somebody could review and approve #408 then that should hopefully vanish.

anesthetic/tension.py
```diff
@@ -0,0 +1,113 @@
...
+        samples["I"] = separate_stats["D_KL"] - joint_stats["D_KL"]
+        samples.set_label("I", r"$\mathcal{I}$")
+
+    samples["logS"] = joint_stats["logL_P"] - separate_stats["logL_P"]
+    samples.set_label("logS", r"$\ln\mathcal{S}$")
+
+    samples["d_G"] = separate_stats["d_G"] - joint_stats["d_G"]
+    samples.set_label("d_G", r"$d_\mathrm{G}$")
+
+    p = chi2.sf(samples["d_G"] - 2 * samples["logS"], df=samples["d_G"])
```
AdamOrmondroyd, line 106: I believe this is $1-p$ in [arXiv:1902.04029](https://arxiv.org/abs/1902.04029), I assume this is intentional?

lukashergt:
Note that we already discussed `logI` vs `I` in a [previous comment](https://github.com/handley-lab/anesthetic/pull/333#discussion_r1778317540). There I also already pointed out the analogy to the Occam equation, but for that analogy to hold (including the minus sign) we need to change our current definition of `I` and multiply it by `-1`.

For the `logI` vs `I` question, see also Shannon information in https://arxiv.org/pdf/1903.06682.

Regarding \log vs \ln, we use `logZ` as a handle, but `'$\ln\mathcal{Z}$'` as the label, so I'd match that here...?

AdamOrmondroyd:
pls decide if you want I or log I then you may have a lgtm

DilyOng:
I had a discussion with @williamjameshandley and have come to the conclusion that according to this paper 1910.07820v1, equation 5 and 6, we prefer 

- `logI` over `I`
- `logI = D_A + D_B - D_AB`
- '\log' over '\ln'

I personally think it is better to call 'tension' as 'sigma' because it is indeed a p-value sigma. Readers might find it easier to understand its nature with the name 'sigma' than the name 'tension'.

I've implemented the changes mentioned above. Can @williamjameshandley and @AdamOrmondroyd please read and approve (if you are happy with the newest version)? This branch needs both of your approval to be merged. Thank you very much!

lukashergt (changes requested):
Hmm, ok, I'm a bit confused now. I view the discussion around the minus sign separate from the `I` vs `logI` debate.

I thought we had [already agreed on `I`](https://github.com/handley-lab/anesthetic/pull/333#discussion_r1938583686) and in contrast to the paper (for good reason I believe)... I'd be happy to hear counter arguments, but just pointing to an equation in a paper which I consider to be a typo does not convince me...

But it's a bit unfair for @DilyOng to get caught in the cross-fire between me and Will. 
@williamjameshandley, would you have time for a video call? 

---

### `I` vs `logI`

Arguments in favour of `I`:

* $"I"$ is the typical symbol for an entropy (which is an _additive_ quantity). 
* [[1903.06682] Quantifying dimensionality: Bayesian cosmological model complexities](https://arxiv.org/pdf/1903.06682)
  - eq. 3: Shannon information $\quad I(\theta)=\log\frac{{P}(\theta)}{\pi(\theta)}$ 
  - eq. (5), KL-divergence: $\quad{D} = \langle\log\frac{{P}}{\pi}\rangle  = \langle I \rangle$
  - eq. (6), model complexity: $\quad\frac{\hat{d}}{2} = I(\hat\theta) - \langle I \rangle$
  - Note how we write $\ {D}$, not $\ \log{D}$, and $\ I$, not $\ \log{I}$.  
* [[2102.11511] Bayesian evidence for the tensor-to-scalar ratio and neutrino masses](https://arxiv.org/pdf/2102.11511)
  - eq. (6), Occam equation: $\quad\log{Z} = \langle\log{L}\rangle - {D}$
  - Note the `log` in $\log{Z}$ and in $\langle\log{L}\rangle$, but not in ${D}$.
* [[1902.04029] Quantifying tensions in cosmological parameters](https://arxiv.org/pdf/1902.04029):
  - eq. (8): $\quad\log{R} = \log{Z}_{AB} - \log{Z}_A - \log{Z}_B$
* Plug in Occam equation: $\quad\log{R} = \log{S} - \left( {D}_{AB} - {D}_A - {D}_B \right)$
  - Note the symmetry to the Occam equation.
  - Note that $\quad\left( {D}_{AB} - {D}_A - {D}_B \right)\quad$ would be a relative entropy if A and B are independent. This is what I would like to call $\mathcal{I}$ drawing on my first point above.
  - On the other hand, $\quad\exp\left( {D}_{AB} - {D}_A - {D}_B \right)\quad$ is not a relative entropy, not even for independent A and B, and this is what is called $\mathcal{I}$ in [[1902.04029]](https://arxiv.org/pdf/1902.04029) and [[1910.07820]](https://arxiv.org/pdf/1910.07820), which I always assumed to be a typo...


---

### Minus sign debate

This comes down to the question whether we want to write 

* Option 1: $\quad\log{R} = \log{S} + \left( D_A + D_B - D_{AB} \right)$ 

or 

* Option 2: $\quad\log{R} = \log{S} - \left( {D}_{AB} - {D}_A - {D}_B \right)$ 

Where the quantity in parenthesis is `I` (or `logI`, see above). The difference comes down to a minus sign in the definition of `I` (or `logI`).

I prefer option 2, because (a) it is analogous to the Occam equation and (b) we are consistently subtracting the separate quantities from the joint quantity, i.e. we'd have consistently:
* $\log R = \log Z_{AB} - \log Z_A - \log Z_B$
* $\log S = \log L_{AB} - \log L_A - \log L_B$
* $\mathcal{I} = D_{AB} - D_A - D_B$ 

---

### `'\ln'` vs `'\log'`

I don't have a _strong_ opinion on whether we are using `'\ln'` or `'\log'` for our plotting labels in general, but I would prefer being consistent. Anesthetic's `logZ` is currently set up with `'\ln'`, so I feel like we should be using that here too.

> R.e. \log vs \ln: with a quick grep anesthetic currently mixes them, personally I'd prefer log since we say "log-likelihood" on a daily basis, but this isn't an issue for here

What @AdamOrmondroyd is catching in terms of `\log` with his `grep` call are the maths equations in the docstrings. All labels used for plotting (for log-evidence, log-likelihood, etc.) we consistently use `\ln`. 

Maths-wise it is irrelevant whether we use $\log$ or $\ln$ or $\log_2$ or $\log_{10}$, so for the underlying equations I like to use the very general `log`. But the moment we actually compute quantities with anesthetic, we have to make a choice about which base we use, and that is where I prefer being precise and explicit by providing that information in the labels, hence `'\ln\mathcal{Z}'`.

---

> I personally think it is better to call 'tension' as 'sigma' because it is indeed a p-value sigma. Readers might find it easier to understand its nature with the name 'sigma' than the name 'tension'.

Yes, I like that :)

PR #335, opened 2023-09-08
Changes for `pandas 2.1`

### LINKED ISSUE #336: Incompatibility with `pandas 2.1`
**Describe the bug**
The test suite fails due to incompatibility with `pandas 2.1`

**To Reproduce**
`pip install anesthetic` and run tests.

**Expected behavior**
Tests all run and pass.

**Screenshots**

**Additional context**

AdamOrmondroyd:
# Description

`pandas 2.1` has introduced further API changes. I think it will be necessary in future to restrict `anesthetic` to a minor `pandas` version

`pandas 2.1` also does not support python 3.8

Fixes #336

# Bugs:

#### `tm.close()` removed
- replace this with `yield; plt.close("all")`
#### `temporary_seed` in `compress_weights`
- needs to be explicitly cast to `int`
#### Issue with loc and masks in groupby stats tests
- have a temporary fix but not convinced I've gotten to the bottom of it
- see `tests/test_samples.py::test_groupby_stats()`
#### `python 3.8`
- `pandas2.1` isn't compatible with python3.8, so remove this requirement and change other tests that use it. I have used python3.11 as this is the fastest available and I see no reason not to
#### documentation
- I think I need Lukas' expertise on this one...

# (new) warnings
#### label-less casting name to label
- Added cast to type of other labels
#### adding labels as names 
recall when we made the labels of a `LabelledDataFrame` the names of the sliced `Series`
```python
if l[s.name]:
    s.name = l[s.name]
```
this complains that:
```
/.../anesthetic/labelled_pandas.py:56: FutureWarning: Series.__getitem__ treating keys as positions is deprecated. In a future version, integer keys will always be treated as labels (consistent with DataFrame behavior). To access a value by position, use `ser.iloc[pos]`
```
I *think* that we are using the newer style of behaviour, so I don't think there is a new problem here. Time will tell...

# Checklist:


AdamOrmondroyd:
@lukashergt if you have any ideas how to fix the documentation I'm all ears... error is only happening with pandas2.1

anesthetic/plotting/__init__.py
```diff
@@ -1,5 +1,2 @@
-"""Plotting public API.
-
-.. autoclass:: anesthetic.plotting.PlotAccessor
-"""
```
AdamOrmondroyd, line 4: I think this is responsible for the issues with building documentation

tests/test_samples.py
```diff
@@ -1517,47 +1517,47 @@ def test_groupby_stats():
         if 'chain' not in col:
             for chain in [1, 2]:
                 mask = mcmc.chain == chain
-                assert_allclose(mcmc.loc[mask, col].mean(),
+                assert_allclose(mcmc.loc[mask][col].mean(),
```
AdamOrmondroyd, line 1520: I worry there is some underlying change here that needs to be understood

anesthetic/plotting/__init__.py
AdamOrmondroyd: Examples have been added to the docstring for the `pandas2.1` `PlotAccessor`, which is where `pd` is required

AdamOrmondroyd: > Examples have been added to the docstring for the `pandas2.1` `PlotAccessor`, which is where `pd` is required

Removing these lines have fixed the compilation locally, but obviously this isn't viable. I will try removing those lines with the `process_docstring`

I've also noticed there are extra errors happening in the CI I'm not getting locally

docs/source/anesthetic.read.rst
```diff
@@ -40,6 +40,7 @@ anesthetic.read.hdf module
 .. automodule:: anesthetic.read.hdf
    :members:
    :undoc-members:
+   :show-inheritance:
```
lukashergt, line 43: This line needs to go, otherwise it will cause autodoc to try and find documentation in parents that do not have any docs...

docs/source/anesthetic.rst
```diff
@@ -108,5 +108,6 @@ anesthetic.weighted\_pandas module
 .. automodule:: anesthetic.weighted_pandas
    :members:
    :undoc-members:
+   :show-inheritance:
```
lukashergt, line 111: This line needs to go, otherwise it will cause autodoc to try and find documentation in parents that do not have any docs...

lukashergt (commented):
Sorry, had a deadline this morning, so was a bit pre-occupied.

> @lukashergt if you have any ideas how to fix the documentation I'm all ears... error is only happening with pandas2.1

We have to remove some of the doc inheritance stuff. The problem is that pandas doesn't document all their classes, so if we inherit from those, we mustn't tell autodoc to try and inherit their non-existent docs...

anesthetic/plotting/__init__.py
lukashergt: I've modified the `docs/source/conf.py` to preload pandas such that the `pd` shouldn't be a problem anymore.

> I've also noticed there are extra errors happening in the CI I'm not getting locally

The full command for building the docs matching the CI is:
```bash
make html SPHINXOPTS="-W --keep-going -n"
```
This ensures that we catch whenever we think we are adding external links, but really its not finding anything. Otherwise things pass silently and we are slowly piling on broken links...

lukashergt:
Would be good if someone could double check that https://github.com/handley-lab/anesthetic/pull/335/commits/43c5be5e0bd9af244d3fcfd34b1ff95504aa1871 is a solid change.

anesthetic/plotting/__init__.py
AdamOrmondroyd: Ah yes of course, thanks! I'll use this in future

AdamOrmondroyd:
> Would be good if someone could double check that [43c5be5](https://github.com/handley-lab/anesthetic/commit/43c5be5e0bd9af244d3fcfd34b1ff95504aa1871) is a solid change.

`anesthetic.plotting` includes changes for both labels and weights, so perhaps it should be in `anesthetic.labelled_pandas.py` also?

docs/source/anesthetic.rst
AdamOrmondroyd: Before I was using `make html SPHINXOPTS="-W --keep-going -n"` I thought you just had a sixth sense for broken docs

tests/test_samples.py
lukashergt: I think the problem here is that `mask` is not a simple boolean array. Would it work to leave things as they are and only change the mask?
```python
mask = (mcmc.chain == chain).to_numpy()
```

AdamOrmondroyd: yes, but then I'm puzzled why the mask in the loop before is fine without `.to_numpy()`, even though it is also weighted

lukashergt: Agreed, seems slightly inconsistent, but I think quite generally it is deemed cleaner to use a list/array as a boolean mask rather than a pandas series/dataframe which includes an index.

AdamOrmondroyd: I have to say I was surprised it worked at all

AdamOrmondroyd: I've changed the other mask to match too

AdamOrmondroyd:
_If the tests pass_ I think this is ready to go

AdamOrmondroyd:
The only other thing is inevitably anesthetic will break again when pandas 2.2 releases. While it is useful that the tests flag this up very dramatically, perhaps it isn't the best way of going forward since it means that `pip install anesthetic` doesn't work for a while. I wonder if restricting matplotlib and pandas to their current minor version in the `pyproject.toml` + a "use latest versions" test might be sensible?

AdamOrmondroyd:
I almost changed this to a patch version increase, but I think the new inclusion of `PlotAccessor` in `weighted_pandas` and `labelled_pandas` is an API change

tests/test_plot.py
```diff
@@ -97,18 +97,19 @@ def test_make_1d_axes():
...
     g00 = grid[0, 0]
     fig, axes = make_1d_axes(paramnames, subplot_spec=g00)
-    assert g00 is axes[0].get_subplotspec().get_topmost_subplotspec()
+    assert g00 is axes.iloc[0].get_subplotspec().get_topmost_subplotspec()
 
     # Check gridspec kwargs
     fig, axes = make_1d_axes(paramnames, gridspec_kw=dict(wspace=0.1))
-    ws = axes[0].get_subplotspec().get_gridspec().get_subplot_params().wspace
+    ws = (axes.iloc[0].get_subplotspec()
+          .get_gridspec().get_subplot_params().wspace)
```
lukashergt, line 112: Ugly line break, how about this?

```python
    ws = axes['A'].get_subplotspec().get_gridspec().get_subplot_params().wspace
```

tests/test_plot.py
```diff
@@ -180,7 +181,8 @@ def test_make_2d_axes_inputs_outputs():
 
     # Check gridspec kwargs
     fig, axes = make_1d_axes(paramnames_x, gridspec_kw=dict(wspace=0.1))
-    ws = axes[0].get_subplotspec().get_gridspec().get_subplot_params().wspace
+    ws = (axes.iloc[0].get_subplotspec()
+          .get_gridspec().get_subplot_params().wspace)
```
lukashergt, line 185: Ugly line break, how about this?

```python
    ws = axes['A'].get_subplotspec().get_gridspec().get_subplot_params().wspace
```

lukashergt (commented):
> The only other thing is inevitably anesthetic will break again when pandas 2.2 releases. While it is useful that the tests flag this up very dramatically, perhaps it isn't the best way of going forward since it means that `pip install anesthetic` doesn't work for a while. I wonder if restricting matplotlib and pandas to their current minor version in the `pyproject.toml` + a "use latest versions" test might be sensible?

Let's wait for @williamjameshandley to comment on #343, too. But if we decide to do that, would we want to implement that as part of this PR or in a separate PR?

lukashergt:
> Let's wait for @williamjameshandley to comment on #343, too. But if we decide to do that, would we want to implement that as part of this PR or in a separate PR?

Ah, nevermind, already building in #344...

tests/test_plot.py
AdamOrmondroyd: It's too long to get through flake8. I didn't want to put it in!

AdamOrmondroyd: sorry I've just read this properly - sure

tests/test_plot.py
AdamOrmondroyd: done

AdamOrmondroyd:
> > Let's wait for @williamjameshandley to comment on #343, too. But if we decide to do that, would we want to implement that as part of this PR or in a separate PR?
> 
> Ah, nevermind, already building in #344...

yes, definitely its own PR

lukashergt (approved):

lukashergt:
I keep forgetting the best course of action when changing required tests... @williamjameshandley, this PR drops python 3.8 checks and instead updates to 3.11. How best to handle the expected but not existing 3.8 checks?

williamjameshandley:
OK that should no longer require 3.8 to pass. I guess my only query with this is why this is 2.4.0 rather than 2.3.5?

AdamOrmondroyd:
> I almost changed this to a patch version increase, but I think the new inclusion of `PlotAccessor` in `weighted_pandas` and `labelled_pandas` is an API change

^^

williamjameshandley:
Is this an API change? -- `weighted_pandas` and `labelled_pandas` objects currently have `PlotAccessor`s in `.plot`, we've just changed how these are added to them.

Why was this necessary?

AdamOrmondroyd:
It was motivated to make the documentation for `weighted_pandas` work.

Perhaps I am mistaken, but I don't think `labelled_pandas` previously used the anesthetic `PlotAccessor`, which I believe it should? This is the API change I refer to, perhaps this is a change for a separate quick PR

AdamOrmondroyd:
Perhaps not a _quick_ PR - I've just noticed that `test_labelled_pandas.py` doesn't actually test plotting behaviour at all

AdamOrmondroyd:
At the moment, `labelled_pandas` does not use the labels for plotting:

`Samples` work properly:
```python
from matplotlib import pyplot as plt
import anesthetic as ac
ns = ac.read_chains("tests/example_data/pc")
ns.plot.scatter('x0', 'x1')
plt.show()
```
Screenshot 2023-09-28 at 15 35 44

but `LabelledDataFrame` does not

```python
l = ac.labelled_pandas.LabelledDataFrame(ns)
l.plot.scatter('x0', 'x1')
plt.show()
```
Screenshot 2023-09-28 at 15 35 19

AdamOrmondroyd:
I've looked a bit closer, and I don't think this change to `labelled_pandas` fixes this properly, so I will revert this change and make a new issue

AdamOrmondroyd:
Now this PR is purely changes for pandas 2.1, think I'm happy with it now

lukashergt (approved):
> I've looked a bit closer, and I don't think this change to labelled_pandas fixes this properly, so I will revert this change and make a new issue

So there might be some stuff in `Samples` that maybe should already be part of `LabelledDataFrame`...? Yes, would be good to track that in an issue and link to this PR.

PR #339, opened 2023-09-15
Changes for `matplotlib3.8`

AdamOrmondroyd:
# Description

[`matplotlib3.8`](https://matplotlib.org/stable/users/prev_whats_new/whats_new_3.8.0.html) released yesterday morning.

- [x] `Axes._get_lines.prop_cycler` has been removed, instead use `Axes._get_lines.get_next_color()` (inevitable with private member usage)
- [x] `plt.hexbin` now plots bins ≥`mincnt`, not >`mincnt`, which changes the behaviour with the default of 0 (corresponding to None in the signature). I have used `np.finfo.tiny` to correct this.
- [x] `tcolors` is deprecated, I have lifted the code used to simulate it in `matplotlib` to restore the test in `test_contour_plot_2d()`.
- [x] `samples.plot.[fast]kde_2d()` now only adds two collections to the axes
- [ ] plt.contour colors are out of whack (one weird extra one and the others are slightly different)

The last three are all symptoms of `plt.contour` returning `QuadContourSet` (one per level, so usually two for a contour plot) rather than `PathCollection`. 

I'm completely stumped by the last one.

Fixes # (issue)

# Checklist:


tests/test_samples.py
```diff
@@ -1245,7 +1245,7 @@ def test_samples_dot_plot():
     assert len(axes) == 2
 
     axes = samples.plot.kde_2d('x0', 'x1')
-    assert len(axes.collections) == 5
+    assert len(axes.collections) == 2
```
AdamOrmondroyd, line 1248: This test will fail on mpl<3.8. Is this a test we really need to do? The number of collections feels like matplotlib just doing matplotlib things

anesthetic/plotting/_matplotlib/core.py
```diff
@@ -131,6 +131,8 @@ def __init__(self, data, x, y, C=None, **kwargs) -> None:
             C = '__weights'
             data[C] = data.get_weights()
             kwargs['reduce_C_function'] = np.sum
+            if 'mincnt' not in kwargs:
+                kwargs['mincnt'] = np.finfo(np.float64).tiny
```
AdamOrmondroyd, line 135: alias for `smallest_normal` but I like `tiny`

tests/test_samples.py
williamjameshandley: I agree (although 2 makes more sense here than 5...). Is there anything explicit we can test to confirm that it has plotted something?

AdamOrmondroyd: Check it's more than 0?

williamjameshandley: That would do

AdamOrmondroyd:
An extra level of `.collections[0]` restores the original test, as the first gets at the `QuadContour` object, and the second gets us to the deprecated API we were using to test before.

The plots at least look identical to the previous matplotlib

AdamOrmondroyd:
Still stumped 😢 using the old api will be deprecated in a couple of minor versions so I'd prefer not to rely on it.

I just can't work out what the five colours from `QuadContourSet.get_edgecolor()` actually correspond to

lukashergt:
> Still stumped 😢 using the old api will be deprecated in a couple of minor versions so I'd prefer not to rely on it.

Agreed, how about we add a matplotlib version check to the tests and test both old and new API. In the actual code I'd find that really annoying, but if we limit it to the tests that should be ok, right?

AdamOrmondroyd:
> > Still stumped 😢 using the old api will be deprecated in a couple of minor versions so I'd prefer not to rely on it.
> 
> Agreed, how about we add a matplotlib version check to the tests and test both old and new API. In the actual code I'd find that really annoying, but if we limit it to the tests that should be ok, right?

But what do we do in two matplotlib's time? My concern is I don't understand the `QuadContourSet` colours

lukashergt:
> But what do we do in two matplotlib's time? 

Keep adding to the tests and we could say anything older than 2 or 3 minor matplotlib versions get's scratched.  

> My concern is I don't understand the `QuadContourSet` colours

`QuadContourSet.get_edgecolor()` should return the colours of the `contour` lines (so it should be one more than the number of levels). `QuadContourSet.get_facecolor()` should return the colours of the `contourf` faces (as many as there are levels). In 3.7 each face and each line was its own `PathCollection` (that's where the number 5 came from, 2 faces and 3 lines). Now in 3.8 it seems these have been grouped into two `QuadContourSet`, one for `contourf` and one for `contour`.

AdamOrmondroyd:
got it, thanks Lukas 😄 hopefully that's this PR done

lukashergt (approved):
Thanks a bunch @AdamOrmondroyd! Please squash and merge.

AdamOrmondroyd:
Whoops! Left that in from while I was remembering how to check the version

PR #342, opened 2023-09-20
Ignore "You are using [hist/kde] as a plot kind" warnings

AdamOrmondroyd:
# Description

Since the `UserWarning`s were added to prevent users from unintentionally using `kde` instead of `kde_1d` etc, the test suite has been (correctly) flagging up these warnings, affecting quality of life

Tests obviously won't pass until either #341 or #339 are merged.

Fixes # (issue)

# Checklist:


lukashergt:
These are warnings that I'd argue _should_ pop up in the tests, so they should not be suppressed, and they should "affect our quality of life"...

The warnings are bound to go away eventually with #320, but until then they are a good reminder that we should eventually clean this up fully.

AdamOrmondroyd:
I agree they will go away with #320 as it stands, but should they?  Unless we actually make hist and kde inaccessible through anesthetic they will still "look odd"

lukashergt:
> I agree they will go away with #320 as it stands, but should they? Unless we actually make hist and kde inaccessible through anesthetic they will still "look odd"

Agreed, it is probably good to keep them, but that means that rather than just filtering them out of the tests, we should make them part of the tests, i.e. make use of `with pytest.warns(UserWarning):`.

I see two ways of implementing this:
1. Change all the tests that currently are filtering the warnings out to instead make use of `pytest.warns` in the appropriate cases.
2. Keep the filtering as is, but add specific tests for the warnings.

What would others think best?

AdamOrmondroyd:
Seeing as the filters don't appear that many times in the test suite, I think 1

lukashergt (approved):
This looks good, thanks @AdamOrmondroyd. Please, squash and merge if you think it is ready. And please feel free to hit "request review" to get a faster reaction from me.

PR #344, opened 2023-09-24
Test latest dependencies

### LINKED ISSUE #343: Restrict upper limit of pandas and matplotlib + "use latest versions" test
Every time a minor update to  pandas or matplotlib is released, inevitably anesthetic breaks. It then takes time to make the necessary changes, which is never as simple as I expect, so we then release a patch to the pyproject.toml which restricts pandas/matplotlib to the previous version, all the while `pip install anesthetic` is broken, which doesn't look good.

I propose that we restrict matplotlib and pandas to minor versions which we know work, which once #339 and #335 are finished, will be
```toml
dependencies = [
    "scipy",
    "numpy",
    "pandas>=2.0.0,<2.2.0",
    "matplotlib>=3.6.1,<3.9.0",
]
```
numpy and scipy have behaved in the time that I've been working on this project so I think just leave them alone.

To make us aware of new versions, I think we should have a test which tries the latest versions, but does not necessarily need to pass for a PR to be merged.

AdamOrmondroyd:
# Description

Add upper limits to `pandas` and `matplotlib` versions, and add an additional test which uses the most recent releases.

Also increase the README python version to 3.8+

Fixes #343 

# Checklist:


AdamOrmondroyd:
First pass done. Will the test become required if this PR was merged? Is there a way to avoid that?

lukashergt:
> First pass done. Will the test become required if this PR was merged? Is there a way to avoid that?

This can be configured in the GitHub settings.

AdamOrmondroyd:
Once #339 is done I'll update this accordingly

@williamjameshandley your gymnastics with version numbers in #335 has confused me even more how the test requirements work...

AdamOrmondroyd:
just spotted a mistake in the README which imo fits in with the theme of this PR

README.rst
```diff
@@ -96,7 +96,7 @@ Dependencies
 
 Basic requirements:
 
-- Python 3.6+
+- Python 3.8+
```
lukashergt, line 99: We are no longer testing python 3.8, so we should best set this to 3.9+.

.github/workflows/CI.yaml
```diff
@@ -163,3 +163,26 @@ jobs:
...
+      - name: Set up latest stable Python 3
+        uses: actions/setup-python@v4
+        with:
+          python-version: 3.x
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install tomli
+          eval "python -m pip install $(./bin/latest_dependencies.py)"
```
lukashergt, line 180: Since this line is the crucial part of this job, how about we turn this line into its own step (separate from pip/tomli/pytest/etc. updates) such that when reviewing it will be faster to check which versions were being installed?

lukashergt (changes requested):
>> There is an issue with this idea, which is that if the next minor pandas/matplotlib releases and doesn't cause any problems, there's nothing to notify us that we should increment the anesthetic requirements
> 
> How about we add a test that runs after the latest dependencies test and checks whether the latest pandas and matplotlib minor versions are greater than our listed dependency?

As mentioned in #343, let's add another job that checks whether our matplotlib and pandas version requirements are up to date.

.github/workflows/CI.yaml
lukashergt: "this line" referring to line 180 only...

README.rst
AdamOrmondroyd: that was silly

.github/workflows/CI.yaml
AdamOrmondroyd: I'll change the minimum dependency check too

.github/workflows/CI.yaml
```diff
@@ -151,15 +151,54 @@ jobs:
...
+
+      - name: Check anesthetic dependencies are up to date
+        run: |
+          latest=$(pip freeze)
+          python -m pip uninstall -y -r <(pip freeze)
+          python -m pip install -e .
+          current=$(pip freeze)
+          if [ $latest != $current ] then
+             exit 1
+          fi
```
lukashergt, line 200: Make this a job rather than a step? Then we directly know whether new versions exist and break or exist and work. Otherwise we will have to look into the tests to find out the reason for the test failure...

AdamOrmondroyd: If you like,  but the setup will have to be repeated

lukashergt: Is there maybe a way of making it separate jobs without repeating the setup? Not sufficiently fluent in CI stuff...

AdamOrmondroyd: Pass. I'm currently concerned with making `if [ $latest != $current ]` work properly, as I haven't done very much in bash myself

AdamOrmondroyd: turns out single vs double quotes is very important

AdamOrmondroyd:
Problem is this doesn't tell us where the problem is

lukashergt:
> Problem is this doesn't tell us where the problem is

Yeah, a check with pytest would probably be more helpful for reviewing, i.e. install with latest versions and then use pytest to make a version check against the versions listet in the dependencies.

AdamOrmondroyd:
It would have to be separate from the rest of the tests, as this is not required to pass

AdamOrmondroyd:
I'm quite a bit happier with this version. Currently includes an explicit list of `[pandas, matplotlib]`, @lukashergt would you also like this to be automated? I could probably recycle bin/latest_dependencies.py

AdamOrmondroyd:
Now automated to include all basic dependencies. Does come at the expense of bin/latest_dependencies.py involving a little more gymnastics

.github/workflows/CI.yaml
```diff
@@ -151,15 +151,66 @@ jobs:
...
+          python -m pip install tomli requests
+
+      - name: Install anesthetic
+        run: python -m pip install -e ".[test]"
+
+      - name: Check anesthetic dependencies are up to date
+        run: python -m pytest ./bin/check_up_to_date.py
+
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v1
```
lukashergt, line 216: As above, do we need to upload to codecov?

lukashergt (approved):
Nice, just a question inline whether we could avoid the uploading to codecov, if not, then I'm happy with this.

.github/workflows/CI.yaml
```diff
@@ -151,15 +151,66 @@ jobs:
...
+      - name: Install pytest and anesthetic
+        run: |
+          python -m pip install pytest
+          python -m pip install --no-deps -e .
+
+      - name: Test with pytest
+        run: python -m pytest
+
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v1
```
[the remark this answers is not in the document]
AdamOrmondroyd: I should have flagged this earlier as I wasn't sure myself. @williamjameshandley what do you think?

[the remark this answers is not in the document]
AdamOrmondroyd: Are all the jobs under CI separate or one in the settings?

AdamOrmondroyd:
Fyi python 3.12 released a couple of days ago, and the tests are all passing fine (no new warnings anyway!). There's also a new m1 mac github runner we might want to run tests on https://github.blog/2023-10-02-introducing-the-new-apple-silicon-powered-m1-macos-larger-runner-for-github-actions/

.github/workflows/CI.yaml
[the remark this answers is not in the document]
lukashergt: Separate. Each job can be added/removed individually.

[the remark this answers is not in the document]
AdamOrmondroyd: lgim then

.github/workflows/CI.yaml
```diff
@@ -151,15 +151,60 @@ jobs:
...
+      - name: Install minimum dependencies
+        run: eval "python -m pip install $(./bin/min_dependencies.py)"
+      - name: Install anesthetic
+        run: python -m pip install -e ".[test]"
 
       - name: Test with pytest
         run: python -m pytest --cov=anesthetic tests
 
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v1
```
lukashergt, line 167: Should we remove the codecov upload for `minimum-dependencies`, too? Not really what this job is about either, right?

.github/workflows/CI.yaml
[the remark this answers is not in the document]
lukashergt: > lgim 

???

.github/workflows/CI.yaml
AdamOrmondroyd: I disagree, since it is necessary that anesthetic works with the lowest versions we allow, and a PR should not be merged unless this is true

.github/workflows/CI.yaml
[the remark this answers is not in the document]
AdamOrmondroyd: "let's get it merged", which I think is what @williamjameshandley thought I meant by "lgtm" (looks good to me, which I learned from pandas)

.github/workflows/CI.yaml
lukashergt: But codecov has nothing to do with whether anesthetic works or not, that is what pytest is for. Codecov should definitely be in the pip and conda jobs (or at least in one of them), but the coverage does not really depend on versions (unless we specifically add version related conditional statements to the code, which we don't). Besides, the `minimum-dependencies` job does not even include extra dependencies, so it doesn't do a full code coverage anyhow...

lukashergt: _If_ the `minimum-dependencies` code coverage were to fail, while the pip code coverage passes, then we wouldn't even know about it, because as long as the pip jobs cover every line at least once, codecov will mark 100% coverage. But I don't even think it is possible for code coverage to be different for the pip jobs and the minimum job...

AdamOrmondroyd: Some of the tests are now different for matplotlib≥3.8, and at one point in beta some of the anesthetic plotting code was different for pandas 1 vs 2. cov would tell us when these lines are no longer needed right?

lukashergt: > Some of the tests are now different for matplotlib≥3.8, 

Code in `tests/` doesn't matter for code coverage, only code in `anesthetic/` matters.

> and at one point in beta some of the anesthetic plotting code was different for pandas 1 vs 2. cov would tell us when these lines are no longer needed right?

In a situation like that we would indeed need the code coverage in `minimum-dependencies` for as long as we wanted to be backwards compatible. However, we already have pandas>=2.0.0, so this shouldn't be an issue, here.

In fact, I think this might be an argument in favour of not having code coverage in `minimum-dependencies`, to flag up once parts of code are no longer needed for newer versions. Currently we wouldn't know if parts of code were no longer needed by newer versions.

AdamOrmondroyd: > > and at one point in beta some of the anesthetic plotting code was different for pandas 1 vs 2. cov would tell us when these lines are no longer needed right?
> 
> In a situation like that we would indeed need the code coverage in `minimum-dependencies` for as long as we wanted to be backwards compatible. However, we already have pandas>=2.0.0, so this shouldn't be an issue, here.
> 
> In fact, I think this might be an argument in favour of not having code coverage in `minimum-dependencies`, to flag up once parts of code are no longer needed for newer versions. Currently we wouldn't know if parts of code were no longer needed by newer versions.

I guess it depends how cut-throat we want to be with older packages. I suppose the pandas2 difference was because pandas2 hadn't actually released yet, and once it did, we dropped support for pandas1.

AdamOrmondroyd: > > Some of the tests are now different for matplotlib≥3.8,
> 
> Code in `tests/` doesn't matter for code coverage, only code in `anesthetic/` matters.

Ah - I'd thought `pytest --cov anesthetic tests` meant that coverage was looking at anesthetic AND tests, but tests is just to tell pytest where the tests are. I don't think I specified this in the new pytest calls

lukashergt: I agree that there might be a point where we'll want to re-introduce this for backwards compatibility, but I'd rather be aware when that happens.

lukashergt (approved):
Thanks a bunch @AdamOrmondroyd! Feel free to squash and merge.

.github/workflows/CI.yaml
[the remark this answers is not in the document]
lukashergt: Ah, yeah, we had this before... I don't retain this stuff...^^

AdamOrmondroyd:
@lukashergt done, please can you check that the PR settings are adjusted accordingly?

lukashergt:
New CI jobs are still not part of branch protection, so PRs won't require these. I guess new jobs need to be added manually to the branch protection rules.

AdamOrmondroyd:
Cool, thanks for checking

lukashergt:
Thanks for taking charge of this, this makes anesthetic a lot friendlier to users other than ourselves.

PR #345, opened 2023-09-28
Fix Cholesky computation for Gelman--Rubin statistic and add per-parameter feature

lukashergt:
There was a little bug in the computation of the Gelman--Rubin statistic. While `np.linalg.cholesky` returns the lower triangular decomposition, `scipy.linalg.cholesky` returns the upper triangular decomposition. Depending on which one we use, we will have to adapt the transposing in matrix multiplications.

This PR also add the feature of returning the Gelman--Rubin statistic _per parameter_, which can sometimes be helpful when assessing convergence and trying to identify the culprit that is difficult to sample.

# Checklist:


anesthetic/samples.py
```diff
@@ -573,7 +573,7 @@ def remove_burn_in(self, burn_in, reset_index=False, inplace=False):
             data = data.reset_index(drop=True, inplace=inplace)
         return data
 
-    def Gelman_Rubin(self, params=None):
+    def Gelman_Rubin(self, params=None, return_per_param=False):
```
AdamOrmondroyd, line 576: Perhaps `per_param` instead of `return_per_param`?

AdamOrmondroyd (changes requested):

AdamOrmondroyd:
Looks good, have just checked that this still runs ok on mac (which was why we changed to scipy cholesky in the first place)

anesthetic/samples.py
lukashergt: Have been contemplating that, too, you suggesting it settles it.

Done in https://github.com/handley-lab/anesthetic/pull/345/commits/299fddb31541ac8aa7e6ca72407b32991661cc6b.

AdamOrmondroyd (approved):

lukashergt:
Ha, was about to hit "request review", but @AdamOrmondroyd was even faster...^^
Thanks, Adam.

AdamOrmondroyd:
You actually beat me first; I'd made the changes to the docs and hit `git push` only to be met with rejection

I should probably have read the scipy cholesky docs closer before implementing #287

AdamOrmondroyd (approved):

PR #347, opened 2023-10-12
rasterize scatter in plot_2d

mrosep:
# Description

Hi all. Opening this pull request mostly as a learning exercise. 

A while back when I was using anesthetic to produce figures for my first year report, I noticed that several of the pdf viewers I was using took a long time to load the scatter plots. For my report I ended up rasterizing just the scatter plots to render these as pngs while keeping everything else in the plot as vectors, and this made the plots much less 'sticky' when loading my pdf. At the time, Will suggested that this could be included as an option in anesthetic so I'm submitting this pull request. Let me know your thoughts - not sure where to proceed from here in terms of carrying out some of the below checklist items so any guidance on this would be appreciated.

Fixes # (issue)

# Checklist:

### DIFF
--- a/anesthetic/samples.py
+++ b/anesthetic/samples.py
@@ -245,7 +245,7 @@ class Samples(WeightedLabelledDataFrame):
 
         return axes
 
-    def plot_2d(self, axes=None, *args, **kwargs):
+    def plot_2d(self, axes=None, rasterize_scatter=False, *args, **kwargs):
         """Create an array of 2D plots.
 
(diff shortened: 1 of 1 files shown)


williamjameshandley (changes requested):
Many thanks for this @mrosep -- I've often found the need to manually implement this to reduce the size of plots, so this would be really useful.

We may want to consider making this eventual default behaviour (perhaps with a deprecation warning for a few iterations first). What do @lukashergt and @AdamOrmondroyd think?

PR #353, opened 2023-11-08
NDE implementation

### LINKED ISSUE #334: Normalising Flows as NDEs for production-grade plots
Had a brief discussion with @williamjameshandley offline about using normalizing flows (NFs) as Neural Density Estimators (NDEs) instead of KDEs for production-grade anesthetic plots. I've been doing some work recently that shows that NFs typically out-perform KDEs when used to estimate the KL divergence/BMD of target distributions. This suggests that they can be used to better represent the underlying samples in a distribution for production grade plots.

We would like to add a `kind=nde` option to the plotting functionality in `anesthetic` and integrate in [`margarine`](https://github.com/htjb/margarine) for NF training. A simple example is shown below.

```python
import numpy as np
from margarine.maf import MAF

samples = np.random.multivariate_normal([0, 0], 
                                        [[1, 0.], [2, 0.1]], size=5000)
f = MAF(samples)
f.train(1000, early_stop=True)

x = np.linspace(samples[:, 0].min(), samples[:, 0].max(), 100).astype(np.float32)
y = np.linspace(samples[:, 1].min(), samples[:, 1].max(), 100).astype(np.float32)
xv, yv = np.meshgrid(x, y, indexing='ij')

fig, axes = plt.subplots(1, 1)
    
lp = f.log_prob(np.array([xv.flatten(), yv.flatten()]).T).numpy()
z = np.exp(lp - lp.max()).reshape(xv.shape)

plt.scatter(samples[:, 0], samples[:, 1], s=1, c='k', alpha=0.5)
axes.contourf(xv, yv, z, cmap='Blues', levels=[0.68, 0.95, 1.00], alpha=0.8)

plt.tight_layout()
plt.savefig('kind=nde.png', dpi=300)
plt.show()
```

Which produces the following plot

kind=nde

For multi-modal distributions we can take advantage of the clustering built into `margarine`. The flows take seconds to minutes to train depending on number of samples and dimensionality.

htjb:
# Description

Adding in optional neural density estimator plot functions as an alternative to KDE and histogram plotters.

Fixes #334 

# Checklist:

### DIFF
--- a/anesthetic/plot.py
+++ b/anesthetic/plot.py
@@ -27,6 +27,8 @@ from anesthetic.utils import (sample_compression_1d, quantile,
                               iso_probability_contours,
                               match_contour_to_contourf, histogram_bin_edges)
 from anesthetic.boundary import cut_and_normalise_gaussian
+from margarine.maf import MAF
+from margarine.clustered import clusterMAF
 
 
 class AxesSeries(Series):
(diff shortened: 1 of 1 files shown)


AdamOrmondroyd:
Just at a glance, you might want to have a look at how the optional dependency of `fastkde` is handled so you can do something similar with `margarine`

anesthetic/plot.py
```diff
@@ -974,6 +974,175 @@ def kde_plot_1d(ax, data, *args, **kwargs):
     return ans
 
 
+def nde_plot_1d(ax, data, *args, **kwargs):
+    """Plot a 1d marginalised distribution.
+
+    This functions as a wrapper around :meth:`matplotlib.axes.Axes.plot`, with
+    a normalising flow as a neural density estimation (NDE) provided by
+    `margarine.maf.MAF` or `margarine.clustered.clusterMAF`
```
AdamOrmondroyd, line 982: Getting the docs to work properly for an optional dependency seems like a pain, so I've copied the fastkde approach of just putting a link to margarine

htjb: I don't understand, what was the issue here?

AdamOrmondroyd: There's probably a few layers to this, but part of it is that margarine only contains submodules (so `from margarine.maf import MAF` works but `import margarine; margarine.maf.MAF` does not, and I couldn't get to the bottom of how to refer to them properly.

You'll be able to see the error messages in the earlier checks

AdamOrmondroyd: Even changing the api hasn't made it straightforward - @lukashergt is the king when it comes to the docs

AdamOrmondroyd:
tests now failing as expected (I'd forgotten to add margarine to the 'all' dependency!)

anesthetic/plot.py
lukashergt: To explain a bit more, the issue was that the docstring contained ``:class:`margarine.maf` ``, which tells autodoc to try and link to the documentation of a class called `maf` in a package called `margarine`, which autodoc wasn't able to find, though. Two issues here:

1. There is no dedicated documentation for the class  ``:class:`margarine.maf` ``, there is only documentation for the method [``:meth:`margarine.maf.MAF` ``](https://margarine.readthedocs.io/en/latest/functions.html#module-margarine.maf.MAF).
2. In order for autodoc to find the margarine documentation, you need to tell it where to look. In `docs/source/conf.py` you can find a dictionary that tells autodoc where to do that for matplotlib, pandas, and Co. You'll need to add the margarine readthedocs url to that dictionary in order to get the automatic linking to work.

AdamOrmondroyd: 1. I think `:class:` is the right thing, since `margarine.maf` is a submodule, and `MAF` is a class inside that submodule.
2. I'd tried adding this and couldn't get it to work 😢

lukashergt: Ah, it appears that the margarine docs don't build all the needed reference targets. I've created a PR htjb/margarine#54 that uses more of the autodoc power that should hopefully remedy that.

htjb: Ahh okay I see! Thanks @lukashergt I will have a look at your PR on margarine asap.

htjb:
@AdamOrmondroyd Cant remember what needs doing here? is it just the tests need tidying? and master merging?

AdamOrmondroyd:
> @AdamOrmondroyd Cant remember what needs doing here? is it just the tests need tidying? and master merging?

The tests need a rethink, because simply adding nde everywhere we previously saw kde slows everything down too much

Also at least one of them appears to have failed at the last time of running, and you're missing some coverage

htjb:
Ahh yes the NDEs are not particularly quick... maybe like a few minutes for a few thousands of samples (dependent on the architecture). Has the potential to slow the tests down...

Looks like the test failed because of a version mismatch between TensorFlow and TensorFlow Probability. What does the `extras` flag turn on in the tests?

lukashergt:
> @AdamOrmondroyd Cant remember what needs doing here? is it just the tests need tidying? and master merging?

This also needed waiting for the margarine docs to work, which is now the case. So now the NDE docstrings properly link to the margarine docs.

htjb:
Closing this PR as it's become stale and would require a complete rewrite given changes recent changes to `margarine`.

PR #354, opened 2023-11-13
Dead points and Truncation

yallup:
# Description

Feature request for an additional convenience alias for `dead_points` matching the `live_points` style, and a `truncate` method that gives the consistent union of the two allowing recalculation of mid run statistics

# Checklist:


anesthetic/samples.py
```diff
@@ -1138,6 +1140,57 @@ def live_points(self, logL=None):
         i = ((self.logL >= logL) & (self.logL_birth < logL)).to_numpy()
```
yallup, line 1140: Currently tests are failing due to the second check here, is this defined correctly (should the logL_birth check be inclusive)?

williamjameshandley: It all depends on the convention as to whether or not you generate new live points subject to a strict or non-strict hard likelihood constraint. At the moment many nested samplers specify strict (polychord and multinest for example), so logL_birth < logL, but this is not necessarily required. Perhaps we could create an issue so that anesthetic can work with either convention?

williamjameshandley: I think that the convention is consistent i.e. this version of the strictness ensures that the 'live points at iteration i' corresponds to 'live points at loglikelihood logL' in a sensible way.

I think the real issue is that samples.live_points() (i.e. the None functionality) is not returning the right thing. This should return 'the last set of live points', and at the moment returns the penultimate set.

I've pushed a new function 'contour' that aims to achieve this. Could you tidy up the docstrings with a new users eye?

williamjameshandley (changes requested):
This is a useful addition (I and @zixiao-h have certainly used this functionality)

Copying inline comments here:

>> @yallup: Currently tests are failing due to the second check here, is this defined correctly (should the logL_birth check be inclusive)?

> @williamjameshandley: I think that the convention is consistent i.e. this version of the strictness ensures that the 'live points at iteration i' corresponds to 'live points at loglikelihood logL' in a sensible way.
> 
> I think the real issue is that samples.live_points() (i.e. the None functionality) is not returning the right thing. This should return 'the last set of live points', and at the moment returns the penultimate set.
> 
> I've pushed a new function 'contour' that aims to achieve this. Could you tidy up the docstrings with a new users eye?

yallup:
> This is a useful addition (I and @zixiao-h have certainly used this functionality)
> 
> Copying inline comments here:
> 
> > > @yallup: Currently tests are failing due to the second check here, is this defined correctly (should the logL_birth check be inclusive)?
> 
> > @williamjameshandley: I think that the convention is consistent i.e. this version of the strictness ensures that the 'live points at iteration i' corresponds to 'live points at loglikelihood logL' in a sensible way.
> > I think the real issue is that samples.live_points() (i.e. the None functionality) is not returning the right thing. This should return 'the last set of live points', and at the moment returns the penultimate set.
> > I've pushed a new function 'contour' that aims to achieve this. Could you tidy up the docstrings with a new users eye?

Ah that makes sense and is consistent with what I was seeing, I added a small change to cater for if a float (implicitly a logL val) is passed but it happens to form a valid index (i.e 10.0). Docstrings seem sensible to me now, added a bit there an an explict contour test

anesthetic/samples.py
```diff
@@ -1171,8 +1171,41 @@ def logL_P(self, nsamples=None, beta=None):
...
+            logL = self.loc[self.logL > self.logL_birth.max()].logL.iloc[0]
+        elif isinstance(logL, float):
+            pass
+        else:
+            try:
+                logL = float(self.logL[logL])
+            except KeyError:
+                pass
+        return logL
+
```
williamjameshandley, line 1206: This is not covered by the tests -- I agree that your additional float check is safer than what was there before. Can we think of anything which the try except covers (given that it is now never hit by the tests?) if not, remove the extra code, and just move to straight return statements from the if elif else block.

williamjameshandley (changes requested):

williamjameshandley (approved):
Please squash and merge!

PR #356, opened 2023-12-05
Fix `./bin/bump version.py` on macOS

### LINKED ISSUE #355: `bin/bump_version.py` does not work on macOS
**Describe the bug**
On macOS, `bin/bump_version.py` doesn't do anything.

**To Reproduce**
`$ ./bin/bump_version.py; git log` on macOS shows nothing has happened.

**Expected behavior**
The version number increases and is committed.

AdamOrmondroyd:
# Description

`./bin/bump_version.py` was not working on macOS, as its version of `sed` requires an backup for in-place editing, or an empty string to turn it off (`-i''`, without the space).

I tried to do something more subtle that would work cross-platform, but getting an empty string argument into `subprocess.run()` doesn't seem straightforward. I think a cleaner solution would be to automate this entirely in python or bash rather than mixing the two, but of course at the cost of a larger diff.

Fixes #355 

# Checklist:


AdamOrmondroyd:
I think the macOS runners were having a bad time earlier, seems ok now

williamjameshandley (approved):
OSX madness, but never mind!

williamjameshandley (dismissed):
Apologies -- I didn't realise this was ready to go -- the red cross is misleading due to the failing but not essential checks.

williamjameshandley (approved):

AdamOrmondroyd:
> Apologies -- I didn't realise this was ready to go -- the red cross is misleading due to the failing but not essential checks.

the :x: indeed do not spark joy. I kind of lost motivation with #359 since some of the warnings go away with recent development pandas

PR #359, opened 2024-01-23
Pandas 2.2.0

AdamOrmondroyd:
# Description

Pandas 2.2.0 just released, with a flurry of changes to the plotting to become more stateless (though not even close to achieving it), and a whole flood of deprecations ready for pandas 3.0 (!!!!!!!)

Unfortunately, the changes to the plotting are incompatible with the old

I'm yet to do anything about the warnings

Fixes # (issue)

# Checklist:


anesthetic/plotting/_matplotlib/hist.py
```diff
@@ -180,18 +179,18 @@ def __init__(
     ) -> None:
         super().__init__(data, bins=bins, bottom=bottom, **kwargs)
 
-    def _calculate_bins(self, data):
-        if 'range' not in self.kwds or self.kwds['range'] is None:
+    def _calculate_bins(self, data, bins):
+        nd_values = data.infer_objects(copy=False)._get_numeric_data()
+        values = np.ravel(nd_values)
+        values = values[~isna(values)]
```
AdamOrmondroyd, line 185: I'm not happy with repeating this a third time, and I haven't checked if the weights definitely need checking for nans

AdamOrmondroyd:
I think that's all the new deprecation warnings, aside from the warnings about constructing from a `BlockManager`. Discussion on this seems to be ongoing, so I'm tempted to wait until pandas 2.2.1 to see if I still need to deal with these. https://github.com/pandas-dev/pandas/issues/57032

AdamOrmondroyd:
PR here which may clean up some of the BlockManager mess, https://github.com/pandas-dev/pandas/pull/57553

anesthetic/plotting/_matplotlib/hist.py
lukashergt: Well, considering that `Hist1dPlot` was created for a `WeightedSeries`, we could simplify things here. 
Even simpler than my previous suggestion:
```python
            xmin, xmax = data.quantile(q).to_numpy().ravel()
```

Not sure what is easier for maintenance in the long run, though, so happy with either approach.

lukashergt (dismissed):
Is there anything else that needs to be taken into account for this?

If not, I am happy with this. Thanks for catching this so quickly @AdamOrmondroyd!

Feel free to handle my inline reply as you see fit.

anesthetic/plotting/_matplotlib/hist.py
AdamOrmondroyd: That's too tidy to resist!

lukashergt (approved):
Thanks @AdamOrmondroyd!

PR #364, opened 2024-03-01
Reading from csv files

williamjameshandley:
# Description

Following conversations with @yallup, this PR implements reading csv files which have been written with the `.to_csv` method. Getting this right for all variations on labels/weights being dropped is a little fiddly, and the current implementation prioritises robustness over speed (reading multiple times in some cases), but will function as a low-latency start point.

For consistency this also moves `WeightedLabelledPandas` into its own file `anesthetic.weighted_labelled_pandas.py` rather than `anesthetic.samples`, which reduces clutter for users just looking at the latter file.

# Checklist:


williamjameshandley:
@yallup does this perform as expected?

yallup (commented):
Just tested and the functionality seems good. Including read_csv into the read_chains logic chain may be useful and then plugs into the gui automatically to boot?

williamjameshandley:
I agree this is a good idea, but leaves a small choice:

If we want it to be consistent with `read_chains` (`read_polychord` etc), then this should really take a root rather than a filename (`root.csv`), which differs from the pandas default `read_csv`.

Thoughts/preferences?

williamjameshandley:
(could always accept both)

yallup:
> I agree this is a good idea, but leaves a small choice:
> 
> If we want it to be consistent with `read_chains` (`read_polychord` etc), then this should really take a root rather than a filename (`root.csv`), which differs from the pandas default `read_csv`.
> 
> Thoughts/preferences?

I was going to remark (before checking that it is actually consistent with pandas as is), that root as a filename arg would be preferable. Although we override pandas I think there is enough local precedent to make this conform to the "root" style chains file reading.

williamjameshandley:
OK, I've implemented both, so you can pass root or root.csv to `anesthetic.read.csv.read_csv`.

yallup (approved):
Does everything I was looking for, am not testing any of the intricacies around labels but looks good to me(rge)

PR #372, opened 2024-03-16
2 tick minimum also for logscale

### LINKED ISSUE #362: Single ticks in log plots
@lukashergt: I've been using #328 for some GAMBIT plots, which are now looking great. I'm coming up against a corner-case which is causing us some aesthetic issues. 

Here is as close to a minimal working examples as I can make: you have e.g. 5 parameters with very wide ranges in log space (common in particle physics). Plotted at publication width (i.e. 6 inches), we find:

```python
from anesthetic import make_2d_axes, Samples
import numpy as np

N = 5
params = [f'a{i}' for i in range(N)]
fig, axes = make_2d_axes(params, logx=params, logy=params)
samples = Samples(np.exp(10*np.random.randn(1000,N)), columns=params)
samples.plot_2d(axes)
fig.set_size_inches(6, 6)
fig.savefig('log_difficulties.png')
```
log_difficulties

In this, the default log-axis tick choice is somewhat unhelpful, in that for some of the plots there is only one tick (with no minor ticks).

The issue arises I think because where we set the [log scales](https://github.com/handley-lab/anesthetic/blob/b93dec5759cee058a8ab17ab54ba7931fbcf16db/anesthetic/plot.py#L219-L230) is separated from the [tick fiddling](https://github.com/handley-lab/anesthetic/blob/b93dec5759cee058a8ab17ab54ba7931fbcf16db/anesthetic/plot.py#L291-L315), the latter of which I think is only valid for the linear case.

@lukashergt Do you have any thoughts on how this might be fixed?

lukashergt:
# Description

As raised in #362, logscales can lead to plots with only a single tick mark. This PR aims at fixing that, such that every plot has a minimum of 2 tick marks.

Fixes #362 

# Checklist:


lukashergt:
Failing tests after https://github.com/handley-lab/anesthetic/pull/372/commits/0262e964c8796c8684c16c2a7fe14913da2093da (implenting https://github.com/handley-lab/anesthetic/pull/372/commits/b59da430b1ff7fe89546d1a1d8d2c3c63bfc5fef) highlight the problem.

anesthetic/samples.py
```diff
@@ -355,6 +355,8 @@ def plot_2d(self, axes=None, *args, **kwargs):
                         else:
                             ax.plot([], [])
 
+        axes._set_logticks()
```
lukashergt, line 358: I tried setting the log ticks in the same place where we generally deal with tick locations, but `plot_2d` overwrites these locations, so this call at the end of `plot_2d` is needed...

williamjameshandley (approved):
Many thanks @lukashergt!

PR #282, opened 2023-04-11
Changes for pandas 2.0.0

AdamOrmondroyd:
# Description

Pandas2 has now released. Due to the large impact of #280, I have started #270 again here.

Some of these changes affect #235, so I have built on top of that here.

Changes:
- [x] implement `_PlanePlot2d._args_adjust()`
- [x] Remove `HistPlot._makeplot()`, copied from pandas #48888 (which is already outdated)
- [x] remove `mad`, this has been taken out of pandas
- [x] require matplotlib>=3.6.1
- [x] The Gelman-Rubin statistic is broken, as the `groupby()` seems to cause the pandas kwargs to `cov()`to be passed to our `cov()`, which raises the `NotImplementedError` - need to see how this fits in with #272
- [x] adjust_docstrings() is broken
- [x] update pandas requirement to ~= 2.0.0 once it actually releases
- [ ] wait for conda to catch up
- [ ] easier to wait for #280 to merge first

Fixes # (issue)

# Checklist:

- [ ] I have performed a self-review of my own code
- [x] My code is PEP8 compliant (`flake8 anesthetic tests`)
- [x] My code contains compliant docstrings (`pydocstyle --convention=numpy anesthetic`)
- [x] New and existing unit tests pass locally with my changes (`python -m pytest`)
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have appropriately incremented the [semantic version number](https://semver.org/) in both README.rst and anesthetic/_version.py - I think this is worthy of a minor increment

PR #382, opened 2024-04-21
Fix to `color='C2'` plot_2d error post pandas 2

williamjameshandley:
# Description

- Addresses a bug that has arisen post pandas 2.0.0 #359 #282 
- Introduces a failing test, and a fix for the bug.

Since then, color is stripped out of the `MPLPlot`-type `self.kwd`s and placed into `self.color`, so the fix is quite straightforward

This might be a moment to re-evaluate the choices in `_PlanePlot2d.make_plot`, which is the only changed code, but is rather hard to divine exactly what it is doing.

Here is a MWE of the issue, which shows `color='C2'` only colors the diagonals in green, ignoring the instruction for anything derived from `_PlanePlot2d`.

```
samples = read_chains('tests/example_data/pc_250')
samples.plot_2d(['x0', 'x1'], color='C2')
```
Figure_1

# Checklist:


lukashergt (approved):
Looks like this comes from https://github.com/pandas-dev/pandas/pull/55904/commits/a0cbc923cb4c953101023432b4bac1b0c5404103. 

They removed the `'color'` keyword for `PlanePlot` and reintroduce it later downstream for the plot types that require it. I think this has to do with resolving potential conflicts between `color` and `cmap` for plots like `HexBinPlot`. Since we use `_PlanePlot2d` only for anesthetic own plots, I think we don't need to worry about that, so the minimal fix here should be fine.

lukashergt (approved):

PR #391, opened 2024-07-16
add reader for dnest4

qacwnfq:
# Description

This is a draft and first attempt at adding a reader for diffusive nested samples.
I've opened the PR to facilitate easier discussion of the required changes

Fixes # (issue)

# Checklist:


qacwnfq:
@williamjameshandley  Hey Will, I've added the complete output from DNest4 for a 2D-Gaussian with mean 0 and variance=unity. The prior is uniform on a box from -10 to 10.

In addition to the raw files from DNest4 (levels.txt, sample.txt, sample_info.txt),
I've added the additional files, that are produced by the postprocessing of DNest4.
They are log_prior_weights.txt, posterior_sample.txt and weights.txt (probably the posterior weights).
Additionally, there is a file called sampler_state.txt. I would have to check what it does.


For the visualization, I think levels.txt, sample.txt, sample_info.txt + posterior_sample.txt should be enough, because then it would be possible to show "live" points or the posterior.

williamjameshandley:
As a first pass, here is one (non-dynamic) way to visualise a dnest run:

```python
import numpy as np
import os
from anesthetic.plot import basic_cmap

levels_file = 'levels.txt'
sample_file = 'sample.txt'
sample_info_file = 'sample_info.txt'
weights_file = 'weights.txt'

root = 'tests/example_data/dnest4/'

levels = np.loadtxt(os.path.join(root, levels_file), dtype=float, delimiter=' ', comments='#')
samples = np.genfromtxt(os.path.join(root, sample_file), dtype=float, delimiter=' ', comments='#', skip_header=1)
sample_info = np.loadtxt(os.path.join(root, sample_info_file), dtype=float, delimiter=' ', comments='#')
weights = np.loadtxt(os.path.join(root, weights_file), dtype=float, delimiter=' ', comments='#')
n_params = samples.shape[1]

import pandas as pd

df = pd.DataFrame(np.concatenate([samples, sample_info],axis=1), columns=['x0', 'x1', 'level', 'log likelihood', 'tiebreaker', 'ID'])
df.ID = df.ID.astype(int)
df.level = df.level.astype(int)
levels = np.sort(df.level.unique())

cmap = basic_cmap('C0')
fig, axes = plt.subplots(3,3, sharex=True, sharey=True)
for j, ax in enumerate(axes.ravel()):
    ls = levels[:j+1]
    for i, l in enumerate(ls):
        color = basic_cmap('C0')((i+1)/len(ls))
        ax.plot(*df[df.level==l][['x0', 'x1']].to_numpy().T, '.', color=color)
        ax.set_xticks([])
        ax.set_yticks([])
        ax.text(-10,10, f'{j+1}', va='top')
        ax.set_xlim(-10, 10)
        ax.set_ylim(-10, 10)

fig.tight_layout()
fig.set_size_inches(7,7)
fig.savefig('dnest.png')
```

dnest

qacwnfq:
That example code is very useful. With it the replay of samples with a colormap is almost done.

For the LX over log(X) curve, the easiest way right now is to add a function that stores the posterior weights and log(X) from DNest4. 
Right now the log(X) is not written to file and redoing the computation in anesthetic should not be part of this PR.
I'll continue making the prototype and once it's done the code will need some refactoring to fit better into the existing architecture.

qacwnfq:
Here is a prototype
prototype

Some bugs are still there, e.g, the higson plot is not correct. It should look more like the ouptut from dnest4.

qacwnfq:
Hi, 

I've made progress and the PR is ready for review.

Since NestedSamples and DiffusiveNestedSamples support different types of plots, I decided that these classes should each provide their own methods that return the supported plot types and the corresponding points.
Furthermore , I changed the TrianglePlot.update method. It now takes a list of sets of samples and a list of colors. The update method tries to reuse existing lines, but if new ones are added, it has to redraw all axes.

qacwnfq:
@williamjameshandley This PR is ready for review :)

anesthetic/samples.py
```diff
@@ -1164,6 +1168,40 @@ def live_points(self, logL=None):
...
+            temperature
+        logX: np.ndarray
+            prior volumes
+
+        Returns
+        -------
+         LX: np.ndarray
+        """
+        LX = self.logL*beta + logX
+        return LX
```
lukashergt, line 1203: I don't like the naming of this function, it's misleading. Makes me think that it returns `L * X`, which is not the case.

anesthetic/samples.py
```diff
@@ -1334,6 +1372,168 @@ def recompute(self, logL_birth=None, inplace=False):
...
 
+    def plot_types(self):
+        """
+        Get types of plots supported by this class.
+
+        Returns
+        -------
+        tuple[str]
+        """
+        return 'live', 'posterior'
```
lukashergt, line 1383: In the bigger scheme of anesthetic, this is misleading. We have used the kwargs `types` and `plot_type` in the past to indicate KDE plots, histograms, scatter plots, etc. These kwargs have been renamed to `kind`, to unify with naming conventions in pandas and matplotlib.

I don't think this should be a method of `NestedSamples`. This seems to be a more GUI specific thing, so a simple list there makes probably more sense.

anesthetic/samples.py
```diff
@@ -1334,6 +1372,168 @@ def recompute(self, logL_birth=None, inplace=False):
...
+        List[array-like]: list of points to plot
+        List[tuple[float]: colors to use
+        """
+        if plot_type == 'posterior':
+            return [self.posterior_points(beta)[label]], [base_color]
+        elif plot_type == 'live':
+            logL = self.logL.iloc[evolution]
+            return [self.live_points(logL)[label]], [base_color]
+        else:
+            raise ValueError("plot_type not supported")
```
lukashergt, line 1413: This, too, does not feel like it should be a method of `NestedSamples`. Too GUI specific.

tests/example_data/dnest4_no_column_names/README.md
lukashergt, line 1: I think it would be cleaner to have a single subfolder `dnest4` in the `example_data` folder. Can we merge everything from `dnest4_no_column_names` into `dnest4`? `dnest4` itself can have subfolders...

tests/test_gui.py
```diff
@@ -11,19 +11,20 @@ def close_figures_on_teardown():
...
     plotter = samples.gui()
 
     # Type buttons
-    plotter.type.buttons.set_active(1)
-    assert plotter.type() == 'posterior'
-    plotter.type.buttons.set_active(0)
-    assert plotter.type() == 'live'
+    for i, plot_type in enumerate(samples.plot_types()):
+        plotter.type.buttons.set_active(i)
+        assert plotter.type() == plot_type
```
lukashergt, line 27: This change, although one line shorter, is not actually better. For unit tests it is better to repeat and make sure any issue can be pinned to a specific line. Loops leave unclear at which iteration in the loop an issue occured. So better to repeat in tests.

That said, excessive repeats are of course annoying to maintain. But the better way of handling that is with pytest's `parametrize` options.

tests/test_reader.py
```diff
@@ -331,3 +340,84 @@ def test_read_csv(tmp_path, root):
...
+
+    assert isinstance(ns, DiffusiveNestedSamples)
+    assert ns.samples_at_level(9, label='x1').shape == (45, 1)
+    assert ns.plot_types() == ('visited points',)
+
+    ns.points_to_plot('visited points',
+                      label='x1',
+                      evolution=0,
+                      beta=1,
+                      base_color='C0')
```
lukashergt, line 366: `test_reader.py` should be about testing the reading of files, so this and the following plotting calls are a bit out of place.

tests/test_reader.py
```diff
@@ -261,6 +262,14 @@ def test_read_polychord():
     assert_array_equal(ns_zero_live[cols], ns[cols])
     assert_array_equal(ns_single_live[cols], ns[cols])
 
+    with pytest.raises(ValueError) as exc_info:
+        ns.points_to_plot('visited points',
+                          label='x1',
+                          evolution=0,
+                          beta=1,
+                          base_color='C0')
```
lukashergt, line 270: Does not belong in `test_reader.py`.

lukashergt (changes requested):
Hi @qacwnfq, thanks for contributing to anesthetic!

I leave a more detailed review of the diffusive nested sampling stuff to @williamjameshandley. But I've left some comments inline about integration into the anesthetic API.

tests/example_data/dnest4_no_column_names/README.md
qacwnfq: done

tests/test_gui.py
qacwnfq: The idea here was not to decrease code length, but to test the logic, i.e., that setting a certain button to active gives the expected plot_type. 
The relationship between buttons and plot_types are then defined somewhere else and this unit test will not have to be updated, if other plot_types are added in the future.


However, I see the point and have reverted the change.

anesthetic/samples.py
```diff
@@ -1164,6 +1168,40 @@ def live_points(self, logL=None):
...
+        ----------
+        i: i
+            nested sampling iteration
+
+        Returns
+        -------
+        live points at teration i
+
+        """
+        return self.nlive.iloc[i]
```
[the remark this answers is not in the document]
qacwnfq: Good point. In diffusive nested sampling points never die and therefore, we do not have the self.nlive Dataframe.
However, we still want to have a unified interface for "regular" nested samples and diffusive nested samples.

I'll rename the function to get_nlive and the `i` to `iteration`. I'm not sure, we should make `iteration` optional, as the function would then allow for different return types depending on the class. 

I've added the renaming changes for now

anesthetic/samples.py
qacwnfq: I'm not sure I agree on the The idea was that each class in samples.py defines the types of plots supported by objects of said class.
This way plot.py does not need to know any internals of the classes in samples.py to obtain the correct points to plot. The same argument can be made for the  `points_to_plot()` methods.

On the other hand, I do not expect, that anesthetic will need to support many more `Sample` classes, therefore it should be fine to move all of the plotting related stuff to plot.py.

Let me know what you think.

anesthetic/samples.py
qacwnfq: My interpretation was, that it kind of returns L * X but in log space and with beta.

I named it LX after the line here: https://github.com/handley-lab/anesthetic/blob/master/anesthetic/gui/plot.py#L255

Maybe we can call it log_LX or do you have an alternative suggestion?

tests/test_gui.py
qacwnfq: I just remembered why I needed this change. The plot_types are different between diffusive and regular nested samples. I've still made the code more explicit to make it more clear if issues occur during testing:

```
    if isinstance(samples, DiffusiveNestedSamples):
        plotter.type.buttons.set_active(0)
        assert plotter.type() == 'visited points'
    else:
        plotter.type.buttons.set_active(0)
        assert plotter.type() == 'live'
        plotter.type.buttons.set_active(1)
        assert plotter.type() == 'posterior'
        ```

qacwnfq:
@lukashergt Hi Lukas, thanks for the review! I've commented my open questions, let me know what you think

PR #395, opened 2024-09-19
stop warning when `logL==logL_birth==-inf`

lukashergt:
Currently we issue a warning when `logL<=logL_birth`, which includes the case `logL==logL_birth==-inf`. The latter has come up a number of times in my latest runs, spamming me with warnings, which ultimately are nothing to worry about. 

This little PR silences this particular warning when  `logL==logL_birth==-inf` is the only occurrence of `logL<=logL_birth`.

# Checklist:


AdamOrmondroyd:
Code changes are simple enough that I'm happy, but I wanted to ask what your use case is?

lukashergt:
> Code changes are simple enough that I'm happy, but I wanted to ask what your use case is?

I have a bunch of PolyChord runs that produce rows with `0` weight, `-inf` loglikelihood, and `0` in all parameters (I believe this might have to do with the synchronous vs asynchronous PolyChord setup). A uniqueness check in anesthetic reduces these rows to a single one, which ultimately gets dropped alongside throwing this warning. 

These rows are not really a problem (since they have zero weight), so they can be dropped without a problem, and should not need the warning. The warning was created for situations of `logL_birth > logL`, which comes up during importance sampling with a new likelihood.

PR #396, opened 2024-09-19
Normalised stats

lukashergt:
This PR introduces a `norm` keyword for the `NestedSamples.stats` method, that allows passing a previously computed `NestedSamples.stats` instance as a normalisation reference, producing additional columns of differences `[Delta_logZ, ...]`.

Typical cosmology use-case would be when you have multiple models (e.g. LCDM, wCDM, w0waCDM) and you want to have evidence and KL-divergence values relative to one particular model (e.g. LCDM). I was getting annoyed having to type this up every time myself...

# Checklist:


anesthetic/samples.py
```diff
@@ -822,6 +822,11 @@ def stats(self, nsamples=None, beta=None):
         beta : float, array-like, optional
             inverse temperature(s) beta=1/kT. Default self.beta
 
+        norm : Series, :class:`Samples`, optional
+            :meth:`NestedSamples.stats` output used for normalisation.
+            Can be either a Series of mean values or Samples produced with
+            matching `nsamples` and `beta`.
+
```
AdamOrmondroyd, line 829: Need to add the additional columns to the docstring, ideally in a way which doesn't add four more cases to the Returns paragraph below

anesthetic/samples.py
```diff
@@ -861,6 +866,26 @@ def stats(self, nsamples=None, beta=None):
         samples.set_label('d_G', r'$d_\mathrm{G}$')
 
         samples.label = self.label
+
+        if norm is not None:
```
AdamOrmondroyd, line 870: What if `norm` has some but not all of the columns from samples?

AdamOrmondroyd (changes requested):
I like the idea, comments inline

anesthetic/samples.py
lukashergt: Then it will throw an error message saying so. Shouldn't usually be a problem, since I expect `norm` to be a `Samples` instance created by a call to `stats` in the first place, which will automatically compute all relevant columns.

anesthetic/samples.py
lukashergt: done

lukashergt (commented):
Thanks @AdamOrmondroyd, all inline comments addressed.

tests/test_samples.py
```diff
@@ -946,39 +946,93 @@ def test_stats():
...
+    stats = pc.stats(nsamples=nsamples, beta=beta,
+                     norm=pc.stats(nsamples=nsamples, beta=beta))
+    assert isinstance(stats, WeightedLabelledDataFrame)
+    assert_array_equal(stats.drop_labels().columns, vals + delta_vals)
+    assert_array_equal(stats.get_labels(), labels + delta_labels)
+    assert stats.index.names == ['beta', 'samples']
+    assert stats.index.levshape == (len(beta), nsamples)
+
     for beta in [1., 0., 0.5]:
+        np.random.seed(42)
```
AdamOrmondroyd, line 1035: Why do you need a seed here?

AdamOrmondroyd (approved):

tests/test_samples.py
lukashergt: I managed to get failing tests in what follows, because of changes to the seed from the tests I added above... Nothing dramatic, just bad chance in combination with the tests checking against 3*std or similar. I could have fiddled around with tolerances, but just trying a new seed there was enough, so I did not dig deeper...

AdamOrmondroyd: ok, I'll reapprove when you've done the version number

AdamOrmondroyd (approved):
lgtm

PR #402, opened 2025-05-21
Float tidy for contour

williamjameshandley:
# Description

This change updates the `NestedSamples.contour` method to correctly handle NumPy floating-point types (e.g., `np.float32`, `np.float64`) when passed as the `logL` argument.

Previously, if `logL` was a NumPy float, it would not be caught by `isinstance(logL, float)` and would fall into an `else` block. This `else` block attempts to interpret `logL` as an index or mask for `self.logL` (i.e., `float(self.logL[logL])`), which is incorrect when a NumPy float is intended to be the direct contour value. This could lead to errors or unexpected behavior.

The fix adds `np.floating` to the `isinstance` check, ensuring that NumPy float types are correctly recognized and used directly as the contour level. Corresponding tests have been added to verify this behavior for `float`, `np.float32`, and `np.float64` inputs. The version number has been incremented to reflect this bugfix.

# Checklist:


lukashergt (dismissed):
Nice, thanks @williamjameshandley

PR #405, opened 2025-06-12
Only align labels of axes within the 2d axes

### LINKED ISSUE #404: Aligned labels with multiple sets of axes in one gridspec
**Describe the bug**
When multiple sets of anesthetic 2d axes are placed in the same gridspec, their axis labels become aligned, which looks ridiculous. My workaround until recently was to place the axes on the bottom left, until I found that this also happens when another subspec is placed within the gridspec.

`tight_layout()` then spaces things out very strangely.

**To Reproduce**
```python
from matplotlib import pyplot as plt
from matplotlib.gridspec import GridSpec
from anesthetic import make_2d_axes

fig = plt.figure()
gs = GridSpec(2, 2)

make_2d_axes(['x', 'y'], fig=fig, subplot_spec=gs[0, 1])
make_2d_axes(['x', 'y'], fig=fig, subplot_spec=gs[1, 0])
# fig.tight_layout()

plt.show()
```

**Expected behavior**
Axis labels to align only within a single group of 2d axes.

**Screenshots**
Without `tight_layout()`:
Image

With `tight_layout()`:
Image

Expected behaviour:
Image

**Additional context**

AdamOrmondroyd:
# Description

`fig.align_labels()` targets every y and x label in the figure, which is not what we want. I think we should only target the labels in the anesthetic axes being created, and leave it to the user to align labels between separate anesthetic axes, should they desire, with another call to `fig.align_labels()`

Fixes #404

This change has stopped my plots looking like this:
image

so now they look like this:
image

# Checklist:


AdamOrmondroyd:
> Nice catch, thanks @AdamOrmondroyd.
> 
> Flexknot for dynamic DE? I should have figured that you guys might be already on this. Been thinking about this, too, looking forward to the paper :)

They're like London buses [arxiv:2503.08658](https://arxiv.org/abs/2503.08658) [arxiv:2503.17342](https://arxiv.org/abs/2503.17342)
> 
> Wait, the Occam penalty is not kicking in for high `n`...?

We reconciled this as the extra flexknot parameters in the unconstrained high-redshift region being more like unused parameters which do not affect the likelihood, so we should expect the evidences to tend to a constant.
> 
> For a log-redshift axis, `1+z` might be more appropriate than `z`...?

I think I went for $z$ to make comparing against some other reconstructions easier, but personally I'll agree with you. I'll change it to $\log(1+z)$ on my poster for eucaifcon next week and see how I feel afterwards.

AdamOrmondroyd:
Huh codecov has stopped this one too - weird

PR #410, opened 2025-08-13
Fix compression for unweighted data to improve alpha transparency

### LINKED ISSUE #400: alpha has no influence on the scatter plot
I modified example https://anesthetic.readthedocs.io/en/latest/quickstart.html#plot-example-2-marginalised-2d-posteriors and noticed, that the transparency of the scatter plot at the top right does not change when alpha is modified. The effect becomes very clear if one uses stronger colors and reduce the value of alpha. The uploaded image was produced with alpha=0.7 and color='red','blue'.

williamjameshandley:
## Summary
- Fixes Issue #400: Alpha transparency appearing ineffective in scatter plots
- The issue wasn't with alpha parameter itself, but with unweighted data compression not working
- Integer values for `ncompress` now properly subsample unweighted DataFrames
- This reduces point overlaps, making alpha transparency visually effective

## Changes
- Enhanced `WeightedDataFrame.compress()` method to handle unweighted data with integer `ncompress` values
- Added random subsampling for unweighted data when `ncompress` is a positive integer
- Added comprehensive tests for unweighted data compression functionality
- Maintains backward compatibility for all existing compression behaviors

## Test plan
- [x] Added tests for unweighted DataFrame compression with integer values
- [x] Added tests for compression with axis parameter
- [x] Verified compression works for values >= original size
- [x] Verified non-integer ncompress values return unchanged data
- [x] All existing tests continue to pass

Fixes #400

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/anesthetic/weighted_pandas.py
+++ b/anesthetic/weighted_pandas.py
@@ -617,7 +617,23 @@ class WeightedDataFrame(_WeightedObject, DataFrame):
             df = df.set_axis(self._get_axis(1-axis), axis=1-axis, copy=False)
             return df
         else:
-            return self
+            # For unweighted data, if ncompress is an integer > 0, randomly subsample
+            if isinstance(ncompress, int) and not isinstance(ncompress, bool) and ncompress > 0:
+                n_current = self.shape[axis]
+                if ncompress >= n_current:
+                    return self  # No need to compress if requested size >= current size
+                # Random subsample
+                np.random.seed(int(self._rand(axis).sum() * 1000) % 2**32)
+                if axis == 0:
+                    indices = np.random.choice(n_current, ncompress, replace=False)
+                    return self.iloc[indices]
+                else:
+                    indices = np.random.choice(n_current, ncompress, replace=False)
+                    return self.iloc[:, indices]
+            else:
+                # For other ncompress values (True, str, etc.) on unweighted data, 
+                # return self unchanged
+                return self
 
     def sample(self, *args, **kwargs):  # noqa: D102
         sig = signature(DataFrame.sample)
--- a/tests/test_weighted_pandas.py
+++ b/tests/test_weighted_pandas.py
@@ -1030,3 +1030,51 @@ def test_read_csv(tmp_path, mcmc_wdf):
     mcmc_wdf.to_csv(filename)
     mcmc_wdf_ = read_csv(filename)
     pandas.testing.assert_frame_equal(mcmc_wdf, mcmc_wdf_)
+
+
+def test_unweighted_dataframe_compression():
+    """Test that compression works for unweighted DataFrames with integer ncompress."""
+    # Create unweighted data
+    np.random.seed(42)
+    data = WeightedDataFrame(np.random.randn(1000, 3), columns=['A', 'B', 'C'])
+    assert not data.isweighted()
+    
+    # Test compression with integer values
+    for n in [100, 200, 500]:
+        compressed = data.compress(n)
+        assert len(compressed) == n
+        assert not compressed.isweighted()
+        assert list(compressed.columns) == list(data.columns)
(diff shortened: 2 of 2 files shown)


williamjameshandley:
## Visual Verification of Fix

I've tested the alpha transparency fix on both master and the fix branch to confirm it resolves the issue. Here are the visual results:

### Before Fix (Master Branch)
\Master branch - compression broken

**Problem**: `compress(100)` and `compress(500)` both return the original 5000 points unchanged. With so many overlapping points, alpha transparency appears ineffective even at low alpha values.

### After Fix (This Branch)  
\Fix branch - compression working

**Solution**: `compress(100)` correctly returns 100 points and `compress(500)` returns 500 points. With fewer overlapping points, alpha transparency is now visually effective.

### Test Results Summary:
- ✅ **Master**: Integer compression fails - `compress(100)` → 5000 points (broken)
- ✅ **Fix**: Integer compression works - `compress(100)` → 100 points (fixed)
- ✅ **Backward compatibility**: Non-integer values (`True`, `'equal'`, `False`, etc.) still work as expected
- ✅ **Alpha transparency**: Now visually effective due to reduced point overlaps

The fix successfully resolves Issue #400 by enabling proper subsampling of unweighted data, which reduces overlaps and makes alpha transparency useful for data visualization.

anesthetic/weighted_pandas.py
```diff
@@ -617,7 +617,28 @@ def compress(self, ncompress=True, axis=0):
             df = df.set_axis(self._get_axis(1-axis), axis=1-axis, copy=False)
             return df
         else:
-            return self
+            # For unweighted data, if ncompress is an integer > 0,
+            # randomly subsample
+            if (isinstance(ncompress, int) and
+                    not isinstance(ncompress, bool) and ncompress > 0):
```
lukashergt, line 623: We can drop the `not isinstance(ncompress, bool)`. That is already excluded by `isinstance(ncompress, int)`.

Also, this `if` could be combined with the preceding `else` to an `elif`, that might help reduce line wrapping.

anesthetic/weighted_pandas.py
```diff
@@ -617,7 +617,28 @@ def compress(self, ncompress=True, axis=0):
...
+                n_current = self.shape[axis]
+                if ncompress >= n_current:
+                    # No need to compress if requested size >= current size
+                    return self
+                # Random subsample
+                np.random.seed(int(self._rand(axis).sum() * 1000) % 2**32)
```
lukashergt, line 629: What is this line doing? Is this needed?

tests/test_weighted_pandas.py
```diff
@@ -1030,3 +1030,51 @@ def test_read_csv(tmp_path, mcmc_wdf):
...
+
+    # Test compression with non-integer values (should return unchanged)
+    compressed = data.compress(True)
+    assert len(compressed) == 1000
+
+    compressed = data.compress('equal')
+    assert len(compressed) == 1000
+
+    compressed = data.compress(False)
+    assert len(compressed) == 1000
```
lukashergt, line 1064: Avoid `for` loops in tests, rather use pytest's parametrize:
```python
@pytest.mark.parametrize('n_requested, n_expected', [(100, 100),
                                                     (500, 500),
                                                     (1000, 1000),
                                                     (1500, 1000),
                                                     (True, 1000),
                                                     (False, 1000),
                                                     ('equal', 1000)])
```

All of these cases can be tested as:
```python
compressed = data.compress(n_requested)
assert len(compressed) == n_expected
assert not compressed.isweighted()
assert list(compressed.columns) == list(data.columns)
```

lukashergt (changes requested):
Nice, some mostly cosmetic comments inline.

PR #430, opened 2026-04-08
fix Conda badge

lukashergt:
Fix the conda badge which was pointed at an incorrect url.

# Checklist:


AdamOrmondroyd:
I'm not going to pretend I understand conda. The old link seems to take you to an older version which tops out at 2.9.1? Is this to do with compatibility with other packages? I guess the link works

README.rst
AdamOrmondroyd: is this wrong too?

AdamOrmondroyd (changes requested):

README.rst
lukashergt: Oh dear, I am not going to pretend I understand conda either... :sweat_smile: 

But yeah, good catch, this needs updating, too. Done in https://github.com/handley-lab/anesthetic/pull/430/commits/64056cd07d4f74cb848611210b87cd7f4262819f.

lukashergt:
> I'm not going to pretend I understand conda. The old link seems to take you to an older version which tops out at 2.9.1? Is this to do with compatibility with other packages? I guess the link works

I think this came with automating the release on conda via GitHub actions quite a while ago, we just missed the necessary README updates.

README.rst
```diff
@@ -79,7 +79,7 @@ via conda
 
 .. code:: bash
 
-    conda install -c handley-lab anesthetic
+    conda install conda-forge::anesthetic
```
AdamOrmondroyd, line 82: I trust you've tested this on your computer (please don't make me install conda)?

AdamOrmondroyd (approved):
approving assuming you've tested the install locally

README.rst
lukashergt: I certainly have not, I don't fancy contaminating my system with conda... I won't make you install it, if you don't make me, deal? :stuck_out_tongue_winking_eye: 

Both installs should work, the update above is the correct form for the new versions according to [conda-forge](https://anaconda.org/channels/conda-forge/packages/anesthetic/overview).

PR #431, opened 2026-04-10
Fix crop invariant contour levels

lukashergt:
`kde_contour_plot_2d` computed iso-probability levels from the KDE evaluated on the plotting meshgrid, whose extent is set by `q`. Because `iso_probability_contours` is a Riemann sum with uniform area weighting, narrowing `q` re-ranks the visible mass and shifts the contours — the "95%" and "68%" labels no longer correspond to the named confidence regions.

The effect is most visible on a standard 2D Gaussian, where the `q=1` window holds only ~47% of the total mass. On `master`, the `q=5` and `q=1` plots disagree: the same data is drawn with contours at different density thresholds.

## Visual Comparison

<details>

<summary> Plotting code </summary>

```python
import numpy as np
import anesthetic
from anesthetic.plot import kde_contour_plot_2d, scatter_plot_2d
np.random.seed(42)
x, y = np.random.randn(2, 1000)
fig, ax = plt.subplots(figsize=(4, 4))
kde_contour_plot_2d(ax, x, y, q=5, label=f"q=5")
kde_contour_plot_2d(ax, x, y, q=2, label=f"q=2", alpha=0.7, lw=2)
kde_contour_plot_2d(ax, x, y, q=1, label=f"q=1", alpha=0.7, lw=2)
ax.legend(title=f"anesthetic {anesthetic.__version__}")
```

</details>

### master

master

### fix (this PR)

fix (this PR)

## Fix

Compute levels with `iso_probability_contours_from_samples` evaluated on the KDE at the sample vertices instead. The sample-based estimator is a Monte-Carlo estimator and is independent of the plotting window, so the `q=5` and `q=1` contours now agree.

## Cleanup

- Dropped the `max(pdf)` that was appended to the levels returned by `iso_probability_contours` and `iso_probability_contours_from_samples`. Callers that need a trailing maximum value (`match_contour_to_contourf`, `np.digitize`) now append it explicitly; the 1D fill loops that only used it to cap their range now iterate over the levels directly.
- Updated `test_q_2d[kde_contour_plot_2d]`: it previously grabbed `p[0].allsegs[0][0]` as the "drawn data", which only worked because the old window-dependent levels happened to produce a window-filling outer contour. The test now flattens polygons across all contourf levels, matching its real intent of checking that `ax.get_xlim()`/`ylim()` equal the bounding box of the drawn region. The reason the unaltered tests would fail can be seen by the tiny green line in the bottom left corner of the second plot: that polygon clearly does not span the entire plotting range.

# Checklist:


AdamOrmondroyd:
ah I'd missed this one!

anesthetic/plot.py
```diff
@@ -1298,27 +1299,46 @@ def kde_contour_plot_2d(ax, data_x, data_y, *args, **kwargs):
...
+    n_samp = min(len(tri.x), max(1000, int(20 / (1 - max(levels)))))
+    idx = np.random.choice(len(tri.x), n_samp, replace=False)
+    x_samp, y_samp, w_samp = tri.x[idx], tri.y[idx], w[idx]
+
+    x_all = np.concatenate([x_grid, x_samp])
+    y_all = np.concatenate([y_grid, y_samp])
+    boundary_kwargs = dict(order=order,
+                           xmin=data_x.min(), xmax=data_x.max(),
+                           ymin=data_y.min(), ymax=data_y.max())
+    P_all = boundary_correction_2d(kde, x_all, y_all,
```
AdamOrmondroyd: why capital P?

AdamOrmondroyd (changes requested):
otherwise lgtm

anesthetic/plot.py
lukashergt: leftover from when `X, Y` were mesh grids and not flat, I'll clean it up

lukashergt: done in https://github.com/handley-lab/anesthetic/pull/431/commits/795c1dc0df821cdda0b566b47ea76bfcf16c8e1c

AdamOrmondroyd:
any preference for this or #433 first?

lukashergt:
> any preference for this or #433 first?

Let's do this first, since this is an actual bug fix.

AdamOrmondroyd (approved):
lgtm; this little bit of green line is my new favourite 

Image

lukashergt:
> lgtm; this little bit of green line is my new favourite

That little green line was a real headache, because it created its own tiny polygon that messed up the tests, because all of a sudden this little polygon did not want to span the entire plotting region as expected... That is seed 42 for you... :sweat_smile:

AdamOrmondroyd:
> > lgtm; this little bit of green line is my new favourite
> 
> That little green line was a real headache, because it created its own tiny polygon that messed up the tests, because all of a sudden this little polygon did not want to span the entire plotting region as expected... That is seed 42 for you... 😅

I've always assumed 42 for Douglas Adams reasons (which I only read last year, loved it, haven't decided whether to read "And Another Thing…"), is that so?

PR #432, opened 2026-04-15
Fix boundary computation to prevent memory blow-up for `ncompress=False`

lukashergt:
# Description

Replace the boundary KDE evaluation in `anesthetic.boundary` with a lower-memory implementation based on `scipy.stats._stats.gaussian_kernel_estimate`. Important bug fix, else the memory can seriously blow up.

# Problem

My previous implementation was competitive speed-wise, because I built a big `nsamples x nplot` array of residuals and then used numpy's vectorisation. Downside: for large `nplot_2d` alongside `ncompress=False` this explodes the memory!

# Implementation

I know, this is a function from a private scipy module, but that function is exactly what is being used under the hood by `kde.evaluate` and it is written in Cython, so it is fast and memory efficient. 

And importantly, it has exactly the functionality we need (which I first started trying to build myself...). It allows passing `values` with shape `(n, p)`, i.e. we can pass other arrays alongside the usual `weights`. The `weights` are just the first column, which will give us the density, but by also passing `weights * data` we are getting at the first moment and thus what we need for the linear correction.

# Checklist:


anesthetic/boundary.py
```diff
@@ -114,9 +120,11 @@ def _corrected_density(f, f_prime, a0, a1, a2, order):
...
     ----------
     f : np.array
-        Uncorrected KDE density.
-    f_prime : np.array
-        Gradient-weighted KDE density.
+        Uncorrected KDE density, shape (M,).
+    moment1 : np.array
+        First bandwidth-scaled kernel moment per axis, shape (d, M).
+        Equivalently, the negative KDE gradient in bandwidth-scaled
+        coordinates, moment1_k = -∂f / ∂(x_k / h_k).
```
AdamOrmondroyd, line 129: 👌

anesthetic/boundary.py
```diff
@@ -62,49 +63,54 @@ def _truncated_moments(x, bw, xmin=None, xmax=None):
...
-def _corrected_density(f, f_prime, a0, a1, a2, order):
+    # some type info needed for Cython
+    output_dtype = np.common_type(kde.covariance, x)
+    spec_dict = {4: 'float', 8: 'double', 12: 'long double', 16: 'long double'}
+    spec = spec_dict[np.dtype(output_dtype).itemsize]
+
+    weighted_fields = np.empty((kde.n, kde.d + 1), dtype=output_dtype)
+    weighted_fields[:, 0] = kde.weights
+    weighted_fields[:, 1:] = kde.weights[:, None] * kde.dataset.T
+    estimate = gaussian_kernel_estimate[spec](kde.dataset.T,
```
AdamOrmondroyd: If I'm gonna allow a private scipy function you could at least call it with the args laid out nicely 😣

AdamOrmondroyd (changes requested):

anesthetic/boundary.py
lukashergt: Not sure what you mean. Do you mean call it with argument names, i.e. `func(a=a, b=b)` rather than `func(a, b)`?

lukashergt: https://github.com/handley-lab/anesthetic/pull/432/commits/1f66ab322e9df76ee5fb0334c19e9c1baf235a53 ?

AdamOrmondroyd: i meant the formatting, i don't like it when all the arguments are all on the right like that

lukashergt: I actually do, but fine.^^
--> https://github.com/handley-lab/anesthetic/pull/432/commits/ef914cd79949d09df4fb516653237a96f91a8ee6

lukashergt:
@AdamOrmondroyd  Anything else missing?

anesthetic/boundary.py
```diff
@@ -62,49 +63,56 @@ def _truncated_moments(x, bw, xmin=None, xmax=None):
...
+        Equivalently, the negative KDE gradient in bandwidth-scaled
+        coordinates, moment1_k = -∂f / ∂(x_k / h_k).
 
     """
-    bw = np.sqrt(np.diag(cov))
-    d = x.shape[0]
-    data = kde.dataset  # (d, N)
-    weights = kde.weights  # (N,)
-
-    diff = x[:, :, None] - data[:, None, :]  # (d, M, N)
```
lukashergt, line 96: This was the culprit that blows up the memory for `ncompress=False`. For some 10'000 samples this gets biiiiig...

AdamOrmondroyd:
> @AdamOrmondroyd Anything else missing?

is this a trick?

lukashergt:
> > @AdamOrmondroyd Anything else missing?
> 
> is this a trick?

Huh? No, is https://github.com/handley-lab/anesthetic/pull/432/commits/ef914cd79949d09df4fb516653237a96f91a8ee6 not what you asked for?

AdamOrmondroyd:
> > > @AdamOrmondroyd Anything else missing?
> > 
> > 
> > is this a trick?
> 
> Huh? No, is [ef914cd](https://github.com/handley-lab/anesthetic/commit/ef914cd79949d09df4fb516653237a96f91a8ee6) not what you asked for?

no that lgtm, just felt like a setup where you knew there was something else missing, but I hadn't spotted yet. Maybe I'm too suspicious (reading death's end atm)

AdamOrmondroyd (approved):
lgtm

anesthetic/boundary.py
AdamOrmondroyd: image

lukashergt:
> (reading death's end atm)

Oh boy... I really liked the first one, the second one was already getting a bit lengthy, but the third one I really had to fight myself through... A bit too ambitious in scope without what it takes to get it there. 

I recommend the Ancillary series! Cool concepts, happy to chat :)

AdamOrmondroyd:
I'm enjoying the third (just less than halfway), was worried before I started about it looking longer than the other two, but I think the very short chapters are responsible for that. I found the lack of chapters in the second certainly something different (and made it both easier and harder to read "at least a chapter" before bed…)

PR #433, opened 2026-04-15
Full boundary correction (all the way into the corners, not just edges)

lukashergt:
Sorry, I should have pointed this to #432 rather than master. We should wait with this one until #432 is merged.

# Description

Replaces the separable (product-of-1D-corrections) 2D boundary correction with a full-covariance treatment based on the linear boundary kernel framework.

This is part of multiple tweaks (#431, #432, #434) around better handling of highly correlated data (as you e.g. get for Starobinsky inflation in $n_s, r$ plots) and edge/corner effects that I have split up a bit by theme. This one cleans up the boundary effects in the corners, not just the edges.

## Maths

The old 2D correction applied the 1D Jones (1993) correction independently per axis and multiplied the factors, which is only an approximation for correlated kernels. The new implementation computes the truncated Gaussian kernel moments (`a0`, `a1`, `A2`) using the full kernel correlation structure via inclusion-exclusion on the bivariate CDF at rectangle corners, and applies the local-linear correction formula `p = (f - a1ᵀ A2⁻¹ m1) / (a0 - a1ᵀ A2⁻¹ a1)` from Chacón & Duong (2018), Sec. 4.3.2, Eq. (4.4).  For `d=1` this reduces exactly to the scalar Jones (1993) formula.

## Visuals

This is an extreme example case! And the really problematic part comes only with an improvement for highly correlated datasets that I am working on, which I haven't pushed yet.

<details>

<summary> Plotting code </summary>

```python
import numpy as np
from anesthetic.plot import kde_contour_plot_2d
num = 5000
nplot = 50**2
fig, axes = plt.subplots(2, 5, figsize=(15, 6))
for ax in axes.ravel():
    for i in range(4):
        d = np.random.normal(size=(2, num))
        w = np.random.uniform(size=num)
        d[1] = d[0] + d[1] * 0.02
        kde_contour_plot_2d(ax, d[0], d[1], weights=w, nplot_2d=nplot, q=0, bw_scale=1.0, ncompress=False)
```

</details>

### 1) Without full boundary correction on master

(i.e. without #434 and without this PR)

Note the little specs in the corners (that's not dirt on your screen :wink:).

image

### 2) Without full boundary correction, but with better correlation handling

(i.e. with #434, but without this PR)

This properly blows up the effects in the corners. 

image

### 3) With full boundary correction, but without the better correlation handling

(i.e. without #434, but with this PR)

Note how the little specs in the corners are gone. However, you can still see the ringing in the 68% contour from the underlying grid, which can get as bad as interrupted contours (happened for me in Starobinsky contours).

image

### 4) With full boundary correction and with better correlation handling

(i.e. with #434 and with this PR)

The full covariance treatments cleans up the corners which the separable approach on master could not. And no more ringing from the grid thanks to the full correlation treatment.

image

## References

* M. C. Jones (1993), 
  "Simple boundary correction for kernel density estimation", 
  _Statistics and Computing_, 3, 135–146. 
  https://doi.org/10.1007/BF00147776
* J. E. Chacón & T. Duong (2018), 
  "Multivariate Kernel Smoothing and Its Applications", 
  _Chapman and Hall/CRC_, Chapter 4. 
  https://doi.org/10.1201/9780429485572-4
  https://www.researchgate.net/publication/345555871_Modified_density_estimation

## Version

In many ways this is a cleanup / tweak of the 2.14.0 minor update from #426. Hence, I am only bumping the micro version.

# Checklist:


lukashergt:
## Math translation

A little bit on the underlying math, to help understanding the implementation, which otherwise might be very confusing when referencing the sources, because the source use different, **conflicting** symbols for their quantities...

### Connection between Chacón & Duong Eq. 4.4 and the Jones-style correction formula

The key point is that Chacón & Duong's symbols $a_0^{CnD}$ and $a_1^{CnD}$ are **kernel coefficients**, whereas in the Jones-style implementation the quantities $a_0^{Jones}$, $a_1^{Jones}$, $a_2^{Jones}$ are **truncated kernel moments**.

### Chacón & Duong notation

Their linear boundary kernel is

$$
K_{{LB}}(w) = \left(a_0^{CnD} + \left(a_1^{CnD}\right)^\top w\right) K(w).
$$

The coefficients are expressed in terms of the truncated moments

$$
\begin{aligned}
    m_0^{CnD} &= \int_B K(w) dw, \\
    m_1^{CnD} &= \int_B w K(w) dw, \\
    M_2^{CnD} &= \int_B w w^\top K(w) dw,
\end{aligned}
$$

via

$$
\begin{aligned}
    a_0^{CnD} &= \frac{1}{m_0^{CnD} - \left(m_1^{CnD}\right)^\top \left(M_2^{CnD}\right)^{-1} m_1^{CnD}}, \\
    a_1^{CnD} &= -a_0^{CnD} \left(M_2^{CnD}\right)^{-1} m_1^{CnD}.
\end{aligned}
$$

### Jones-style implementation notation

In Jones, these truncated moments are

$$
\begin{aligned}
    a_0^{Jones} &\leftrightarrow m_0^{CnD}, \\
    a_1^{Jones} &\leftrightarrow m_1^{CnD}, \\
    a_2^{Jones} &\leftrightarrow M_2^{CnD}.
\end{aligned}
$$

So the Chacón & Duong coefficients become

$$
\begin{aligned}
    a_0^{CnD} &= \frac{1}{a_0^{Jones} - \left(a_1^{Jones}\right)^\top \left(a_2^{Jones}\right)^{-1} a_1^{Jones}}, \\
    a_1^{CnD} &= -a_0^{CnD} \left(a_2^{Jones}\right)^{-1} a_1^{Jones}.
\end{aligned}
$$

### From the kernel formula to the estimator formula

At a fixed evaluation point, the corrected estimator is

$$
\begin{aligned}
    p &= \sum_i w_i \left(a_0^{CnD} + \left(a_1^{CnD}\right)^\top u_i\right) K_i \\
      &= a_0^{CnD} \sum_i w_i K_i + \left(a_1^{CnD}\right)^\top \sum_i w_i u_i K_i.
\end{aligned}
$$

Using

$$
\begin{aligned}
    f &:= \sum_i w_i K_i, \\
    m_1^{Jones} &:= \sum_i w_i u_i K_i,
\end{aligned}
$$

this becomes

$$
p = a_0^{CnD} f + \left(a_1^{CnD}\right)^\top m_1^{Jones}.
$$

Substituting the coefficient formulas gives

$$
p = \frac{f - \left(a_1^{Jones}\right)^\top \left(a_2^{Jones}\right)^{-1} m_1^{Jones}}{a_0^{Jones} - \left(a_1^{Jones}\right)^\top \left(a_2^{Jones}\right)^{-1} a_1^{Jones}}.
$$

This is exactly the Jones-style correction formula used in the code.

AdamOrmondroyd:
do you want to change the base branch to #432?

lukashergt:
> do you want to change the base branch to #432?

Not as simple as that, I would have to rebase this one and make some clean-ups for that to show correctly, which is not worth it if #432 is basically about to be merged... So I'll just merge master in here, once #432 is merged in master.

AdamOrmondroyd:
perhaps we should give #432 a "beta" version then?

lukashergt:
> perhaps we should give #432 a "beta" version then?

Na, nothing beta about #432, that is the pressing bug-fix which I have isolated to push through faster than this more complicated full boundary update. In hindsight I should have called #426 "beta", but back then I thought that would be enough (plot (2) above proves otherwise...).

AdamOrmondroyd:
I'm looking forward to seeing whatever it was you were plotting which sent you down this path

lukashergt:
> I'm looking forward to seeing whatever it was you were plotting which sent you down this path

Collaboration work... might be a while... (has been a while...)

AdamOrmondroyd:
In your maths, why is the second moment a capital M?

lukashergt:
> In your maths, why is the second moment a capital M?

Because it is a matrix, and because that is what C&D are using. 

Hence also captial `A2` in the code.

AdamOrmondroyd:
What about the third moment? [Uppestcase](https://www.google.com/url?sa=t&source=web&rct=j&opi=89978449&url=https://www.youtube.com/watch%3Fv%3DBIRxUzjhqBg&ved=2ahUKEwiDrdnd0fSTAxX8UUEAHURdJtwQwqsBegQIFxAB&usg=AOvVaw1QNgXsLaBzhhUmgfcEihGa) letters?

AdamOrmondroyd:
if we're using _scipy, might as well use their tools to infer the types. Let's see if this passes

AdamOrmondroyd:
Might throw in a test on a couple of different dtypes and check consistency

lukashergt:
> if we're using _scipy, might as well use their tools to infer the types. Let's see if this passes

Ha, this is what I had at first, but then I thought I'd better minimise dependency on private functions. Honestly, either works for me.

AdamOrmondroyd:
I think I'd rather this way; I prefer a louder failure.

Either way, might be worth testing the behaviour with different precisions? At least that it runs?

tests/test_boundary.py
```diff
@@ -43,6 +73,28 @@ def test_boundary_correction_1d():
...
+
+    _, ax = plt.subplots()
+    kwargs = dict(bw_method=0.25, q=0, nplot_1d=300)
+    for order in [-1, 0, +1]:
+        p, = kde_plot_1d(ax, d, weights=w, order=order, **kwargs)
+
+        pdouble, = kde_plot_1d(ax, d.astype(np.double),
+                               weights=w.astype(np.double),
+                               order=order, **kwargs)
+        assert_array_equal(p.get_ydata(), pdouble.get_ydata())
```
lukashergt, line 90: Isn't this a given, and thus kinda needless, since `np.double` is the same as `np.float64` (idem for 2D)?

AdamOrmondroyd: Sorry, skill atrophy on my part. Meant to go for a higher precision forgetting double=float

AdamOrmondroyd: tracing it through, the dtype for that function always comes in as float. I'll remove the test

AdamOrmondroyd (approved):
🙏

lukashergt:
Thanks @AdamOrmondroyd 

("skill atrophy"... you keep teaching me new words/phrases...^^)

PR #435, opened 2026-04-18
Clean boundary closures

lukashergt:
# Description

When the evaluation range spans the full data extent (`q=0`), the KDE boundary correction may not reach zero at the data edges --- e.g. for top-hat priors where the density doesn't naturally taper off. This PR inserts `nextafter` evaluation points just outside `data.min()` and `data.max()` in both `kde_plot_1d` and `kde_contour_plot_2d`, ensuring clean contour/curve closure at the data boundaries.

Motivation: Inflation models sometimes have stronger theoretical bounds from reheating assumptions that data bounds. These theoretical bounds typically cause sharp cuts to the posteriors.

## Visualisation

<details>
<summary>Plotting code</summary>

```python
import numpy as np
from scipy import stats
from anesthetic import make_2d_axes, Samples

num = 10000

x = stats.truncnorm.rvs(-1, 2, size=num)
y = stats.uniform.rvs(loc=-2, scale=4, size=num)
w = np.random.uniform(size=num)
s = Samples(data=np.array([x, y]).T, weights=w, columns=['x', 'y'])

fig, axes = make_2d_axes(['x', 'y'], upper=False, figsize=(4, 4))
s.plot_2d(axes, q=0, bw_scale=1.5)
for ax in axes.to_numpy().ravel():
    if ax is not None:
        ax.set_xlim(-2.5, 2.5)
        ax.set_ylim(-2.5, 2.5)
```

</details>

### Before

image

### After

image

# Checklist:

- [x] I have performed a self-review of my own code
- [x] My code is PEP8 compliant (`flake8 anesthetic tests`)
- [x] My code contains compliant docstrings (`pydocstyle --convention=numpy anesthetic`)
- [x] New and existing unit tests pass locally with my changes (`python -m pytest`)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] I have appropriately incremented the [semantic version number](https://semver.org/) in both README.rst and anesthetic/_version.py

PR #434, opened 2026-04-16
Better handling of highly correlated kde contours

lukashergt:
# Description

## What

`kde_contour_plot_2d` currently evaluates the boundary-corrected KDE on an axis-aligned rectangular grid (`np.mgrid`) and draws with `contourf`/`contour`. For highly correlated posteriors the grid wastes most of its budget on empty space while undersampling the thin correlation ridge, producing jagged / raggedy contours.

This PR:

- Replaces `contourf`/`contour` with **`tricontourf`/`tricontour` over a Delaunay triangulation in kernel-whitened space** (`scaled_triangulation`), so triangles are well-shaped with respect to the correlation structure (resurrected from #216).
- Splits the plotting vertex budget roughly 50/50 between an axis-aligned grid (for clean plotting edges) and data-driven vertices from `triangular_sample_compression_2d` (for coverage of curved / correlated structure that might fall through the cracks of the grid otherwise).
- Raises the default `ncompress` to `max(n_samp, neff(weights, beta='equal'))`, so the compressed-sample vertices also cover the level-estimation and plot-vertex budgets.
- Sets `n_samp = max(nplot // 2, 1000, int(20 / (1 - max(levels))))`, guaranteeing at least ~20 samples in the outermost tail for iso-probability level estimation regardless of `max(levels)`.
- Uses `np.meshgrid(np.linspace(...))` instead of `np.mgrid[...]` for the rectangular grid, so endpoints land exactly on `xmin/xmax/ymin/ymax` (machine-precision; matters for downstream `set_xlim` propagation in triangle plots).

## Why

This resurrects the correlation-aware `scaled_triangulation` + `tricontour[f]` approach that used to live in `kde_contour_plot_2d` but was removed in #216 ("now that we're evaluating on a grid anyhow"). Correlation-awareness was collateral damage in the architectural cleanup; this PR restores it on top of the cleaned-up `nplot`/`ncompress` machinery from #180, #181, #201, #216. The compressed samples now serve dual duty as both KDE anchors and plot vertices, which is what makes this clean.

The concrete motivation are
* **banana-shaped and strongly correlated posteriors** in general, and
* **Starobinsky inflation** in particular, where we get correlations and curved degeneracies between $n_s$, $r$, and $N_*$, and reheating parameters such as $\rho_{reh}$ or $w_{reh}$.

This is part of multiple tweaks for that purpose, this one cleaning up the boundary effects in the corners, not just the edges:

[...]

PR #436, opened 2026-04-27
Rotated grid for KDE evaluation for highly correlated data

lukashergt:
# Description

Replaces #434 and #435.

Turns out, the `tricontour` approach from #434 does still lead to rather jagged contours compared to `contour`. This seems to be rather inherent to the triangulation approach. While it does help nicely with correlated data and data of unusual shapes, I am no longer convinced (or, I guess, _again_ not convinced --> #216) that it is worth degrading the general appearance of contours for these special cases. 

Instead, this PR still uses `contour`, but now optionally on a rotated grid. For highly correlated data (`corr > 0.99`) it switches automatically to a rotated grid in the direction of the major axis of the covariance. Alternatively, a `grid_angle` can be provided to fix the orientation of the grid in that direction. The `grid_angle` can be specified as a single angle in degrees, thus specifying the major direction, or alternatively as a tuple of two angles in degrees, specifying both the major and the minor direction, such that even non-orthogonal grids can be specified.

Nice little bonus: With such a rotated grid, we can perform boundary correction on sloped boundaries, see example below.

## Why

See #434 and #435.

### With full boundary correction and with the rotated grid from this PR.

<details>

<summary> Same example as in #433. </summary>

```python
import numpy as np
from anesthetic.plot import kde_contour_plot_2d
num = 5000
nplot = 50**2
fig, axes = plt.subplots(2, 5, figsize=(15, 6))
for ax in axes.ravel():
    for i in range(4):
        d = np.random.normal(size=(2, num))
        w = np.random.uniform(size=num)
        d[1] = d[0] + d[1] * 0.02
        kde_contour_plot_2d(ax, d[0], d[1], weights=w, nplot_2d=nplot, q=0, bw_scale=1.0, ncompress=False)
```

</details>

No more ringing from the regular (x, y) grid, the rotated grid nicely tracks the correlation.

image

### Cool additional feature: boundary correction along rotated boundary

<details>

<summary> Plotting code. </summary>

```python
import numpy as np
from anesthetic.plot import kde_contour_plot_2d, scatter_plot_2d
from anesthetic.examples.perfect_ns import correlated_gaussian

# np.random.seed(42)
nlive = 500
sigma = 0.1
cov = sigma**2 * np.eye(2)
s = correlated_gaussian(nlive, mean=[0.5, 0.5], cov=cov, columns=['a', 'b'])
mask = s.a - s.b > 0
s = s[mask]

fig, (ax1, ax2) = plt.subplots(1, 2, sharex=True, sharey=True, figsize=(7, 3))
scatter_plot_2d(ax1, s.a, s.b, zorder=0, c='k', alpha=0.1);
scatter_plot_2d(ax2, s.a, s.b, zorder=0, c='k', alpha=0.1);
contf, cont = kde_contour_plot_2d(ax1, s.a, s.b, weights=s.get_weights(), lw=2, alpha=0.7);
contf, cont = kde_contour_plot_2d(ax2, s.a, s.b, weights=s.get_weights(), lw=2, alpha=0.7, grid_angle=45);
```

</details>

On the left you can see the regular approach with a grid in x, y space.

On the right you can see the rotated grid approach. By specifying `grid_angle=45` the boundary on the slope gets nicely corrected as if it were a usual x or y boundary (e.g. useful for ordered neutrino masses or similar).

image

# Checklist:


AdamOrmondroyd:
Good job I hadn't gotten around #433 or #434 then.

The `grid_angle=45` figure is some serious anesthetic tekkers @yallup

lukashergt:
> Good job I hadn't gotten around #433 or #434 then.

Agreed! Very relieved that you hadn't gotten around to that... :sweat_smile: 
Was trying them out in action and was dissatisfied, so I went back to the drawing board...

If worried whether this PR is more trustworthy then those previous ones: This one is much less invasive. The rotated grid only triggers for highly correlated data (or if you explicitly request it by specifying a `grid_angle`), otherwise it is the same regular grid as before.

AdamOrmondroyd:
What happened to the 1D kdes going to zero? Is that here too? I wasn't necessarily convinced on that one

anesthetic/plot.py
```diff
@@ -1636,24 +1640,19 @@ def _plot_window(ax, axis):
...
-        axes_name = f"axes['{ax.get_xlabel()}']['{ax.get_ylabel()}']"
-    else:
-        axes_name = "ax"
     raise ValueError(
-        f"Cannot plot KDE contours: the `{z}` variable has zero variance "
-        f"and no axis limits are set. Call {axes_name}.set_{axis}lim(...) "
-        f"before (and not again after!) plotting to define the display range."
+        f"Cannot plot KDE contours: the {axis}-axis variable has zero "
+        f"variance and no axis limits are set. Call `ax.set_{axis}lim(...)` "
+        f"on the " f"corresponding axis before (and not again after!) "
```
AdamOrmondroyd: what is going on with these strings?

lukashergt:
> What happened to the 1D kdes going to zero? Is that here too? I wasn't necessarily convinced on that one

Now it is (see https://github.com/handley-lab/anesthetic/pull/436/commits/e13ab18e90c7053fa5223204c623648613d8f5e2). 

EDIT: Ah, that is not actually what you meant. You were referring to #435. Anyhow, this comment is about constant data in 1D KDEs.

Little mock demonstration:
```python
import numpy as np
from anesthetic import Samples, make_2d_axes
num = 10000
params = ['w_reh', 'N_star', 'n_s', 'r']
labels = [r'$w_\mathrm{reh}$', r'$N_\ast$', r'$n_\mathrm{s}$', r'$r$']
N_star = np.random.normal(52, 2, size=num)
n_s = 1-2/N_star
r = 12/N_star**2
w_reh = np.random.uniform(-1/3, 1, size=num)
w_fix = np.zeros_like(w_reh)
sw = Samples(data=np.column_stack((w_reh, N_star, n_s, r)), columns=params, labels=labels)
s0 = Samples(data=np.column_stack((w_fix, N_star, n_s, r)), columns=params, labels=labels)
fig, axes = make_2d_axes(params, upper=False, figsize=(5, 5))
sw.plot_2d(axes, ncompress=False, bw_scale=1.5, q=0, lw=0.8, label=r"variable $w_\mathrm{reh}$")
s0.plot_2d(axes, ncompress=False, bw_scale=1.5, q=0, lw=0.8, label=r"fixed $w_\mathrm{reh}=0$")
axes.iloc[-1, 0].legend(bbox_to_anchor=(len(axes), len(axes)), loc=0);
```
constant data in 1D densities and 2D contours

anesthetic/plot.py
lukashergt: Fixed in https://github.com/handley-lab/anesthetic/pull/436/commits/3c95d3382ebe2fc5d9b95a405b8149e0c0437690.

lukashergt:
> What happened to the 1D kdes going to zero? Is that here too? I wasn't necessarily convinced on that one

This question is referring to the changes first shown in #435, which are incorporated here.

Could you be more specific? What are you not convinced by? Implementation or final look?

For Gaussians or anything else that drops off to zero, this will be invisible.

For top-hats and other distributions (notably some derived parameters) that are prior bound in a way that the probability density does not gradually go to zero, it will show the boundaries of the prior (or theoretical) support. Note, it will only do so if you push the plotting limits all the way out to hit the edges of that support window (i.e. typically when setting `q=0`), plus, the plotting window will need to go beyond the data limits (either by setting the view limits explicitly, or by plotting other data with wider support), else the boundaries might be hidden by the plot window as in [the figure above](https://private-user-images.githubusercontent.com/18258042/584836372-3f1a39f2-b38c-483e-bbe0-872b5c250ccd.png?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NzczODUwNDQsIm5iZiI6MTc3NzM4NDc0NCwicGF0aCI6Ii8xODI1ODA0Mi81ODQ4MzYzNzItM2YxYTM5ZjItYjM4Yy00ODNlLWJiZTAtODcyYjVjMjUwY2NkLnBuZz9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNjA0MjglMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjYwNDI4VDEzNTkwNFomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTRiOTEwZjc1ZGFjZjJjZWViN2Q2YzVhODk0NzI5NjVmNWNkNDY0NTQ2ZGI0M2Q2MWQxOWUxNjNmYzkxZjU0MDgmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JnJlc3BvbnNlLWNvbnRlbnQtdHlwZT1pbWFnZSUyRnBuZyJ9.VYc0ifbvN0BZKpQmDe2w8Va0I0rLKX5UTDak7-C60GE).

To me, this makes the difference between 
1) the posterior looking like it is suffering from a visualisation mistake, or
2) an intentional boundary to the 1D density, be it from prior considerations or theoretical constraints.

AdamOrmondroyd:
> To me, this makes the difference between
> 
> 1. the posterior looking like it is suffering from a visualisation mistake, or
> 2. an intentional boundary to the 1D density, be it from prior considerations or theoretical constraints.

but shouldn't a function which drops discontinuously to zero be, erm, discontinuous?

lukashergt:
> but shouldn't a function which drops discontinuously to zero be, erm, discontinuous?

If you were asked to draw a top-hat or the Heaviside step function on a blackboard, how would you draw that? How do you think most people would?

To me this is less a question about mathematical rigour, but about visualisation choices. After all, we are not plotting functions here, but estimations of probability density, with one ingredient being a sharp boundary for which the estimated density needs correction. The vertical line is a visualisation of said boundary. 
A line that ends in the air somewhere might spark the question: "What went wrong there? Did they forget part of the relevant domain?" (Keeping also in mind that we do not draw the zero line outside the prior boundaries...) 
A closed boundary is more intentional: "Up to this point and no further." 

But ok, if you feel strongly about this, we could add a toggle to easily switch this feature on or off...? What default?

AdamOrmondroyd:
I'll compromise with a toggle. I'd probably do a secret third thing at the blackboard: a dashed line at the discontinuity, but that's not a serious suggestion for here.

I'm sure you think the toggle should be on by default, while I think it should be off. We need a tiebreaker @williamjameshandley

AdamOrmondroyd:
I'm having a local test failure which is fun

lukashergt:
> I'm having a local test failure which is fun

What is going wrong? Is it the `test_triangular_sample_compression_2d_retriangulate` test?

AdamOrmondroyd:
> > I'm having a local test failure which is fun
> 
> What is going wrong? Is it the `test_triangular_sample_compression_2d_retriangulate` test?

yes. 
```
====================================================================== FAILURES ======================================================================
________________________________________________ test_triangular_sample_compression_2d_retriangulate _________________________________________________

    def test_triangular_sample_compression_2d_retriangulate():
        # Dataset known to trigger re-triangulation due to numerical precision
        # differences between `ConvexHull` and `Triangulation.get_trifinder`.
        # The data is nearly collinear (high negative correlation) so the scaled
        # triangulation has very thin triangles, making `trifinder` fragile.
        # Three interior points at machine-epsilon distance from different hull
        # edges increase the chance that at least one triggers j<0 on any platform.
        x = np.array([
            0.963503384885994, 0.963503150900029, 0.963503061245431,
            0.963503961899544, 0.963502018551754, 0.96350193347797,
            0.963504894819475, 0.96350146654318, 0.963500284866466,
            0.963500010757046,
            # interior points at machine-epsilon distance from hull edges
            0.963503380133156,     # ~2e-16 from edge 1-0
            0.963503956129408423,  # ~2e-17 from edge 3-0
            0.963500283495918874,  # ~2e-18 from edge 8-9
        ])
        y = np.array([
            0.00383119232978469, 0.00383124043089176, 0.00383125886154424,
            0.00383107371275365, 0.00383147321494960, 0.00383149070440749,
            0.00383088193551402, 0.00383158669736725, 0.00383182963283048,
            0.00383188598672533,
            0.00383119330683777,
            0.003831074898923960,
            0.003831829914599955,
        ])
        cov = np.array([[3.2950564925744617e-06, -7.036390165774016e-07],
                        [-7.036390165774016e-07, 1.5033900475204585e-07]])
        n_hull = len(ConvexHull(np.column_stack([x, y])).vertices)
        assert n_hull == 10
>       tri, w = triangular_sample_compression_2d(x, y, cov, n=3)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

tests/test_utils.py:172:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
anesthetic/utils.py:860: in triangular_sample_compression_2d
    trifinder = tri.get_trifinder()
                ^^^^^^^^^^^^^^^^^^^
.venv/lib/python3.12/site-packages/matplotlib/tri/_triangulation.py:202: in get_trifinder
    self._trifinder = TrapezoidMapTriFinder(self)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^
.venv/lib/python3.12/site-packages/matplotlib/tri/_trifinder.py:45: in __init__
    self._initialize()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <matplotlib.tri._trifinder.TrapezoidMapTriFinder object at 0x1151be390>

    def _initialize(self):
        """
        Initialize the underlying C++ object.  Can be called multiple times if,
        for example, the triangulation is modified.
        """
>       self._cpp_trifinder.initialize()
E       RuntimeError: Triangulation is invalid

.venv/lib/python3.12/site-packages/matplotlib/tri/_trifinder.py:89: RuntimeError
================================================================== warnings summary ==================================================================
tests/test_boundary.py::test_boundary_projection_snapping
  /Users/adam/phd/anesthetic/.wt/hotfix/anesthetic/boundary.py:184: RuntimeWarning: invalid value encountered in multiply
    hess[..., i, i] = np.where(inf_i, 0, -ui * grad[..., i]) - corr_hess

tests/test_scripts.py::test_gui
  /Users/adam/phd/anesthetic/.wt/hotfix/anesthetic/scripts.py:22: UserWarning: FigureCanvasAgg is non-interactive, and thus cannot be shown
    plt.show()

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
============================================================== short test summary info ===============================================================
FAILED tests/test_utils.py::test_triangular_sample_compression_2d_retriangulate - RuntimeError: Triangulation is invalid
========================================= 1 failed, 1746 passed, 36 skipped, 2 warnings in 118.90s (0:01:58) =========================================
```

lukashergt:
Ah, curious, not the failure I was expecting... I was expecting that maybe the hull point count might fail because of precision differences, but this looks like the desired noise addition was not sufficient to make the triangulation work in your case... Does it even enter the noise block? Does increasing the noise help for you locally?

AdamOrmondroyd:
on 702d it enters the noise block. What exactly do you mean by "increasing the noise"?

lukashergt:
> on 702d it enters the noise block. What exactly do you mean by "increasing the noise"?

What do you mean by 702d?

What I mean is increase the `1e-3` in the noise block, e.g. to `1e-2`:
```python
        noise = np.sqrt(max(0, evals[0])) * 1e-3
        noise += np.sqrt(max(0, evals[1])) * 1e-3 if abs(cov_xy) > 0 else 0
        noise += _plot_window(ax, 'x') * 1e-3 if var_x <= 0 else 0
        noise += _plot_window(ax, 'y') * 1e-3 if var_y <= 0 else 0
        noise = noise * np.random.normal(size=data_x.size)
```

AdamOrmondroyd:
> > on 702d it enters the noise block. What exactly do you mean by "increasing the noise"?
> 
> What do you mean by 702d?
> 

commit 702d990452503a7d418fa0fc5d23e7eeb1abef86.

> What I mean is increase the `1e-3` in the noise block, e.g. to `1e-2`:
> 
> ```python
>         noise = np.sqrt(max(0, evals[0])) * 1e-3
>         noise += np.sqrt(max(0, evals[1])) * 1e-3 if abs(cov_xy) > 0 else 0
>         noise += _plot_window(ax, 'x') * 1e-3 if var_x <= 0 else 0
>         noise += _plot_window(ax, 'y') * 1e-3 if var_y <= 0 else 0
>         noise = noise * np.random.normal(size=data_x.size)
> ```

ok I think we're talking at crossed purposes, i thought you meant the `any j<0` branch. The test which was failing doesn't touch that bit right? I'm seeing the second `get_trifinder()` blow up inside `triangular_sample_compression_2d` itself, not anything downstream of the plot.py noise injection.

lukashergt:
> ok I think we're talking at crossed purposes, i thought you meant the `any j<0` branch. The test which was failing doesn't touch that bit right? I'm seeing the second `get_trifinder()` blow up inside `triangular_sample_compression_2d` itself, not anything downstream of the plot.py noise injection.

Ah, you are right, here in this test we are calling `triangular_sample_compression_2d` directly. Normally, we would have the noise block happening beforehand that should ensure that the triangulation does not fail. 

Ok, this is weird. Why does it have no problem determining the convex hull, but then has a problem with the triangulation...?

lukashergt:
Ah, wait, do you still have the problem or is it resolved?

anesthetic/utils.py
```diff
@@ -836,25 +836,55 @@ def triangular_sample_compression_2d(x, y, cov, w=None, n=1000):
...
+    i = np.concatenate([i_hull, i_inner])
+
+    # Triangulate the chosen vertices in whitened coords: matplotlib's
+    # Delaunay and trifinder are numerically fragile when the original (x, y)
+    # cloud is very tight or near-collinear, but in whitened space the
+    # covariance is identity so the triangulation is well-conditioned.
+    L = np.linalg.cholesky(cov)
+    Linv = np.linalg.inv(L)
+    xw, yw = Linv @ np.vstack([x, y])
+    tri_w = Triangulation(xw[i], yw[i])
```
lukashergt: Why was this changed in https://github.com/handley-lab/anesthetic/pull/436/commits/afa43d507080fcf2f53811d7fba72565723750e2 ?

Why not keep it as
```python
tri = scaled_triangulation(x[i], y[i], cov)
```
Seems like needless code repetition, or am I missing something?

lukashergt: Ok, we want to continue working in whitened space, but in that case I suggest cleaning up `scaled_triangulation` (see https://github.com/handley-lab/anesthetic/pull/436/commits/e6ad64482c9e367f6232ae8a74917ec2f69cfc13).

AdamOrmondroyd:
Without the `np.asarray` cast, I get a bunch of ugly type warnings 
image

lukashergt:
> Without the `np.asarray` cast, I get a bunch of ugly type warnings 

But those are false positives. We should not change perfectly fine code because of false positives, right? That means the type checker needs improving.

`trifinder(x, y)` already returns a numpy array.

AdamOrmondroyd:
We'd picked up an invalid value encountered in multiply warning with `test_boundary_projection_snapping()` sorted by doing the where slightly earlier

AdamOrmondroyd (dismissed):
alright, lgtm

lukashergt:
Wait, I don't follow that commit: https://github.com/handley-lab/anesthetic/pull/436/commits/ec1bad33969dd3792c3a8eb279d1ac22f3e12e2c 

What happened to the minus sign?

AdamOrmondroyd:
> Wait, I don't follow that commit: [ec1bad3](https://github.com/handley-lab/anesthetic/commit/ec1bad33969dd3792c3a8eb279d1ac22f3e12e2c)
> 
> What happened to the minus sign?

Oh lord good spot

written the original way, you have possible infs * 0, which gives a runtimewarning and a nan, which is then of course replaced by 0 in the where. My (now corrected) version makes -ui safe before doing the multiplication

AdamOrmondroyd (dismissed):

AdamOrmondroyd (dismissed):
apologies serious skill atrophy happening on my end. still lgtm

AdamOrmondroyd:
@lukashergt are you still happy with this? Or have you gone a secret third route?

lukashergt:
I'm not sure `pad` is the appropriate keyword for the closed boundary. `pad` is used quite a lot in matplotlib as a number adding white space, which is not what we are doing here... But I haven't come up with a better name, yet, which is why I haven't gotten back to this yet...

AdamOrmondroyd:
`closed_boundary`?

AdamOrmondroyd:
`close_beyond_support`?

AdamOrmondroyd:
`anchor`?

AdamOrmondroyd:
`seal`?

lukashergt:
Haha, not that easy, right? `closed_boundary` could work. I also thought of `closed_contours`, which would be great for 2D, but not as appropriate for 1D...

AdamOrmondroyd:
`continuous_boundary`? I think it has to be more than one word

AdamOrmondroyd:
`clip_to_zero`?

AdamOrmondroyd:
@lukashergt just noticed this is still in my taskwarrior

lukashergt:
> [let's see if 👀 means Lukas likes it](https://github.com/handley-lab/anesthetic/pull/436/commits/b7d1cb8331dce49dcdbf5524dcf4292dfef21863)

Haha, I hadn't seen that commit message before.^^ `clip_to_zero` is pretty accurate as descriptions go, so yeah, let's stick with that.


> @lukashergt just noticed this is still in my taskwarrior

Sorry @AdamOrmondroyd. Since I had that in my devel branch, I was lacking the pressure to get this over the finish line. Looks good, just need another approve, please.

AdamOrmondroyd (approved):

PR #442, opened 2026-05-27
fix AUR release

AdamOrmondroyd:
# Description

Fixes CI

The `aur-release` job has been failing on every push to master since OpenSSH dropped DSA support: `KSXGitHub/github-actions-deploy-aur@v2.7.0`
defaults `ssh_keyscan_types` to `rsa,dsa,ecdsa,ed25519`, so the "Add aur.archlinux.org to known hosts" step errors out with `Unknown key type "dsa"`
and the publish step never runs.

v4.1.3 of the action drops `dsa` from the default (`rsa,ecdsa,ed25519`), so the bump alone fixes it — no input override needed.

Breaking changes between v2.7.0 and v4.1.3, neither of which affects us:
- v3.0.0: `allow_empty_commits` default flipped to `false`. We always bump `pkgver`, so PKGBUILD changes on every release.
- v4.0.0: fixes handling of `inputs.pkgbuild` when it doesn't point at a file literally named `PKGBUILD`. We pass `./PKGBUILD`, so no-op.
# Checklist:


AdamOrmondroyd:
PRs like this it feels daft to increment the version number. Also no real way to test it short of just merging

AdamOrmondroyd:
@lukashergt which flavour of linux do you use?

lukashergt:
> @lukashergt which flavour of linux do you use?

Manjaro i3

AdamOrmondroyd:
@lukashergt can we try this one before 2.15.0? the rotation is so cool it would be nice for it to have a round version number for a moment, plus CI changes are evil

lukashergt (approved):
Since this is just CI stuff, I'll just approve this. Not really familiar with AUR releases, though...

AdamOrmondroyd:
I assume it's relevant for @williamjameshandley , but until we add the CLAUDE.md who can say. I just think it's best to have green CI so that when it goes red we know something new is broken

AdamOrmondroyd:
at any rate it seems to have worked https://aur.archlinux.org/packages/python-anesthetic

AdamOrmondroyd:
I suspect the pypi test, which often seems to fail, is pypi just not updating fast enough

PR #438, opened 2026-05-19
Add CLAUDE.md for Claude Code orientation

williamjameshandley:
## Summary

- Adds a root-level `CLAUDE.md` orienting future Claude Code instances to the repo.
- Covers project overview (pandas-subclass + monkey-patched matplotlib backend), layout, dev workflow, CI matrix expectations, code style, testing conventions (notably the `close_figures_on_teardown` fixture lint check and optional-dependency skip helpers), and common pitfalls (MultiIndex columns, weights-as-index-level, read_chains root semantics).
- Complements `README.rst` rather than duplicating it; favours pointers over re-explanation.

## Test plan

- [ ] `flake8` / `pydocstyle` unaffected (markdown only)
- [ ] Document renders correctly on GitHub
- [ ] Spot-check links and file paths referenced in CLAUDE.md

PR #444, opened 2026-06-24
not `setuptools_scm` after all: tidy up the version scripts in `bin/`

AdamOrmondroyd:
# Description

~#438 reminded me that the version number appears twice, and then we check they match. I've previously used setuptools_scm on personal projects to generate the version number from the git tags, but I always just created the tags manually. Here, the README.rst version is picked up by the CI, to create the tag (I will be borrowing this for my own stuff!)~

`setuptools_scm` turned out to cost rather more than the duplication it removes — writeup in the comment below. What's left is the `bin/` tidy-ups: a shared `readme_version()` in `bin/utils.py` used by the three scripts that parse the README's `:Version:` line, and `gen_PKGBUILD.py` no longer doing `open('anesthetic/_version.py').readlines()[0].split("=")[1]`. `pyproject.toml`, `anesthetic/__init__.py` and `.github/` are byte-identical to master.

Fixes nothing really. Just something to keep me occupied while claude is busy.

# Checklist:

^ ~ironic~


lukashergt:
The following would still work as usual?
```python
import anesthetic
anesthetic.__version__
```

AdamOrmondroyd:
> The following would still work as usual?
> 
> ```python
> import anesthetic
> anesthetic.__version__
> ```

Mostly, seuptools_scm generates the anesthetic._version file upon install so it works mostly just the same. The missing piece (which I think is causing the conda tests to fail) is what to do when the package hasn't actually been installed.

AdamOrmondroyd:
the alternative is that if _version.py doesn't exist (because anesthetic hasn't been installed) then just grep from the README, but then it's not clear why you wouldn't just grep from the README. One feature I do like of setuptools_scm is that if you're not on a tag then you get the SHA of the last git commit:

image

AdamOrmondroyd:
Update: I've reverted the `setuptools_scm` part of this and kept only the `bin/` tidy-ups. Writing up why, since it's not obvious and I'd otherwise try it again in a year.

The approach worked, but installing `setuptools_scm` registers a `setuptools.file_finders` entry point that **replaces** setuptools' sdist file list with "every git-tracked file". Building both branches locally:

| | sdist | contents |
|---|---|---|
| `master` | 121 kB | as published |
| this branch | **6.1 MB** | + `tests/example_data` (4 MB of it `nf_output_points.txt`) |

That's what would have gone to PyPI, and what the AUR PKGBUILD would download and checksum. Fixable with a `MANIFEST.in` allowlist (I had it down to 130 kB), but it's a permanent maintenance surface — any new git-tracked data directory silently ends up in the tarball unless someone remembers.

Three more things it needed:

- `anesthetic/__init__.py` imports `_version` unconditionally, so with the file gitignored a bare checkout can't be imported at all — which is why the conda CI job grew `conda install pip` + an editable install just to make `import anesthetic` work. Fixable with a `try/except` fallback, but it means `__version__` has to have a fake value in some circumstances.
- `fetch-depth: 0` wherever a release is built, otherwise setuptools_scm sees no tags and produces `0.1.dev1+...`. `test-build-n-publish` was silently doing this.
- **The one that decided it:** the conda-forge recipe builds with `--no-build-isolation` and lists only `setuptools` + `pip` in `host:`, so `setuptools_scm` wouldn't be present. I simulated that build against the new sdist — it doesn't fail, it silently produces a package whose metadata says **`version 0.0.0`** (`anesthetic.__version__` is still right, since `_version.py` ships inside the sdist, so the two disagree). The feedstock's test stage is `import anesthetic` + `pip check`, neither of which would catch it. It needs a one-line PR to `anesthetic-feedstock` adding `setuptools-scm`, landed *before* the release. (blackjax does have this in their recipe, which is why theirs is fine.)

So: four new failure modes across PyPI, conda-forge and AUR, three of them silent, in exchange for removing a duplication that `bump_version.py` already writes in one command and `check_version.py` already enforces on every PR. Not worth it here — the two-file bump was never actually the problem.

What's left in this PR is just the cleanups: a shared `readme_version()` in `bin/utils.py` used by the three scripts that parse the README, and `gen_PKGBUILD.py` no longer doing `open('anesthetic/_version.py').readlines()[0].split("=")[1]`. `pyproject.toml`, `anesthetic/__init__.py` and `.github/` are byte-identical to master again.

Unrelated thing I noticed while diffing tarballs: our published sdist ships `tests/*.py` but **not** `tests/utils.py`, which every test module imports — so the test suite in the released tarball can't run. Been true for a while. Happy to fix separately if it's worth fixing.

bin/check_version.py
```diff
@@ -1,24 +1,23 @@
...
 
 run("git", "fetch", "origin", "master")
-previous_version = run("git", "show", "remotes/origin/master:" + vfile)
+previous_version = run("git", "show", "remotes/origin/master:" + VFILE)
 previous_version = previous_version.split("=")[-1].strip().strip("'")
 
-readme_version = run("grep", ":Version:", README)
-readme_version = readme_version.split(":")[-1].strip()
+with open(README) as f:
+    readme = readme_version(f)
```
lukashergt: How about keeping `readme_version` as variable name, and making the function name more descriptive by calling it `get_readme_version`?

bin/bump_version.py
```diff
@@ -1,13 +1,13 @@
...
 import sys
 
-vfile = "anesthetic/_version.py"
+VFILE = "anesthetic/_version.py"
 README = "README.rst"
 
-current_version = run("cat", vfile)
-current_version = current_version.split("=")[-1].strip().strip("'")
+with open(README) as f:
+    current_version = readme_version(f)
```
lukashergt: Why are we getting `current_version` in `check_version.py` from `VFILE`, but here from `README`?

AdamOrmondroyd: No good reason, think this crept in with the back and forth trying to get `setuptools_scm` to work

AdamOrmondroyd: ah — it was to avoid having `get_vfile_version` as well as `get_readme_version()`. which version would you say is _the_ version?

lukashergt: Ha, good question, there is a reason why these have to be consistent...^^

lukashergt (approved):

lukashergt:
lgtm

AdamOrmondroyd:
I'm just a bit sad that setuptools_scm didn't deliver like I'd hoped :(

lukashergt:
> I'm just a bit sad that setuptools_scm didn't deliver like I'd hoped :(

Yeah, something more automated would be nice, but I do prefer having at least one hard coded version number in a human readable format somewhere...

AdamOrmondroyd:
> > I'm just a bit sad that setuptools_scm didn't deliver like I'd hoped :(
> 
> Yeah, something more automated would be nice, but I do prefer having at least one hard coded version number in a human readable format somewhere...

It did, in the README, which then was used to automatically generate the github tag, which is picked up by the python package with setuptools_scm.

I've been bitten recently by another project whose readme version hadn't been updated with the package version, neither of which reflected that there had been essentially a complete rewrite since either of those versions. If it wasn't for conda etc I'd have been so chuffed with this `sl`

AdamOrmondroyd:
[to save you googling](https://www.cyberciti.biz/tips/displays-animations-when-accidentally-you-type-sl-instead-of-ls.html)

lukashergt:
> [to save you googling](https://www.cyberciti.biz/tips/displays-animations-when-accidentally-you-type-sl-instead-of-ls.html)

image

PR #445, opened 2026-08-03
Reduce `stats` peak memory

lukashergt:
# Description

I ran into very high peak memory use when calling `NestedSamples.stats(nsamples=5000)` on a large PolyChord run (~80 parameters with [hillik likelihood](https://github.com/mtristram/hillik) from [arXiv:2511.04733](https://arxiv.org/abs/2511.04733)...). It exceeded a 32 GB memory limit, even though the final statistics table itself is quite small.

Most of the memory came from pandas creating several large `(npoints, nsamples)` matrix copies during the calculation. This PR reduces that peak by using NumPy broadcasting, in-place operations, and dropping large matrices as soon as they are no longer needed. I also moved the shared calculations into two small helpers, which avoids duplicating the same work in `stats`, `logZ`, `D_KL`, `d_G`, and `logL_P`.

I left the simpler `nsamples=None` paths alone where the memory use is small.

## Benchmark

For a local comparison, I used 20,000 nested points, 10 columns, and `nsamples=5000`. At this size, a single `(npoints, nsamples)` matrix already takes about 0.8 GB, so having several copies around quickly adds up.

|             | `master` | This PR |
|-------------|---------:|--------:|
| Peak memory | ~6.4 GB  | ~1.7 GB |
| Runtime     | ~15.1 s  | ~6.4 s  |

This reduces peak memory by about 73%. The NumPy operations also make the calculation more than twice as fast, although reducing memory was the main goal.

I checked the results against `master`; they agree to numerical precision. I also added small exact tests for the `logdX` boundary and interior calculations, and for the new weight normalization helper.

# Checklist:


anesthetic/samples.py
```diff
@@ -1107,16 +1225,15 @@ def D(self, nsamples=None):
...
         """Kullback--Leibler divergence."""
         logw = self.logw(nsamples, beta)
-        logZ = self.logZ(logw, beta)
-        betalogL = self._betalogL(beta)
-        S = (logw*0).add(betalogL, axis=0) - logZ
-        w = np.exp(logw-logZ)
-        D_KL = (S*w).sum()
+        logZ, w = self._compute_logZ_and_weights(logw)
+        columns = logw.columns if logw.ndim > 1 else None
+        del logw  # free it before the next full-size matrix is allocated
```
AdamOrmondroyd, line 1230: don't see many `del`s these days

AdamOrmondroyd (approved):
lgtm. nice work

anesthetic/samples.py
lukashergt: Yeah... Didn't see a way around this one without degrading readability. And this `(npoints, nsamples, nbeta)` one can get pretty big...

Thanks for the review!

PR #446, opened 2026-08-06
Read only selected chain columns to reduce peak memory

lukashergt:
# Description

This is another PR targeting peak memory reduction. See also #445.

This came out of trying to work with large Hillik chains ([GitHub](https://github.com/mtristram/hillik), [arXiv:2511.04733](https://arxiv.org/abs/2511.04733)). These can contain around 80 nuisance parameters and millions of samples, so loading every column quickly becomes too large for memory even when you only want to plot a few cosmological parameters.

The main goal of this PR is therefore to avoid loading data that you already know you won't need.

### `columns`
I added a `columns` option to `read_chains` for Cobaya, GetDist, PolyChord, MultiNest, Nested_fit, and UltraNest chains. It accepts parameter names as well as positional indices and slices. The readers select these columns while loading the files, rather than loading the full table and dropping columns afterwards. Without `columns`, the readers keep their previous behaviour.

I also added `read_paramnames`, which reads parameter names and labels without loading the numerical samples. This makes it possible to inspect the parameters in a chain first, before deciding which ones to load.

### `renames`

I added a separate `renames` option for changing parameter names after selection. For compatibility, the previous full-column renaming behaviour through `columns` still works for now, but raises a `FutureWarning`.

### MCMC: `burn_in`, `thin`, `compress_repeats`

For Cobaya and GetDist MCMC chains, I moved burn-in removal and thinning optionally into the reading process. The readers now discard burned samples and rows removed by thinning before allocating the final sample table. This makes a substantial difference when the original chains are much larger than the part to keep.

The `compress_repeats` kwarg follows naturally from MCMC thinnig. When selecting only parameters of interest, oversampling nuisance parameters can leave many consecutive identical rows (ignoring `logL`). The `compress_repeats` kwarg allows then to merge those runs and sum their weights, separately for each chain and after burn-in removal and thinning. The sample order and chain information is kept as is. This effectively marginalises over the omitted nuisance parameters. The likelihood bookkeeping columns (`logL`, `logP` etc.) are dropped by this, because they would otherwise prevent identical parameter rows from merging.

There are also `MCMCSamples` methods for burn-in, thinning, and repeat-compression _after_ reading. `inplace` and `reset_index` handling is made consistent.

# Checklist:


anesthetic/read/ultranest.py
```diff
@@ -1,40 +1,76 @@
...
     try:
         import h5py
     except ImportError:
         raise ImportError('h5py is required to read UltraNest results')
-    with h5py.File(filepath, 'r') as fileobj:
+    with h5py.File(points_file, 'r') as fileobj:
         points = fileobj['points']
         _, ncols = points.shape
-        x_dim = ncols - 3 - num_params
+        x_dim = ncols - 3 - nparams
```
AdamOrmondroyd: with changes like this no wonder there are 1.4klines to look at

AdamOrmondroyd:
blatant copy-paste from claude as it's late, 1.4klines is a lot for one night, and I need to update my computer and don't want to lose these possibly good suggestions

---

## Confirmed bugs

**1. `renames` silently destroys labels.** The label dict is keyed on the original names, but `columns` is renamed just before construction, so nothing matches:

```python
read_chains('tests/example_data/pc', renames={'x0': 'y0'})
# labels: ['', '$x_1$', '$x_2$', '$x_3$']    # x0's label is gone
```

Every reader has this — `columns = [renames.get(c, c) for c in columns]` without a corresponding remap of `labels`.

**2. UltraNest can't take out-of-order columns.** `anesthetic/read/ultranest.py` does `points[:, [3+x_dim+i for i in indices]]`, and h5py requires strictly increasing fancy indices:

```python
read_chains('tests/example_data/un', columns=['x2', 'x0'])
# TypeError: Indexing elements must be in increasing order
```

The other five readers handle reordering fine (checked the values, not just the headers). Fix is presumably to sort, read, then reorder in memory.

## Behaviour changes worth being deliberate about

**3. `chain` is now always kept** for single-chain Cobaya — the old code dropped it, and the `params.remove('chain')` line in the test is deleted. It'll show up as an extra panel in default `plot_2d`/`plot_1d`. It does fix `remove_burn_in` on single-chain Cobaya, so it's defensible, but worth calling out explicitly rather than slipping in.

**4. The anesthetic-1.0 `burn_in` guard is gone** from `read_chains`. `read_chains(root, burn_in=0.5)` now works for Cobaya/GetDist with new semantics (rows in the *stored* chain, not the expanded one), and dies with an obscure `TypeError` for PolyChord/MultiNest. Fine to remove the guard, but the nested readers should probably reject `burn_in` explicitly.

**5. The `_norm_columns` deprecation path is a bit of a trap.** A name-list that fails lookup *but happens to have length `nparams`* is silently reinterpreted as a rename map. So a single typo in an otherwise-full column list turns into a rename rather than a `KeyError`.

## Minor

**6.** `_read_mcmc_chains` rebinds the outer `indices` inside the loop (in the `compress_repeats` branch). Harmless today only because `usecols` is already computed, but it's a landmine.

**7.** `minuslog` is allocated and resized but never written when `compress_repeats=True`.

**8.** `read_polychord` still does `np.unique(data, axis=0)` after column selection — with few columns selected, rows that were distinct in the full table can now collapse. Didn't trigger on the test data, but the dedup is no longer over the full row.

lukashergt:
I know this seems like a lot, but actually it is much less than it appears, because for one, it is the same thing over and over, just applied to the different readers. And for another, it is just re-ordering lines such that a `vimdiff anesthetic/read/getdist.py anesthetic/read/cobaya.py` or a `vimdiff anesthetic/read/polychord.py anesthetic/read/ultranest.py` is easier to read.


### Addressing your points:

1. I don't agree this is a bug. If you'd only view this as a functionality to rename aliases, e.g. `ombh2` to `omega_b`, then indeed you'd expect to retain the label. However, if you are renaming `x0` to `y0` then I would find it very confusing to retain a `'$x_0$'` label... So I'd say, if you rename a parameter, then you should also pass along `labels` with the new label for that parameter, e.g.:
   ```
   read_chains(
       root,
       columns=['x0'],
       renames={'x0': 'y0'},
       labels={'y0': '$y_0$'},
   )
   ```
   I have adapted the `renames` docstring to clarify the behaviour in https://github.com/handley-lab/anesthetic/pull/446/commits/75d802db1bbb398322a8f0913532d0a162e19ac9.

2. I didn't view this as a bug, but I deliberately ignored this as UltraNest not being as flexible as the others, because this comes down to the HDF5 reader not being as flexible... But we can indeed actually easily pre-process the user input `columns` to sort and deduplicate, then read and re-order. Tackled in https://github.com/handley-lab/anesthetic/pull/446/commits/346ab43da6a587049d0bc0b6a4d54419c66ce64d and https://github.com/handley-lab/anesthetic/pull/446/commits/7e89e4f18bc8550bf64c654701a5b9038b66f4cf.

3. Indeed, that is deliberate, and you already mentioned the reason. There are MCMC methods that require a valid `chain` column, so even a single chain should have one. This makes things also consistent between single-chain and multi-chain samples. 

4. The old guard was specifically a temporary anesthetic-1.0 compatibility guard and marked for removal anyhow. `burn_in` is already documented as applying only to Cobaya and GetDist MCMC chains. For nested samples it makes no sense to begin with and the raised TypeError is sufficient imo.

5. Well, that is me trying to avoid a major version update. The old renaming behaviour is still fully functional. And the suggested accident gets met with a warning that should be enough to point out the mistake.

6. Ok, not that much of a landmine in my opinion, but fine. Renamed in https://github.com/handley-lab/anesthetic/pull/446/commits/f4cdb207426ed626c2a4b6eba4ec6a9d934dc632.

7. That it is never written is deliberate, for the same reason that `logL` is dropped for `compress_repeats=True`. MCMC chains with nuisance oversampling will result in many repeated rows in the parameters of interest (e.g. cosmo params), but the nuisance parameters and thus `logL` (and `minuslogposterior` etc.) will still be different, so if we keep them in the dataframe then we won't be able to compress in the first place. We could try to not even allocate `minuslog` when `compress_repeats=True`, but I don't think the minimal efficiency gain (just a single 1D array) is worth the code complication.

8. Technically maybe possible (not even sure...), but keep in mind that `logL` and `logL_birth` are included in this `np.unique`... It exists only to safeguard against potential point duplicates in `_dead-birth.txt` and `_phys_life-birth.txt`, which are the same points and should not be repeated. So practically I think this point is irrelevant. Or what do you think is the likelihood of genuinely having such duplicated lines in a nested sampling run? Is this really something we need to worry about?

AdamOrmondroyd:
Actual me this time:

1. I agree

2. honestly I'm not that familiar with UltraNest having never used it in anger

3. I agree

4. I agree

5. ~Maybe the previous PR should've been an alpha of itself~ getting this pr and #447 mixed up

6. lgtm

7. yep sounds right

8. repeated `logL` does indeed mean something has gone badly wrong with the sampling, but I think I've encountered it before. I guess this isn't a new "bug" just possibly more likely to occur. I think it's fine to leave.

AdamOrmondroyd (approved):
lgim

lukashergt:
> lgim

Typo or new shorthand I don't get?^^ 
Let's get it merged?

AdamOrmondroyd:
> > lgim
> 
> Typo or new shorthand I don't get?^^ Let's get it merged?

I think either you or Will used to read lgtm as "let's get it merged" which is what it felt like today, rather than lgtm. Been a pretty mid bank holiday, which is why I'm on github

PR #447, opened 2026-08-07
More robust default `ncompress` handling for unweighted plotting

lukashergt:
# Description

Some colleagues tried out anesthetic by passing data directly from Emcee, which apparently produces unweighted samples, and because the chain was very long (~90'000 samples) the plotting was struggling. The problem: for unweighted data the default `ncompress='equal'` returns the total number of samples, which is too high for a KDE default (I might actually do that for a paper plot, but not for day-to-day exploration), and definitely too high for a scatter plot...

New default behaviour (obviously this can be overwritten with a custom `ncompress`):
* `scatter_plot_2d`: `ncompress=min(1000, len(data))`
* `kde_contour_plot_2d`: `ncompress=min(10000, neff('equal'))`

Other changes:
* Added some more explicit documentation of kwargs to `plot_1d` and `plot_2d`, which are arguably the most user facing methods.
* Changed `nplot_1d` and `nplot_2d` to `ngrid_kde` to make it clearer that this applies to KDEs, and to make it possible to use the same kwarg for both 1D and 2D on the fly (i.e. when I don't feel like typing out `lower_kwargs` and `diagonal_kwargs` separately). The old kwargs still work but throw a FutureWarning.

# Checklist:


anesthetic/plot.py
```diff
@@ -1547,9 +1570,11 @@ def scatter_plot_2d(ax, data_x, data_y, *args, **kwargs):
         alias_mapping=dict(lw=['linewidth', 'linewidths'],
                            ls=['linestyle', 'linestyles'],
                            color=['c'],
-                           mfc=['fc', 'facecolor'],
-                           mec=['ec', 'edgecolor'],
-                           cmap=['colormap']),
+                           cmap=['colormap'],
+                           mfc=['fc', 'facecolor', 'markerfacecolor'],
+                           mec=['ec', 'edgecolor', 'markeredgecolor'],
+                           markeredgewidth=['mew'],
```
AdamOrmondroyd, line 1675: mewing mentioned

anesthetic/plot.py
```diff
@@ -895,8 +897,8 @@ def kde_plot_1d(ax, data, *args, **kwargs):
           effective samples in :func:`anesthetic.utils.neff`
           with ``beta=ncompress``.
 
-    nplot_1d : int, default=100
-        Number of plotting points to use.
+    ngrid_kde : int, default=200
+        Number of grid points where the KDE is being evaluated.
```
lukashergt: This PR could also be used to clean up the explicit vs implicit kwarg situation in these plot functions. For most of the parameters listed it makes no sense not to explicitly list them in the function signature. This obviously excludes the kwargs passed to matplotlib, I just mean the anesthetic specific parameters.

Example:
```
def kde_plot_1d(ax, data, weights=None, ncompress=False, ngrid_kde=200,
                levels=None, bw_method=None, bw_scale=1, clip_to_zero=True,
                **kwargs):
```

Thoughts?

anesthetic/plotting/_matplotlib/core.py
```diff
@@ -61,7 +62,12 @@ def _get_xticks(self):
 
 def _compress_weights(kwargs, data):
     if isinstance(data, _WeightedObject):
-        return data.compress(kwargs.pop('ncompress', 'equal'))
+        if data.isweighted():
+            ncompress = int(neff(data.get_weights(), beta='equal'))
```
AdamOrmondroyd, line 69: I reckon `beta='equal'` should be exposed as a kwarg of `_compress_weights()`

anesthetic/samples.py
```diff
@@ -129,10 +129,38 @@ def plot_1d(self, axes=None, *args, **kwargs):
         label : str, optional
             Legend label added to each axis.
 
+        q : int or float or tuple, default=5
+            Quantile to determine the data range to be plotted.
+
+            * ``0``: full data range, i.e. ``q=0``: quantile range (0, 1)
```
AdamOrmondroyd, line 132: do we like single or double backticks? there's a right mix throughout the repo (this isn't blocking, just an observation)

AdamOrmondroyd (changes requested):
just one actual suggestion, plus a change I'll push in a moment

anesthetic/plot.py
AdamOrmondroyd: I'd prefer that. I like it when defaults are the defaults :)

AdamOrmondroyd: this presumably includes 2ef2d82 lol

AdamOrmondroyd: maybe worth its quick own PR instead? wouldn't mind some vim golf to heal my ailing brain

anesthetic/samples.py
lukashergt: In pure txt, I would prefer single backticks, but in readthedocs I would prefer inline code... Unfortunately, sphinx renders single backticks as italics and double backticks as inline code... So since I was torn, I asked Codex/Claude, who told me to use double backticks, so there you go... `\_(*¬*)_/`

anesthetic/plot.py
lukashergt: "ailing brain"? Hope everything is ok?

lukashergt: > this presumably includes [2ef2d82](https://github.com/handley-lab/anesthetic/commit/2ef2d827bb662c03520ea34e2343703fa582c74f) lol

Urgh... yup, docstring--default mismatch there... `1` corresponds to `'entropy'`, not `'equal'`...

anesthetic/plotting/_matplotlib/core.py
lukashergt: No, I don't think that is needed. A custom `beta` could be passed as `ncompress`, which `WeightedMPLPlot._compress_weights` passes on to `WeightedDataFrame.compress(ncompress=ncompress)`, which in turn passes it to `utils.compress_weights(ncompress=ncompress)`, where it gets passed to `neff(beta=ncompress)`. So passing a custom `beta` value is already possible.

Makes you wonder why we are doing this change here, why not just pass on the `'equal'`. This comes down to the introduction of the default cap to 1000, because anything more than that for a scatter plot would be a bit much I think, at least by default...?

anesthetic/plot.py
lukashergt: > maybe worth its quick own PR instead? wouldn't mind some vim golf to heal my ailing brain

Feel free. If not, I'll try to clean that up this evening.

AdamOrmondroyd: too much clauding man, can't get it right by myself anymore 🫠

AdamOrmondroyd: what about parameters which default to mutable things like lists? I think these are generally handled by popping or some secret second thing

lukashergt: > what about parameters which default to mutable things like lists? I think these are generally handled by popping or some secret second thing

I think the standard is a function signature with `something=None` and then `if something is None: something = [0.95, 0.68]` after the docstring.

AdamOrmondroyd: > > maybe worth its quick own PR instead? wouldn't mind some vim golf to heal my ailing brain
> 
> Feel free. If not, I'll try to clean that up this evening.

I've decided cba since the docstrings have already changed here

AdamOrmondroyd: would you like them keyword-only?

lukashergt: `ncompress, ngrid_kde, levels, bw_method, bw_scale, clip_to_zero` etc.? Yes, I'd say keyword only, right? 
(Only exception maybe `weights`...? But probably best to stay close to `master`...)

lukashergt: > I've decided cba since the docstrings have already changed here

"cba"... and Ormondroyd continues to make me google interesting stuff...^^

AdamOrmondroyd: this might mess with normalize_kwargs, for instance `facecolor` has aliases

lukashergt: > this might mess with normalize_kwargs, for instance `facecolor` has aliases

Yeah, it might be best to keep the matplotlib related kwargs as kwargs, i.e. the standard things that get passed on to `plt.plot` or similar, such as color or linestyle stuff.

AdamOrmondroyd: > > I've decided cba since the docstrings have already changed here
> 
> "cba"... and Ormondroyd continues to make me google interesting stuff...^^

clutch brake accelerator, dunno what you've been googling

Edit: as if you're still googling instead of asking local models smh my head

lukashergt: `smh abbreviation` in Google works perfectly fine, tyvm very much ;)

(Keep in mind that I stuck to my flip phone all the way to 2021...)

AdamOrmondroyd: i am mostly joking. I've started putting my phone in black and white when I can to stop myself scrolling. I'd've thought that a flip phone would push you more towards abbreviations though? Could you ssh from it?

anesthetic/plotting/_matplotlib/hist.py
```diff
@@ -198,6 +198,8 @@ def _plot(
         bins,
         **kwds,
     ):
+        # ncompress not valid for histograms
+        kwds.pop('ncompress', None)
```
AdamOrmondroyd, line 204: this is in line with `FastKde1dPlot` with `weights` and `bw_method`

anesthetic/plot.py
```diff
@@ -767,7 +768,8 @@ def make_2d_axes(params, labels=None, lower=True, diagonal=True, upper=True,
     return fig, axes
 
 
-def fastkde_plot_1d(ax, data, *args, **kwargs):
+def fastkde_plot_1d(ax, data, *args, xmin=None, xmax=None,
+                    levels=(0.95, 0.68), density=False, **kwargs):
```
AdamOrmondroyd: I've swapped the list for a tuple so that it's immutable. Might be inclined to change this elsewhere if you're happy with it? Shouldn't break anything

anesthetic/plotting/_matplotlib/core.py
```diff
@@ -59,19 +60,27 @@ def _get_xticks(self):
...
+        else:
+            ncompress = len(data)
+        if ncompress_max is not None:
+            ncompress = min(ncompress_max, ncompress)
+        ncompress = kwargs.pop('ncompress', ncompress)
+        return data.compress(ncompress, weighted=False)
```
lukashergt: I had missed that fastkde also uses the `_CompressedMPLPlot`. Fastkde should not be capped (it does its own form of binning/compression), hence the `_default_ncompress_max` class attribute.

anesthetic/weighted_pandas.py
```diff
@@ -371,33 +372,48 @@ def quantile(self, q=0.5, interpolation='linear'):  # noqa: D102
             return np.nan
         return quantile(self.to_numpy(), q, self.get_weights(), interpolation)
 
-    def compress(self, ncompress=True, weighted=False):
+    def compress(self, ncompress=True, weighted=True):
```
AdamOrmondroyd: did you mean to change the default? this feels like a frighteningly breaking change

anesthetic/weighted_pandas.py
```diff
@@ -636,34 +652,48 @@ def quantile(self, q=0.5, axis=0, numeric_only=None,
             return super().quantile(q=q, axis=axis, numeric_only=numeric_only,
                                     interpolation=interpolation, method=method)
 
-    def compress(self, ncompress=True, axis=0, weighted=False):
+    def compress(self, ncompress=True, axis=0, weighted=True):
```
AdamOrmondroyd: ditto

AdamOrmondroyd (commented):
this just jumped out at me

anesthetic/weighted_pandas.py
lukashergt: This was an inconsistent default between `weighted_pandas.py` and `samples.py` that this change fixes, i.e. `wdf.compress()` behaved differently from `samples.compress()`...

I chose `True` as default, here, because that felt more appropriate for anesthetic and `samples.py` is probably more user-facing than `weighted_pandas.py`... That said, I wonder whether the better default would be to return weighted for weighted input and unweighted for unweighted input...?

anesthetic/weighted_pandas.py
lukashergt: ditto

anesthetic/weighted_pandas.py
AdamOrmondroyd: that makes sense. btw, from my claude: "posterior_points() relies on the default, and its docstring says "Get equally weighted posterior points" — but it currently returns a weighted object (isweighted() == True, weights all 1). Numerically equivalent, but it changes the plotting path, the CSV index structure, and any user code branching on isweighted()."

lukashergt: Hmm, arguably `"Get equally weighted posterior points"` is still perfectly accurate... 
But why keep the 1s around, right? 
Adjusted in https://github.com/handley-lab/anesthetic/pull/447/commits/044a4473ffaf133df4c559c22b7f630ab82be109.

anesthetic/plotting/_matplotlib/core.py
```diff
@@ -59,19 +60,27 @@ def _get_xticks(self):
...
     else:
         return data
```
lukashergt: This is one of the central places where changes were needed:
* Scatter plot and FastKDE stuff need unweighted samples, so we need to check if they are weighted or not, and if so, compress them to unweighted (else just keep all samples in the first instance).
* For scatter plots we do not want a crazy number of points by default, so we are capping this to `ncompress_max=1000`.
* FastKDE plots create their own Fourier grid of the data, so they should get all points (as long as they are no longer weighted), so `ncompress_max=None` for those.

anesthetic/samples.py
```diff
@@ -460,8 +576,11 @@ def to_hdf(self, path_or_buf, key, *args, **kwargs):  # noqa: D102
         return anesthetic.read.hdf.to_hdf(path_or_buf, key, self,
                                           *args, **kwargs)
 
-    def compress(self, ncompress=True, axis=0, weighted=True):  # noqa: D102
-        return Samples(super().compress(ncompress, axis, weighted))
+    def compress(self, ncompress=True, axis=0, weighted=None):  # noqa: D102
+        compressed = super().compress(ncompress, axis=axis, weighted=weighted)
+        if ncompress is False:
+            return compressed
+        return Samples(compressed)
```
lukashergt, line 583: Here we are switching defaults from `weighted=True` to "output matches input". 
Now consistent between `weighted_pandas.py` and `samples.py`.

anesthetic/weighted_pandas.py
```diff
@@ -371,33 +372,52 @@ def quantile(self, q=0.5, interpolation='linear'):  # noqa: D102
             return np.nan
         return quantile(self.to_numpy(), q, self.get_weights(), interpolation)
 
-    def compress(self, ncompress=True, weighted=False):
+    def compress(self, ncompress=True, weighted=None):
```
lukashergt, line 375: Here we are switching defaults from `weighted=False` to "output matches input". 
Now consistent between `weighted_pandas.py` and `samples.py`.

anesthetic/weighted_pandas.py
```diff
@@ -371,33 +372,52 @@ def quantile(self, q=0.5, interpolation='linear'):  # noqa: D102
...
+                return self
+            if not weighted and not np.issubdtype(w.dtype, np.integer):
+                raise ValueError(
+                    'Requesting no compression (`ncompress=False`) together '
+                    'with unweighted output (`weighted=False`) is not '
+                    'supported for reliability-weighted samples.'
+                )
+        elif (isinstance(ncompress, (bool, str)) and not self.isweighted()
+              and not weighted):
             return self
```
lukashergt, line 419: This is checking for various special cases of combinations of `ncompress` (True or False), `weighted` (True or False), whether input is weighted or not, and if it is weighted, whether we have frequency weights (e.g. MCMC) or reliability weights (e.g. nested sampling).

Same for the DataFrame version below.

lukashergt (commented):
> I'm a bit overwhelmed by the amount of changes here, but what strikes me is the changes in the last two commits didn't require any tests to change, which suggests that they're not being tested properly

Yeah, the many docstring additions make this seem like a bigger change than it is.

I'm highlighting the major code changes inline below.

As for the last two commits that you are referring to, they are both really just default behaviour choices:
* https://github.com/handley-lab/anesthetic/pull/447/commits/5f57c924ee6870eebbcbe50adc0501971825a1ea: This is just a change in default behaviour, which we generally do not test for. Previously we had `weighted_samples.py` and `samples.py` behave differently by default; this makes them consistent.
* https://github.com/handley-lab/anesthetic/pull/447/commits/044a4473ffaf133df4c559c22b7f630ab82be109: This is just a change to make things work as before with the new defaults. Things would be correct with or without this change.

The rest comes down to docstring additions and promotion from `kwarg` entry to named keyword in function signature.

anesthetic/plotting/_matplotlib/core.py
AdamOrmondroyd: Just switched this around to save ncompress being computed then thrown away, functionally identical.

anesthetic/plot.py
```diff
@@ -1302,15 +1395,15 @@ def kde_contour_plot_2d(ax, data_x, data_y, *args, **kwargs):
     if ax.get_yaxis().get_scale() == 'log':
         data_y = np.log10(data_y)
 
-    ncompress = kwargs.pop('ncompress', 'equal')
-    nplot = kwargs.pop('nplot_2d', 1000)
-    grid_angle = kwargs.pop('grid_angle', None)
-    bw_method = kwargs.pop('bw_method', None)
-    bw_scale = kwargs.pop('bw_scale', 1)
-    order = kwargs.pop('order', None)
+    if ncompress is None:
```
AdamOrmondroyd, line 1398: the balls up is around here. 'equal' is also wrong as there's the 10k cap. Which is also hard-coded into this function.

AdamOrmondroyd (changes requested):

anesthetic/plot.py
lukashergt: Bah, it's fine. `None` often is a placeholder for some specific default. The docstring clarifies the default, so the balls down for me ;) 

I've adapted the `_default_ncompress_max` to infinity for FastKDE in https://github.com/handley-lab/anesthetic/pull/447/commits/e0587ae8427d34f8d3bf53780cc6d2e7fb235fa5.

AdamOrmondroyd (approved):
lgtm

PR #449, opened 2026-09-01
Tuple of tuples for q

lukashergt:
# Description

This PR **allows `q` to be specified independently for the x and y axes** of 2D plotting functions. This is useful when combining parameters with different support, for example a positivity-constrained `r` with an approximately Gaussian `n_s`.

Alongside the existing shared `q` inputs, this supports:

- `q=(1, 2)` for different integer sigma ranges
- `q=((0, 0.95), 5)` for mixed per-axis specifications
- `q=(0, 0.95, 0.025, 0.975)` as a flat shorthand

Note, this is useful **only for single 2D plots, not for triangle plots** where you have much more than 2 parameters to consider. But this builds the foundation for in the future having dict-type kwargs for triangle plots that could allow per-parameter specifications.

As a **related consistency fix, fastKDE plotting ranges** are now determined from sample quantiles rather than quantiles of the reconstructed PDF on its padded grid. This was raised by @AdamOrmondroyd in [PR #447](https://github.com/handley-lab/anesthetic/pull/447#issuecomment-5428480233) and avoids numerical tails inflating the requested plotting range, which has rendered the tests much more consistent between KDE plots and FastKDE plots.

# Checklist:


anesthetic/plot.py
```diff
@@ -836,15 +836,14 @@ def fastkde_plot_1d(ax, data, *args, q=5, density=False, levels=(0.95, 0.68),
...
-
     try:
         from anesthetic.kde import fastkde_1d
         x, p, xmin, xmax = fastkde_1d(data, xmin, xmax)
     except ImportError:
         raise ImportError("You need to install fastkde to use fastkde")
     p /= p.max()
-    i = ((x > quantile(x, q[0], p)) & (x < quantile(x, q[-1], p)))
+    q = quantile_plot_interval(q=q)
+    i = (x >= quantile(data, q[0])) & (x <= quantile(data, q[-1]))
```
lukashergt, line 846: This addresses the issue raised in https://github.com/handley-lab/anesthetic/pull/447#issuecomment-5428480233 that fastkde was behaving differently to kde in determining the plotting window. Now they are consistent, both using sample quantiles.

tests/test_plot.py
```diff
@@ -660,19 +660,14 @@ def test_contour_plot_2d(contour_plot_2d):
...
-    if contour_plot_2d is fastkde_contour_plot_2d:
-        assert xmin <= 0
-        assert xmax >= 1
-        assert ymin <= 0
-        assert ymax >= 1
-    elif contour_plot_2d is kde_contour_plot_2d:
-        assert xmin == pytest.approx(0, abs=0.01)
-        assert xmax == pytest.approx(1, abs=0.01)
-        assert ymin == pytest.approx(0, abs=0.01)
-        assert ymax == pytest.approx(1, abs=0.01)
```
lukashergt, line 672: This is now much simpler, no more need to distinguish kde from fastkde, both can use the previous kde branch.

tests/test_plot.py
```diff
@@ -660,19 +660,14 @@ def test_contour_plot_2d(contour_plot_2d):
...
-@pytest.mark.parametrize('plot_1d', [kde_plot_1d, hist_plot_1d])
+    assert xmin == pytest.approx(0, abs=0.01)
+    assert xmax == pytest.approx(1, abs=0.01)
+    assert ymin == pytest.approx(0, abs=0.01)
+    assert ymax == pytest.approx(1, abs=0.01)
+
+
+@pytest.mark.parametrize('plot_1d', [kde_plot_1d, hist_plot_1d,
+                                     skipif_no_fastkde(fastkde_plot_1d)])
 def test_q_1d(plot_1d):
```
lukashergt, line 671: Just added fastkde to this test, which now works out of the box.

lukashergt (commented):
This should hopefully be a much easier PR.

Below some comments addressing the adapted fastkde handling.

tests/test_plot.py
AdamOrmondroyd: 🙏

